PR #1, opened 2018-06-25
Basic build

netoneko:
Image size: 3.4Mb


electricmonk:
so `Dockerfile.build` is our build agent image while `Dockerbuild.export` is the production deliverable?

netoneko:
Yes.

PR #3, opened 2018-06-27
Dev

IdoZilberberg:


netoneko:
Regarding events: it looks like they only use certain messages and do not take into account vchains/contract addresses which I suppose we are going to need very soon.

PR #12, opened 2018-07-18
transaction pool will not transfer invalid transactions

erankirsh:


services/transactionpool/test/add.go
```diff
@@ -46,4 +46,23 @@ var _ = Describe("transaction pool", func() {
 
 	})
 
+	It("dont forward an invalid transaction with gossip", func() {
+
+		tx := test.TransferTransaction().MakeInvalid().Build()
+
+		gossip.When("BroadcastForwardedTransactions", &gossiptopics.ForwardedTransactionsInput{
```
talkol: Suggest to change the arg instead of a hard requirement to `mock.Any`

services/transactionpool/test/add.go
```diff
@@ -46,4 +46,23 @@ var _ = Describe("transaction pool", func() {
...
+			Message: &gossipmessages.ForwardedTransactionsMessage{
+				SignedTransactions: []*protocol.SignedTransaction{tx},
+			},
+		}).Return(&gossiptopics.EmptyOutput{}, nil).Times(0)
+
+		_, err := service.AddNewTransaction(&services.AddNewTransactionInput{
+			SignedTransaction: tx,
+		})
+
+		Expect(err).ToNot(HaveOccurred())
```
talkol: According to spec we must return error in this case so PublicApi the caller will know the transaction is invalid and tell the user

services/transactionpool/validate_transation.go
```diff
@@ -0,0 +1,13 @@
+package transactionpool
+
+import (
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+)
+
+func (s *service) validateTransaction(input *protocol.SignedTransaction) (bool, error) {
```
talkol: Consider combining the bool and the error to one return value - error, because every time we have false we need to give a reason in the error.. so the bool is redundant

services/transactionpool/validate_transation.go
```diff
@@ -0,0 +1,13 @@
+package transactionpool
```
talkol: Spelling mistake - "transation" instead of "transaction"

services/transactionpool/validate_transation.go
talkol: Let's make a rule that these external files are no longer receivers, they are static functions that receive all of their dependencies (like VM) in their arguments

test/transactions.go
```diff
@@ -32,3 +37,8 @@ func (t *transferTransaction) WithAmount(amount uint64) *transferTransaction {
 	t.builder.Transaction.InputArguments[0].Uint64Value = amount
 	return t
 }
+
+func (t *transferTransaction) WithInvalidContent() *transferTransaction {
```
talkol: The 30 minutes constant should be configurable so let's use an over the top number in the meantime like a year

test/harness/network.go
```diff
@@ -105,6 +106,24 @@ func (n *acceptanceTestNetwork) SendTransfer(nodeIndex int, amount uint64) chan
 	return ch
 }
 
+func (n *acceptanceTestNetwork) SendInvalidTransfer(nodeIndex int) chan *client.SendTransactionResponse {
```
talkol: SendTransfer and SendInvalidTransfer are now mostly duplicate.. are we planning to add more of these? or does this mean we need to extract something outside to the caller?

test/transactions.go
```diff
@@ -12,6 +16,7 @@ func TransferTransaction() *transferTransaction {
 			Transaction: &protocol.TransactionBuilder{
 				ContractName: "BenchmarkToken",
 				MethodName:   "transfer",
+				Timestamp:    primitives.Timestamp(time.Now().Unix()),
```
talkol: See spec:
https://github.com/orbs-network/orbs-spec/blob/master/interfaces/primitives/protocol.proto#L26

https://golang.org/pkg/time/#Time.Unix

you need UnixNano()

test/harness/network.go
erankirsh: We will later refactor if needed

test/transactions.go
erankirsh: Notice that it's 35 minutes into the future. A future timestamp is invalid as stated in spec:
https://github.com/orbs-network/orbs-spec/blob/master/behaviors/services/transaction-pool.md#addnewtransaction-method

erankirsh: @OdedWx will add a new type to TransactionStatus to indicate this

services/transactionpool/errors.go
```diff
@@ -0,0 +1,14 @@
+package transactionpool
+
+import (
+	"fmt"
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+)
+
+type ErrInvalidTransaction struct {
```
electricmonk: Today @jlevison and I had a discussion and he says that it's not "dogmatic" to use error types, but rather use `fmt.Errorf`

services/transactionpool/service.go
```diff
@@ -27,12 +27,18 @@ func NewTransactionPool(gossip gossiptopics.TransactionRelay, reporting instrume
...
-	s.gossip.BroadcastForwardedTransactions(&gossiptopics.ForwardedTransactionsInput{
-		Message: &gossipmessages.ForwardedTransactionsMessage{
-			SignedTransactions: []*protocol.SignedTransaction{input.SignedTransaction},
-		},
-	})
+	err := validateTransaction(input.SignedTransaction)
+	if err != nil {
+		s.reporting.Info(fmt.Sprintf("transaction is invalid [%v]", input.SignedTransaction))
+		return nil, err
+	} else {
```
electricmonk: as much as I hate it, we decided to not use `else` after `return`, so we'd better stick to the style

services/transactionpool/test/add.go
```diff
@@ -46,4 +46,19 @@ var _ = Describe("transaction pool", func() {
 
 	})
 
+	It("dont forward an invalid transaction with gossip", func() {
```
electricmonk: missing a `Describe` clause here. Also, `It("doesn't forward...")`

services/transactionpool/test/add.go
```diff
@@ -46,4 +46,19 @@ var _ = Describe("transaction pool", func() {
...
+
+		tx := test.TransferTransaction().WithInvalidContent().Build()
+
+		gossip.When("BroadcastForwardedTransactions", mock.Any).Return(&gossiptopics.EmptyOutput{}, nil).Times(0)
```
electricmonk: I'm not sure if Times(0) actually does anything. Did you see the test fail?

services/transactionpool/test/add.go
```diff
@@ -46,4 +46,19 @@ var _ = Describe("transaction pool", func() {
...
+
+		_, err := service.AddNewTransaction(&services.AddNewTransactionInput{
+			SignedTransaction: tx,
+		})
+
+		Expect(err).To(HaveOccurred())
+		Expect(gossip).To(test.ExecuteAsPlanned())
+
+	})
+
```
electricmonk: For the sake of completeness, I'd also add a happy path test here. Seems a bit lonely now :-(

test/harness/network.go
electricmonk: Wait for 3 :)

test/transactions.go
```diff
@@ -12,6 +16,7 @@ func TransferTransaction() *transferTransaction {
 			Transaction: &protocol.TransactionBuilder{
 				ContractName: "BenchmarkToken",
 				MethodName:   "transfer",
+				Timestamp:    primitives.Timestamp(time.Now().UnixNano()),
```
electricmonk: see orbs-network/orbs-spec/issues/49, the name of the field does not reflect its meaning. What does the timestamp represent? date it was created? added to the system? in which timezone? the end user's? ours? utc?

services/transactionpool/test/add.go
erankirsh: Yup. If it happens more than 0 times the test fails

test/transactions.go
erankirsh: Already asked @OdedWx to change to TimestampNano

services/transactionpool/errors.go
talkol: I disagree, look at go's source code, error types are defined like we define them

also, using fmt.Errorf effectively makes all errors strings. Since we want to add extra info to specific errors (like return codes that need to be forwarded to the user), we need rich types and can't rely on strings

services/transactionpool/service.go
talkol: Kiss to @electricmonk !

services/transactionpool/test/add.go
erankirsh: This is the first test...

services/transactionpool/errors.go
erankirsh: Let's discuss offline and make a decision

test/transactions.go
electricmonk: I don't think this is good enough. I'd like the field name to reflect the meaning as well as the units. For instance: `dateCreatedInNanos`. 

BTW, we (@jlevison, @netoneko and myself) assumed this field was in millis!

services/transactionpool/errors.go
electricmonk: I also prefer error types but we need to be consistent with ourselves.

PR #18, opened 2018-07-19
Replaced Loop Control with latch on gossip message types

electricmonk:
I was never happy with LoopControl (even though it's cool! :-) ) as it's not really in par with the concepts of black-box testing. Today I figured out we can achieve the same behavior with a latch on transport messages, waiting until at least one PRE_PREPARE message was sent, in the case of our consensus acceptance test.

Along the way, I also added some missing reporting messages in Block Storage

### DIFF
--- a/bootstrap/node.go
+++ b/bootstrap/node.go
@@ -35,8 +35,7 @@ func NewNode(
 	blockPersistence := blockStorageAdapter.NewLevelDbBlockPersistence(nodeConfig)
 	stateStorageAdapter := stateStorageAdapter.NewLevelDbStatePersistence(nodeConfig)
 	logger := instrumentation.NewStdoutLog()
-	loopControl := instrumentation.NewSimpleLoop(logger)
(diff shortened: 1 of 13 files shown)


services/consensusalgo/leanhelix/algorithm.go
```diff
@@ -7,10 +7,13 @@ import (
 	"github.com/orbs-network/orbs-spec/types/go/protocol/gossipmessages"
 	"github.com/orbs-network/orbs-spec/types/go/services"
 	"github.com/orbs-network/orbs-spec/types/go/services/gossiptopics"
+	"fmt"
 )
 
 func (s *service) leaderProposeNextBlockIfNeeded() error {
-	if s.blocksForRounds[s.lastCommittedBlockHeight+1] != nil {
+	nextBlockHeight := s.lastCommittedBlockHeight + 1
+
```
erankirsh: suggest to change nextBlockHeight to proposedBlockHeight. 
I think we should avoid next and previous in general because they are confusing.

PR #19, opened 2018-07-19
Feature/talkol/transport broadcast self

talkol:
It started with just stopping broadcast to send to self

This caused a big refactor in the consensus

Which caused a big refactor in its implementation to make it pretty after the tests passed

### DIFF
--- a/services/consensusalgo/leanhelix/algorithm.go
+++ b/services/consensusalgo/leanhelix/algorithm.go
@@ -9,12 +9,8 @@ import (
 	"github.com/orbs-network/orbs-spec/types/go/services/gossiptopics"
 )
 
-func leaderProposeNextBlock(
-	transactionPool services.TransactionPool,
-	lastCommittedBlockHeight primitives.BlockHeight,
-) (*protocol.BlockPairContainer, error) {
-
-	proposedTransactions, err := transactionPool.GetTransactionsForOrdering(&services.GetTransactionsForOrderingInput{
+func (s *service) leaderProposeNextBlock() (*protocol.BlockPairContainer, error) {
+	proposedTransactions, err := s.transactionPool.GetTransactionsForOrdering(&services.GetTransactionsForOrderingInput{
 		MaxNumberOfTransactions: 1,
 	})
 	if err != nil {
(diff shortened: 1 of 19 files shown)


services/consensusalgo/leanhelix/algorithm.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"github.com/orbs-network/orbs-network-go/services/blockstorage"
+	"github.com/orbs-network/orbs-spec/types/go/primitives"
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+	"github.com/orbs-network/orbs-spec/types/go/protocol/gossipmessages"
+	"github.com/orbs-network/orbs-spec/types/go/services"
+	"github.com/orbs-network/orbs-spec/types/go/services/gossiptopics"
+)
+
+func (s *service) leaderProposeNextBlockIfNeeded() error {
```
electricmonk: this is very nice, I like the properly named functions. Please add `reporting.info` calls in all functions so that we have better instrumentation of the algorithm in action

services/gossip/service.go
```diff
@@ -74,15 +76,15 @@ func (s *service) SendLeanHelixPrePrepare(input *gossiptopics.LeanHelixPrePrepar
 		RecipientMode: gossipmessages.RECIPIENT_LIST_MODE_BROADCAST,
 		Topic:         gossipmessages.HEADER_TOPIC_LEAN_HELIX,
 		LeanHelix:     gossipmessages.LEAN_HELIX_PRE_PREPARE,
-		NumPayloads:   1,
+		NumPayloads:   uint32(len(input.Message.BlockPair.TransactionsBlock.SignedTransactions)),
```
electricmonk: I think there's a bug here that was here before. Num of payloads should be `len(txs) + 1`, as the block header is the first payload, no?

services/transactionpool/service.go
```diff
@@ -33,8 +33,11 @@ func (s *service) AddNewTransaction(input *services.AddNewTransactionInput) (*se
 			SignedTransactions: []*protocol.SignedTransaction{input.SignedTransaction},
 		},
 	})
+
 	//This is commented out because currently transport broadcast will also broadcast to myself. So HandleForwardedTransactions will be the on to add this transaction.
```
electricmonk: LOL, you removed the commented code but not the comment. This is why comments not inline with the commented code are evil

test/acceptance/simple_transfer.go
```diff
@@ -19,9 +20,13 @@ var _ = Describe("a leader node", func() {
 		network.SendTransfer(0, 1000000) //this is invalid because currently we don't allow (temporarily) values over 1000 just so that we can create invalid transactions
 		network.SendTransfer(0, 22)
 
+		fmt.Println("\nWaiting for node 0 blocks")
```
electricmonk: please remove the prints

PR #20, opened 2018-07-22
Base58 + Address crypto package

jlevison:
- Base58 implementation (base on trezor code, refactored from c->go)
- Address struct according to the cryptosdk (lib) implementation, in native go

### DIFF
--- a/crypto/address/address.go
+++ b/crypto/address/address.go
@@ -111,8 +111,10 @@ func NewFromRaw(a []byte, publicKey []byte) (*Address, error) {
 
 	cs := binary.BigEndian.Uint32(a[NETWORK_ID_SIZE+VERSION_SIZE+VCHAIN_ID_SIZE+ACCOUNT_ID_SIZE:])
 	if _, err := newAddress.generateFullAddress(); err != nil {
-		return nil, fmt.Errorf("f/ailed to generate full address for new address during checksum test")
-	} else if actualCs, err := newAddress.Checksum(); err != nil {
+		return nil, fmt.Errorf("failed to generate full address for new address during checksum test")
+	}
+
+	if actualCs, err := newAddress.Checksum(); err != nil {
 		return nil, fmt.Errorf("failed to generate full address for new address during checksum test")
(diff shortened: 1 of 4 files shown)


crypto/address/address.go
```diff
@@ -0,0 +1,298 @@
...
+	NETWORK_ID_SIZE      = 1
+	VERSION_SIZE         = 1
+	ACCOUNT_ID_SIZE      = 20
+	CHECKSUM_SIZE        = 4
+	PUBLIC_KEY_SIZE      = 32
+)
+
+func validateAddressParts(pk []byte, vcId []byte, net string, v uint8) error {
+	if len(pk) != PUBLIC_KEY_SIZE {
+		return fmt.Errorf("pk invalid, cannot create address")
```
electricmonk: we need to be consistent with regards to error types. So far, @jlevison is for using `Errorf` while @talkol is for using concrete types.

Regardless, please add the offending field to the error string so that caller knows what they sent wrong.

crypto/address/address.go
```diff
@@ -0,0 +1,298 @@
...
+	if aid, err := newAddress.AccountId(); err != nil {
+		return nil, fmt.Errorf("unable to create account id for new address")
+	} else if !bytes.Equal(aid, accountId) {
+		return nil, fmt.Errorf("account id mismatch, invalid address")
+	}
+
+	cs := binary.BigEndian.Uint32(a[NETWORK_ID_SIZE+VERSION_SIZE+VCHAIN_ID_SIZE+ACCOUNT_ID_SIZE:])
+	if _, err := newAddress.generateFullAddress(); err != nil {
+		return nil, fmt.Errorf("f/ailed to generate full address for new address during checksum test")
+	} else if actualCs, err := newAddress.Checksum(); err != nil {
```
electricmonk: styling issue - if we're being idiomatic, `else` after `return` is redundant

crypto/address/address.go
jlevison: added errors.Wrap where needed

PR #23, opened 2018-07-23
Feature/talkol/benchmark consensus create

talkol:
Initial creation and a lot of refactoring


services/gossip/codec_test.go
```diff
@@ -0,0 +1,105 @@
+package gossip
+
+import (
+	"github.com/google/go-cmp/cmp"
+	"github.com/orbs-network/orbs-network-go/test"
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+	"testing"
+)
+
+var blockPairTable = []struct {
```
electricmonk: really, really hard to figure out what this means
Please create a factory method, or a builder-of-builders, so that a reader can understand from a glance at the table what is expected. 
In addition, consider adding an explanation (string) to the errors (both encode and decode) so that the reader can understand *why* we expect any specific entry in the table to fail

services/gossip/topic_benchmark_consensus.go
```diff
@@ -0,0 +1,98 @@
+package gossip
```
electricmonk: nice! this is a very good feature of Golang you're making use of here, the ability to split a "class" to multiple files by context

services/gossip/codec.go
```diff
@@ -0,0 +1,82 @@
+package gossip
+
+import (
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+)
+
+func encodeBlockPair(blockPair *protocol.BlockPairContainer) ([][]byte, error) {
```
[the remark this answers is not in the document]
talkol: hmm.. not sure it's a huge chunk of memory because it's only pointers. Let's assume you have a block with 1000 transactions and 1000 transaction receipts. Each transaction can be 1KB so the block itself will be over 1MB in size. But the memory chunk here will only be 2000*pointersize = 8KB-16KB which isn't too bad..

I do agree that we will eventually need to move to a Writer instead of []byte.. but I'll leave that to the gossip team :)

services/gossip/topic_benchmark_consensus.go
talkol: good feedback from Shai! achievement unlocked!

services/gossip/codec_test.go
talkol: yes.. I thought 10 times before adding this unit test.. I'm not in the gossip team but I needed to encode and decode a full block in gossip for my consensus implementation and I just couldn't leave this without a unit test

I'm wondering if we should leave this to them to improve the way the codec is written.. or write it according to my own liking..

talkol: I'll refactor the test and add builders, hold on

talkol:
We have a new config value which controls which consensus algo is active on this virtual chain.

Fixed the codec table test with builders.

services/gossip/codec.go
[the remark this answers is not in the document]
electricmonk: I don't understand your math. If the outcome of this method is a fully-serialized block-pair to be transmitted over TCP, it will be 1MB. What am I missing?

services/gossip/topic_benchmark_consensus.go
electricmonk: :)

services/consensusalgo/benchmarkconsensus/test/context.go
```diff
@@ -0,0 +1,68 @@
...
+	isLeader bool,
+) *context {
+
+	leaderPublicKey := []byte{0x01}
+	nodePublicKey := leaderPublicKey
+	if !isLeader {
+		nodePublicKey = []byte{0x02}
+	}
+
+	config := config.NewHardCodedConfig(
```
electricmonk: Please don't instantiate `HardCodecConfig` here; the consensus algo needs a subset of that config so better to implement a stub struct in the test to implement that interface.

services/consensusalgo/benchmarkconsensus/test/context.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/orbs-network/orbs-network-go/config"
+	"github.com/orbs-network/orbs-network-go/instrumentation"
+	"github.com/orbs-network/orbs-network-go/services/consensusalgo/benchmarkconsensus"
+	testInstrumentation "github.com/orbs-network/orbs-network-go/test/harness/instrumentation"
+	"github.com/orbs-network/orbs-spec/types/go/protocol/consensus"
+	"github.com/orbs-network/orbs-spec/types/go/services"
+	"github.com/orbs-network/orbs-spec/types/go/services/gossiptopics"
+)
+
+type context struct {
```
electricmonk: let's sit tomorrow morning and refactor this. It's really the wrong direction IMO. Look at the test for Block Storage for what I'd like to do (test driver rather than dumb context): https://github.com/orbs-network/orbs-network-go/blob/master/services/blockstorage/test/commit_block.go

services/consensusalgo/benchmarkconsensus/test/create_blocks_test.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil {
+		t.Fatal("Did not create block with ConsensusContext:", err)
+	}
+}
+
+func TestNonLeaderDoesNotCreateBlocks(t *testing.T) {
+	c := newContext(false)
+	c.consensusContext.When("RequestNewTransactionsBlock", mock.Any).Return(nil, nil).Times(0)
+	c.createService()
+	err := test.ConsistentlyVerify(c.consensusContext)
```
electricmonk: `ConsistentlyVerify`? is this your implementation of 'eventually'?

services/consensusalgo/leanhelix/service.go
```diff
@@ -49,8 +51,8 @@ func NewLeanHelixConsensusAlgo(
 	}
 
 	gossip.RegisterLeanHelixHandler(s)
-	if config.ConstantConsensusLeader().Equal(config.NodePublicKey()) {
-		go s.buildBlocksEventLoop()
+	if config.ActiveConsensusAlgo() == consensus.CONSENSUS_ALGO_TYPE_LEAN_HELIX && config.ConstantConsensusLeader().Equal(config.NodePublicKey()) {
```
electricmonk: now it's time to extract this to a function `isEnabled` :)

services/consensusalgo/leanhelix/service.go
```diff
@@ -103,8 +105,8 @@ func (s *service) buildBlocksEventLoop() {
 		}
 
 		// validate the current proposed block
-		if s.blocksForRounds[s.lastCommittedBlockHeight + 1] != nil {
-			err := s.leaderCollectVotesForBlock(s.blocksForRounds[s.lastCommittedBlockHeight + 1])
+		if s.blocksForRounds[s.lastCommittedBlockHeight+1] != nil {
```
electricmonk: why are you removing the spaces around the plus sign?

services/gossip/codec.go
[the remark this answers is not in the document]
talkol: No, the entire point of having unserialiazable containers is not to have a single chunk of 1MB to be transmitted over TCP. We are creating a `[][]byte` meaning an array of []byte chunks. This is the chunked array you talked about. Every chunk is already serialized (before arriving to the gossip) but sits somewhere else in memory. We don't serialize them together, we just take a bunch of pointers and we transmit every chunk separately.

This model can be changed to a Writer very easily. This is a great task for the gossip team, just change the interface of the transport from a `[][]byte` to a Writer. Pretty easy to do and does not require any architecture changes, it's just an implementation detail of gossip which is great.

services/consensusalgo/benchmarkconsensus/test/context.go
talkol: Yeah.. but this is super annoying. Because I need to implement pretty much HardcodedConfig from scratch. We need some way to create partials of HardcodedConfig.. Maybe create it using a struct instead of a ctor and then I can pass only what I want to initialize and it will initialize the rest to nil. I don't want every component to have a boilerplate implementation similar to HardcodedConfig.

services/consensusalgo/leanhelix/service.go
talkol: Not me, it's `go fmt`. I'm running it automatically after every save.

services/consensusalgo/benchmarkconsensus/test/create_blocks_test.go
talkol: There are 2 functions - `eventually` and `consistently`. Ginkgo has both in its API and it turned out I needed both. I also added specialized cases for mocks to reduce boilerplate. The full code is here: https://github.com/orbs-network/orbs-network-go/pull/23/files#diff-9668350988ac1f28905c796e007c35e5

services/consensusalgo/benchmarkconsensus/test/context.go
talkol: I'm letting the context emerge from my tests in TDD :)
I didn't want to create it with a special direction in mind, what you see is the result of refactoring from test to test. I want to see where the next tests take it

services/blockstorage/service.go
```diff
@@ -86,7 +86,7 @@ func (s *service) ValidateBlockForCommit(input *services.ValidateBlockForCommitI
 }
 
 func (s *service) RegisterConsensusBlocksHandler(handler handlers.ConsensusBlocksHandler) {
-	panic("Not implemented")
+	s.consensusBlocksHandlers = append(s.consensusBlocksHandlers, handler)
```
erankirsh: @talkol Why did you need to implement this? At any point (now) there's only one consensus active.

services/gossip/codec.go
```diff
@@ -0,0 +1,82 @@
...
+	if blockPair == nil || blockPair.TransactionsBlock == nil || blockPair.ResultsBlock == nil {
+		return nil, &ErrCodecEncode{"BlockPair", blockPair}
+	}
+	payloads := make([][]byte, 0, 5+
```
erankirsh: @talkol Can you please add a comment here for where the 5+ is coming from

PR #25, opened 2018-07-23
More tests for block storage

netoneko:


services/blockstorage/service.go
```diff
@@ -82,7 +162,15 @@ func (s *service) GetLastCommittedBlockHeight(input *services.GetLastCommittedBl
 }
 
 func (s *service) ValidateBlockForCommit(input *services.ValidateBlockForCommitInput) (*services.ValidateBlockForCommitOutput, error) {
-	panic("Not implemented")
+	if protocolVersionError := s.validateProtocolVersion(input.BlockPair); protocolVersionError != nil {
+		return &services.ValidateBlockForCommitOutput{}, protocolVersionError
```
jlevison: why not return nil on error (we are working with pointers here, it should work)

services/blockstorage/service.go
```diff
@@ -120,13 +208,37 @@ func (s *service) validateBlockDoesNotExist(txBlockHeader *protocol.Transactions
...
 
-func (s *service) validateProtocolVersion(txBlockHeader *protocol.TransactionsBlockHeader) error {
+func (s *service) validateBlockHeight(blockPair *protocol.BlockPairContainer) error {
+	expectedBlockHeight := s.lastCommittedBlockHeight + 1
+
+	txBlockHeader := blockPair.TransactionsBlock.Header
+	rsBlockHeader := blockPair.ResultsBlock.Header
+
+	if txBlockHeader.BlockHeight() != expectedBlockHeight {
+		return fmt.Errorf("block height is %d, expected %d", txBlockHeader.BlockHeight(), expectedBlockHeight)
```
jlevison: add which block to the message (tx/rs)

services/blockstorage/service.go
```diff
@@ -120,13 +208,37 @@ func (s *service) validateBlockDoesNotExist(txBlockHeader *protocol.Transactions
...
+
+	return nil
+}
+
+func (s *service) validateProtocolVersion(blockPair *protocol.BlockPairContainer) error {
+	txBlockHeader := blockPair.TransactionsBlock.Header
+	rsBlockHeader := blockPair.ResultsBlock.Header
+
 	if txBlockHeader.ProtocolVersion() != ProtocolVersion {
 		err := fmt.Errorf("protocol version mismatch: expected 1 got %d", txBlockHeader.ProtocolVersion())
```
jlevison: change expected 1 to expected %d and use the constant (this is actually my mistake from last week i think)

services/blockstorage/service.go
```diff
@@ -120,13 +208,37 @@ func (s *service) validateBlockDoesNotExist(txBlockHeader *protocol.Transactions
...
 		s.reporting.Error(err)
 		return err
 	}
 
+	if rsBlockHeader.ProtocolVersion() != ProtocolVersion {
+		return fmt.Errorf("protocol version mismatch: expected 1 got %d", rsBlockHeader.ProtocolVersion())
```
jlevison: same as above (copy paste, right?)

services/blockstorage/service.go
```diff
@@ -62,12 +63,91 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
...
+		return result, nil
+	}
+
+	return s.lookForTransactionsBlockHeader(input.BlockHeight)
 }
 
-func (s *service) GetResultsBlockHeader(input *services.GetResultsBlockHeaderInput) (*services.GetResultsBlockHeaderOutput, error) {
-	panic("Not implemented")
+func (s *service) lookForResultsBlockHeader(height primitives.BlockHeight) (*services.GetResultsBlockHeaderOutput, error) {
+	txBlock, err := s.persistence.GetResultsBlock(height)
```
jlevison: styling, maybe change this to
if x,err := action(); err != nil

services/blockstorage/service.go
```diff
@@ -62,12 +63,91 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
 	return nil, nil
 }
 
+func (s *service) lookForTransactionsBlockHeader(height primitives.BlockHeight) (*services.GetTransactionsBlockHeaderOutput, error) {
+	txBlock, err := s.persistence.GetTransactionsBlock(height)
```
jlevison: styling, maybe change this to
if x,err := action(); err != nil

jlevison: also, not sure here, but if the GetTranBlock(h) is only used here, maybe change the return object to match this function (a bit lost in the API when looking at it like this)

services/blockstorage/service.go
```diff
@@ -62,12 +63,91 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
 	return nil, nil
 }
 
+func (s *service) lookForTransactionsBlockHeader(height primitives.BlockHeight) (*services.GetTransactionsBlockHeaderOutput, error) {
```
electricmonk: why `lookFor...`?

services/blockstorage/service.go
```diff
@@ -62,12 +63,91 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
...
+	}
+
+	return &services.GetTransactionsBlockHeaderOutput{
+		TransactionsBlockProof: txBlock.BlockProof,
+		TransactionsBlockHeader: txBlock.Header,
+		TransactionsBlockMetadata: txBlock.Metadata,
+	}, nil
+}
+
 func (s *service) GetTransactionsBlockHeader(input *services.GetTransactionsBlockHeaderInput) (*services.GetTransactionsBlockHeaderOutput, error) {
```
electricmonk: a method that opens a long-lived goroutine might need to take a `Context` argument to support cancellation (for instance, on shutdown / test end). See PR #29. This would mean changing the API, but I think this might be required anyway. @talkol @OdedWx thoughts?

services/blockstorage/service.go
```diff
@@ -62,12 +63,91 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
...
+		ResultsBlockProof: txBlock.BlockProof,
+		ResultsBlockHeader: txBlock.Header,
+	}, nil
+}
+
+func (s *service) GetResultsBlockHeader(input *services.GetResultsBlockHeaderInput) (result *services.GetResultsBlockHeaderOutput, err error) {
+	if input.BlockHeight > s.lastCommittedBlockHeight && input.BlockHeight - s.lastCommittedBlockHeight <= 5 {
+		c := make(chan *services.GetResultsBlockHeaderOutput)
+
+		go func() {
```
electricmonk: this is the exact same functionality as in `GetTransactionsBlockHeader`, right? if so, consider extracting to a common local function that supports this "wait till block is committed" behavior

services/blockstorage/service.go
```diff
@@ -62,12 +63,91 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
...
+			// TODO extract to a config
+			const interval = 10
+			const timeout = 10000
+
+			for i:=0; i < timeout; i+= interval {
+				if input.BlockHeight <= s.lastCommittedBlockHeight {
+					lookupResult, err := s.lookForTransactionsBlockHeader(input.BlockHeight)
+
+					if err == nil {
```
electricmonk: is this good enough? I think there should be a distinction between error types. Should all errors be retried?

services/blockstorage/test/component_test.go
```diff
@@ -4,9 +4,71 @@ import (
...
+	out, err := d.blockStorage.GetLastCommittedBlockHeight(&services.GetLastCommittedBlockHeightInput{})
+	Expect(err).ToNot(HaveOccurred())
+	return out
+}
+
+func (d *driver) getBlock(height int) *protocol.BlockPairContainer {
+	return d.storageAdapter.ReadAllBlocks()[height - 1]
+}
+
+func NewDriver() *driver {
```
electricmonk: why is this public? the tests files are in the same package as the driver

services/blockstorage/test/get_block_headers.go
```diff
@@ -0,0 +1,105 @@
...
+			driver := NewDriver()
+			driver.expectCommitStateDiff()
+
+			block := test.BlockPairBuilder().Build()
+			driver.commitBlock(block)
+
+			output, err := driver.blockStorage.GetTransactionsBlockHeader(&services.GetTransactionsBlockHeaderInput{BlockHeight:1})
+
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output.TransactionsBlockHeader).To(Equal(block.TransactionsBlock.Header))
```
electricmonk: make sure that your persistence adapter (in-memory) does not hold a pointer to the block but rather clones its bytes, so this equality check actually checks content and not just pointers

services/blockstorage/test/get_block_headers.go
```diff
@@ -0,0 +1,105 @@
...
+var _ = Describe("Block storage", func () {
+	When("asked to provide transactions block header", func () {
+		It("returns transactions block header", func () {
+			driver := NewDriver()
+			driver.expectCommitStateDiff()
+
+			block := test.BlockPairBuilder().Build()
+			driver.commitBlock(block)
+
+			output, err := driver.blockStorage.GetTransactionsBlockHeader(&services.GetTransactionsBlockHeaderInput{BlockHeight:1})
```
electricmonk: consider hiding the input message struct in the driver, so that instead this could look like:
```golang
driver.getTxBlockHeader(1)
```

services/blockstorage/test/get_block_headers.go
```diff
@@ -0,0 +1,105 @@
...
+			Expect(output.TransactionsBlockMetadata).To(Equal(block.TransactionsBlock.Metadata))
+			Expect(output.TransactionsBlockProof).To(Equal(block.TransactionsBlock.BlockProof))
+		})
+
+		It("blocks if requested block is in near future", func (done Done) {
+			driver := NewDriver()
+			driver.expectCommitStateDiff()
+
+			block := test.BlockPairBuilder().Build()
+			driver.commitBlock(block)
```
electricmonk: why do you need the first commit?

services/blockstorage/test/get_block_headers.go
```diff
@@ -0,0 +1,105 @@
...
+				result <- output
+			}()
+
+			for i:=2; i <=6 ; i++ {
+				driver.commitBlock(test.BlockPairBuilder().WithHeight(i).Build())
+			}
+
+			Expect(driver.getLastBlockHeight().LastCommittedBlockHeight).To(Equal(primitives.BlockHeight(6)))
+
+			block5 := driver.getBlock(5).TransactionsBlock
```
electricmonk: why do you need the comparison again? the added functionality here is to wait until txBlock is ready, so I'd only check that the call returns something, no need to assert on content again

services/blockstorage/test/get_block_headers.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/orbs-network/orbs-network-go/test"
+	"github.com/orbs-network/orbs-spec/types/go/services"
+	"github.com/orbs-network/orbs-spec/types/go/primitives"
+)
+
+var _ = Describe("Block storage", func () {
+	When("asked to provide transactions block header", func () {
```
electricmonk: * add a test for timeout if block is not persisted
* add a test for block not in the near future
both of these will force you to extract configuration (as you don't want, for instance, to make the test wait 10 seconds)

services/blockstorage/test/get_block_headers.go
```diff
@@ -0,0 +1,105 @@
...
+			driver.commitBlock(block)
+
+			output, err := driver.blockStorage.GetResultsBlockHeader(&services.GetResultsBlockHeaderInput{BlockHeight:1})
+
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output.ResultsBlockHeader).To(Equal(block.ResultsBlock.Header))
+			Expect(output.ResultsBlockProof).To(Equal(block.ResultsBlock.BlockProof))
+		})
+
+		It("blocks if requested block is in near future", func (done Done) {
```
electricmonk: note that this test is *exactly* like the the test for `GettTxBlockHeader`
this hints to me that the "blocking the call until condition is met" should be a separate unit with its own unit test, rather than a component test. WDYT?

services/blockstorage/test/validation.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/orbs-network/orbs-spec/types/go/services"
+)
+
+var _ = Describe("Block storage", func () {
+	When("asked to validate transactions block", func() {
+		It("checks protocol version", func() {
+			driver := NewDriver()
+			block := test.BlockPairBuilder().Build()
+
+			_, err := driver.blockStorage.ValidateBlockForCommit(&services.ValidateBlockForCommitInput{block})
```
electricmonk: again, please consider pushing the boilerplate to the driver

services/blockstorage/service.go
netoneko: We retry until it times out, which is expected behavior from the spec.

services/blockstorage/service.go
netoneko: because `GetTransactionsBlockHeader` and `getTransactionsBlockHeader` would be confusing

services/blockstorage/service.go
electricmonk: I'm not sure this is well-defined in the spec. Should you retry on unrecoverable errors?

services/blockstorage/service.go
electricmonk: `load`?

services/blockstorage/test/get_block_headers.go
netoneko: I have a use case that's most likely to happen: read from the end of the initialized chain.

services/blockstorage/test/get_block_headers.go
netoneko: What if it waits and returns the last available block and not the one we requested? Or some other block?

electricmonk: cool, but you only need to assert that the block that returned has the expected height, you don't need to assert on content

services/blockstorage/service.go
netoneko: I prefer not to have null pointer dereference

services/blockstorage/service.go
netoneko: All of it will change when we merge #33

services/blockstorage/service.go
netoneko: I have introduced the config to fix this.

PR #31, opened 2018-07-24
Feature/talkol/benchmark consensus continued

talkol:


services/consensusalgo/benchmarkconsensus/test/harness.go
```diff
@@ -71,3 +76,82 @@ func (h *harness) createService(ctx context.Context) {
...
+func (h *harness) receiveCommit(blockPair *protocol.BlockPairContainer) {
+	h.service.HandleBenchmarkConsensusCommit(&gossiptopics.BenchmarkConsensusCommitInput{
+		Message: &gossipmessages.BenchmarkConsensusCommitMessage{
+			BlockPair: blockPair,
+		},
+	})
+}
+
+func (h *harness) expectIgnoreCommit() {
+	h.blockStorage.Reset().When("CommitBlock", mock.Any).Return(nil, nil).Times(0)
```
electricmonk: why do you need to call Reset()? this is usually a smell...

services/consensusalgo/benchmarkconsensus/test/harness.go
```diff
@@ -71,3 +76,82 @@ func (h *harness) createService(ctx context.Context) {
...
+		t.Fatal("Did not register with Gossip:", err)
+	}
+	ok, err = h.blockStorage.Verify()
+	if !ok {
+		t.Fatal("Did not register with BlockStorage:", err)
+	}
+}
+
+func (h *harness) expectBlockCreation() {
+	h.consensusContext.Reset().When("RequestNewTransactionsBlock", mock.Any).Return(nil, nil).AtLeast(1)
```
electricmonk: the name of the function doesn't reflect what it does. It doesn't actually expect that a block was created, right? it expects the algo to ask for a new transaction block... I'd name it `expectRequestNewTransactionBlock()`.

in addition, maybe return the `MockFunction` so that I could do:
```golang
h.expectRequestNewTransactionBlock().AtLeast(1)
h.expectRequestNewTransactionBlock().Times(0)

// etc
```

services/consensusalgo/benchmarkconsensus/test/leader_test.go
```diff
@@ -2,19 +2,15 @@ package test
...
-	"github.com/orbs-network/go-mock"
 	"github.com/orbs-network/orbs-network-go/test"
 	"testing"
 )
 
 func TestLeaderCreatesBlocks(t *testing.T) {
 	test.WithContext(func(ctx context.Context) {
 		h := newHarness(true)
-		h.consensusContext.Reset().When("RequestNewTransactionsBlock", mock.Any).Return(nil, nil).AtLeast(1)
+		h.expectBlockCreation()
```
electricmonk: now it seems like you pushed too much behavior into the driver and it's no longer clear from reading the test what exactly is the expected behavior

services/consensusalgo/benchmarkconsensus/test/harness.go
```diff
@@ -71,3 +76,82 @@ func (h *harness) createService(ctx context.Context) {
...
+}
+
+func (h *harness) verifyBlockCreation(t *testing.T) {
```
electricmonk: This doesn't, strictly, verify that blocks where created. It verifies that eventually all expectations from the mock were met. 

Please consider renaming this as well as the counterpart `verifyNoBlockCreation`

services/consensusalgo/benchmarkconsensus/test/non_leader_test.go
```diff
@@ -2,39 +2,135 @@ package test
...
+
+		var replyBlockHeight primitives.BlockHeight
+
+		h.expectCommitSaveAndReply(&replyBlockHeight)
+		b1 := builders.BlockPair().
+			WithHeight(1).
+			WithBenchmarkConsensusBlockProof(nil, h.config.ConstantConsensusLeader()). //TODO: fix private key
+			Build()
+		h.receiveCommit(b1)
+		if replyBlockHeight != 1 {
```
electricmonk: please add `expected block height %d but got %d`

services/consensusalgo/benchmarkconsensus/test/non_leader_test.go
```diff
@@ -2,39 +2,135 @@ package test
...
-		err := test.ConsistentlyVerify(h.consensusContext)
-		if err != nil {
-			t.Fatal("Did create block with ConsensusContext:", err)
-		}
+		h.verifyNoBlockCreation(t)
 	})
 }
 
-func TestNonLeaderIgnoreFutureBlockHeight(t *testing.T) {
+func TestNonLeaderCommitsAndRepliesToValidBlockHeights(t *testing.T) {
```
electricmonk: I don't like this test:
* I don't understand why you need to assert on 3 blocks
* It takes time to understand that you're asserting on a long-going story
* Maybe this needs to be a unit test?

services/gossip/codec.go
```diff
@@ -8,7 +8,7 @@ func encodeBlockPair(blockPair *protocol.BlockPairContainer) ([][]byte, error) {
 	if blockPair == nil || blockPair.TransactionsBlock == nil || blockPair.ResultsBlock == nil {
 		return nil, &ErrCodecEncode{"BlockPair", blockPair}
 	}
-	payloads := make([][]byte, 0, 5+
+	payloads := make([][]byte, 0, 5+ // the 5 is coming from the 5 payloads we're always sending first (txHeader, txMetadata...)
```
electricmonk: instead of the comment, set 5 to a constant `sizeOfHeaderAndMetadataAndFooAndBarAndBaz`, this is more self explanatory and will make it easier for people to know they need to change it if the number of fixed payloads change

services/consensusalgo/benchmarkconsensus/test/non_leader_test.go
```diff
@@ -2,39 +2,135 @@ package test
...
+			WithBenchmarkConsensusBlockProof(nil, h.config.ConstantConsensusLeader()). //TODO: fix private key
+			Build()
+		h.receiveCommit(b1)
+		if replyBlockHeight != 1 {
+			t.Fatalf("Replied committed with wrong last block height: %d", replyBlockHeight)
+		}
+		h.verifyCommitSaveAndReply(t)
+
+		h.expectCommitSaveAndReply(&replyBlockHeight)
+		b2 := builders.BlockPair().
```
electricmonk: maybe the builder can have a `ThatIsNextOf(prevBlock)` method, that automatically sets height and prev block hash

test/eventually_test.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+func TestEventuallyVerifySuccessWithTwoMocks(t *testing.T) {
+	p1 := &personMock{}
+	p1.When("GetName").Return("john").Times(1)
+	p2 := &personMock{}
+	p2.When("GetName").Return("smith").Times(1)
+	go func() {
+		time.Sleep(1 * time.Millisecond)
```
electricmonk: consider controlling the flow in this test using stuff in the `sync` package rather than using sleeps, as this is fragile

services/consensusalgo/benchmarkconsensus/test/harness.go
talkol: Some of these mocks are required for construction (like the register). This means the harness always sets them just so it won't panic. If the test explicitly wants to set them in order to verify how they were called I preferred to reset so they won't accumulate. I'm basically replacing the default behavior of the harness here

services/consensusalgo/benchmarkconsensus/test/leader_test.go
talkol: Maybe I should return the mock function like you suggested in the last comment and this will move this behavior back to the test

services/consensusalgo/benchmarkconsensus/test/harness.go
talkol: Good idea

services/consensusalgo/benchmarkconsensus/test/harness.go
talkol: I'll rename both

services/consensusalgo/benchmarkconsensus/test/non_leader_test.go
talkol: Originally it was two tests:

1. Test that two commits in succession changes the reply

2. Test that an out of order commit (old one) doesn't change the reply

Because test 2 was exactly test 1 with an added block at the end, I removed test 1 to avoid duplication.

I don't want to move this to a unit test because it's part of the component spec

Do you want me to split it back to 2 tests maybe?

services/gossip/codec.go
talkol: Good point

services/consensusalgo/benchmarkconsensus/test/non_leader_test.go
talkol: It does, it's called `WithPrevBlockHash(prevBlock *protocol.BlockPairContainer)`

it takes a previous block and sets the hashes accordingly

test/eventually_test.go
talkol: I'll try

services/consensusalgo/benchmarkconsensus/algorithm.go
```diff
@@ -46,3 +55,83 @@ func (s *service) generateNewProposedBlock() (*protocol.BlockPairContainer, erro
...
+		}
+		prevRxHash := crypto.CalcResultsBlockHash(s.lastCommittedBlock)
+		if !blockPair.ResultsBlock.Header.PrevBlockHashPtr().Equal(prevRxHash) {
+			return errors.Errorf("results prev block hash does not match prev block: %s", prevRxHash)
+		}
+	}
+
+	// block proof
+	if !blockPair.ResultsBlock.BlockProof.IsTypeBenchmarkConsensus() {
+		return errors.Errorf("incorrect block proof type: %s", blockPair.ResultsBlock.BlockProof.Type())
```
erankirsh: Logically this check is suppose to be before the prev block hash ptr

services/consensusalgo/benchmarkconsensus/algorithm.go
```diff
@@ -46,3 +55,83 @@ func (s *service) generateNewProposedBlock() (*protocol.BlockPairContainer, erro
...
+
+func (s *service) nonLeaderCommitAndReply(blockPair *protocol.BlockPairContainer) error {
+	_, err := s.blockStorage.CommitBlock(&services.CommitBlockInput{
+		BlockPair: blockPair,
+	})
+	if err != nil {
+		return err
+	}
+	if blockPair.TransactionsBlock.Header.BlockHeight() > s.lastCommittedBlockHeight() {
+		s.lastCommittedBlock = blockPair
```
erankirsh: What are you checking here? if it's a previous block CommitBlock will return error. If it's a future block it was handled in `nonLeaderValidateBlock` 
Is this for a race I don't see?

test/eventually_test.go
talkol: I only succeeded removing the sleeps from the first two tests. I wasn't successful in removing them from the `EventuallyVerify` and `ConsistentlyVerify` tests that work with mocks

electricmonk: let's review together

test/builders/blocks.go
```diff
@@ -76,6 +80,30 @@ func (b *blockPair) WithHeight(blockHeight primitives.BlockHeight) *blockPair {
 	return b
 }
 
+func (b *blockPair) WithPrevBlockHash(prevBlock *protocol.BlockPairContainer) *blockPair {
```
erankirsh: I suggest we differentiate lean helix block pair from benchmark block pair

services/consensusalgo/benchmarkconsensus/algorithm.go
talkol: done

services/consensusalgo/benchmarkconsensus/algorithm.go
talkol: clarified the code

test/builders/blocks.go
talkol: done, we now have 3 methods:

`LeanHelixBlockPair`
`BenchmarkConsensusBlockPair`
`BlockPair` - if you don't care which one it directs to `LeanHelixBlockPair`.. we can kill this one but then we break lots of code in the system

test/eventually_test.go
talkol: separately from this PR though so it doesn't get stuck

PR #33, opened 2018-07-25
Improved logging

netoneko:
Work in progress (needs README about logging)

@electricmonk please take a look at the API


bootstrap/httpserver/server.go
```diff
@@ -6,10 +6,10 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/orbs-network/orbs-network-go/instrumentation"
+	"github.com/orbs-network/membuffers/go"
+	. "github.com/orbs-network/orbs-network-go/instrumentation"
```
talkol: Please do not import everything from a package, it's bad practice
The only place we allow it is when we import from gomega and ginkgo (and we plan on stop using both)

bootstrap/node.go
```diff
@@ -38,9 +39,10 @@ func NewNode(
 	ctx, ctxCancel := context.WithCancel(context.Background())
 	nodeConfig := config.NewHardCodedConfig(networkSize, nodePublicKey, constantConsensusLeader, activeConsensusAlgo)
 
+	logger := instrumentation.GetLogger(instrumentation.Node(string(base58.Encode([]byte(nodePublicKey)))))
+
```
erankirsh: Why base58? I think Tal added a stringer to those.

instrumentation/basic_meter.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type BasicMeter interface {
+	Done()
+}
+
+func (m *basicMeter) Done() {
+	m.end = time.Now().UnixNano()
+	diff := float64(m.end-m.start) / NanosecondsInASecond
+
```
erankirsh: I don't understand where start is initialized?

instrumentation/formatters.go
```diff
@@ -0,0 +1,139 @@
+package instrumentation
```
erankirsh: I think most of the code here should be redundant since membuffers already have a stringer for all protos (containers, primitives, etc...)

bootstrap/httpserver/server.go
```diff
@@ -34,7 +34,7 @@ func NewHttpServer(address string, reporting instrumentation.Reporting, publicAp
 	}
 
 	go func() {
-		reporting.Infof("starting server on address %s", address)
+		reporting.Info("Starting server on address", String("address", address))
```
erankirsh: I need to understand what we're trying to achieve with have this typed logging.

electricmonk: I think that the idea is structured logging. Not sure that this _specific_ log entry is a representative example, though :)

bootstrap/httpserver/server.go
```diff
@@ -18,11 +18,11 @@ type HttpServer interface {
 
 type server struct {
 	httpServer *http.Server
-	reporting  instrumentation.Reporting
+	reporting  BasicLogger
 	publicApi  services.PublicApi
 }
 
-func NewHttpServer(address string, reporting instrumentation.Reporting, publicApi services.PublicApi) HttpServer {
+func NewHttpServer(address string, reporting BasicLogger, publicApi services.PublicApi) HttpServer {
```
electricmonk: if we follow the same convention as in services, the logger for `server` should be `reporting.For("http-server")`. This would also help with the metering, as the name of the meter would be `request-process-time`, with the `http-server` prefix coming from the parent logger.

instrumentation/basic_meter.go
netoneko: When we call `logger.Meter()`

instrumentation/formatters.go
netoneko: I added support for anything that confirms to `Stringer` via `Stringable(key, value)`

jlevison:
merging so we can start using the logger - there are some more changes that need to be done in regards of metrics and formatting / serializations

PR #35, opened 2018-07-25
Feature/stagestore basic nb

noambergIL:
StateStorage:
keep state of _only_ last successfull block in memory 
funcs:
readkeys (currently disregard block height)
commitstatediff
getstateheight


services/statestorage/service.go
```diff
@@ -1,44 +1,74 @@
...
-	persistence adapter.StatePersistence
+	persistence            adapter.StatePersistence
+	lastResultsBlockHeader *protocol.ResultsBlockHeader
 }
 
 func NewStateStorage(persistence adapter.StatePersistence) services.StateStorage {
 	return &service{
-		persistence: persistence,
+		persistence:            persistence,
+		lastResultsBlockHeader: (&protocol.ResultsBlockHeaderBuilder{}).Build(),
```
erankirsh: Hmmm. What did you try to do here?
I can't imagine you'll need to create a state storage with with empty results block

ronnno: Agree, this does not make much sense. 
Service naturally assumes it as at least one block. Since there is no init process yet, this init simulates an ad-hoc genesis block until we implement Init flow.

I think we have two options here - one is to make all the tests pass even when this is nil. The other is to add a //TODO to make sure this is removed once Init flow is implemented. Do you see a third?

noambergIL:
one needs to be able to return next desired ... so block 0 + 1 :)

services/statestorage/service.go
```diff
@@ -1,44 +1,74 @@
...
 	for _, stateDiffs := range input.ContractStateDiffs {
 		for i := stateDiffs.StateDiffsIterator(); i.HasNext(); {
-			s.persistence.WriteState(i.NextStateDiffs())
+			s.persistence.WriteState(stateDiffs.ContractName(), i.NextStateDiffs())
 		}
 	}
-
-	return nil, nil
+	s.lastResultsBlockHeader = input.ResultsBlockHeader
+	hi := committedBlock + 1
```
erankirsh: What's hi? I guess next desired block height? Suggest to change the name

services/statestorage/service.go
```diff
@@ -1,44 +1,74 @@
...
+	return &services.CommitStateDiffOutput{NextDesiredBlockHeight: hi}, nil
 }
 
 func (s *service) ReadKeys(input *services.ReadKeysInput) (*services.ReadKeysOutput, error) {
-	var state []*protocol.StateRecord
-	for _, stateDiff := range s.persistence.ReadState() {
-		state = append(state, stateDiff)
+	if input.ContractName == "" {
```
erankirsh: I would also check for ContractName isn't nil

services/statestorage/service.go
```diff
@@ -1,44 +1,74 @@
...
+		return nil, fmt.Errorf("missing contract name")
+	}
+
+	contractState := s.persistence.ReadState(input.ContractName)
```
erankirsh: I think it would be better if ReadState would return err. Than you would avoid the if contractState == nil  and just check the err

test/builders/statediffs.go
```diff
@@ -1,28 +1,40 @@
...
 type sdiff struct {
 	builder *protocol.ContractStateDiffBuilder
+	records []*protocol.StateRecordBuilder
 }
 
 func ContractStateDiff() *sdiff {
 	return &sdiff{
 		builder: &protocol.ContractStateDiffBuilder{
 			ContractName: "BenchmarkToken",
-			StateDiffs: []*protocol.StateRecordBuilder{
```
erankirsh: See my comment above for the CommitStateDiff builder. 
I think this should not be removed. By default, ContractStateDiff should be built with values.

services/statestorage/service.go
ronnno: nil is not possible because the underlying type of ContractName is string

services/statestorage/test/read_keys.go
```diff
@@ -0,0 +1,101 @@
+package test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Reading a Key", func() {
+	When("not providing a contract name", func() {
```
erankirsh: I think those 3 tests:
not providing a contract name
providing a non existing contract
key doesn't exist
should be pushed into unit tests. 
Why? 
When you read those tests it seems for a moment that you are actually testing the driver and not the service itself. When diving down into it, I see that you actually test the service, but the abstraction of the driver is redundant in this case. 
I think that the general rule here is that if you have a test that is completely stateless, you should make it a unit test.

services/statestorage/test/get_block_height.go
```diff
@@ -0,0 +1,52 @@
...
+			Expect(height).To(Equal(0))
+			Expect(timestamp).To(Equal(0))
+		})
+	})
+
+	When("Block commits", func() {
+		It("Returns the current block height and timestamp", func() {
+			d := newStateStorageDriver()
+			heightBefore, _, err := d.getBlockHeightAndTimestamp()
+			contract1 := builders.ContractStateDiff().WithContractName("contract1").WithStringRecord("key1", "v1").WithStringRecord("key2", "v2").Build()
```
electricmonk: why do we care about keys and contract name here?

test/builders/statediffs.go
```diff
@@ -1,28 +1,40 @@
...
-				{Key: []byte("amount"), Value: []byte{10}},
-			},
 		},
 	}
 }
 
 func (s *sdiff) Build() *protocol.ContractStateDiff {
+	if len(s.records) == 0 {
+		s.records = append(s.records, &protocol.StateRecordBuilder{Key: []byte("amount"), Value: []byte{10}})
```
erankirsh: Why did you move the amount to the build? (Same as the comment above.)

services/statestorage/test/get_block_height.go
```diff
@@ -0,0 +1,52 @@
...
+			heightAfter, timestampAfter, err := d.getBlockHeightAndTimestamp()
+
+			Expect(err).ToNot(HaveOccurred())
+			Expect(heightAfter).To(Equal(heightBefore + 1))
+			Expect(timestampAfter).To(Equal(6579))
+
+		})
+	})
+
+	When("Block fails to commits", func() {
```
electricmonk: I don't understand the test here... aren't you confounding commit behavior with get height behavior? why do you need the first 3 commits?

test/harness/network.go
```diff
@@ -6,11 +6,11 @@ import (
 	"github.com/orbs-network/orbs-network-go/bootstrap"
 	"github.com/orbs-network/orbs-network-go/config"
 	"github.com/orbs-network/orbs-network-go/instrumentation"
+	stateStorageAdapter "github.com/orbs-network/orbs-network-go/services/statestorage/adapter"
```
erankirsh: I think I've talked to Ron about this. 
It's bad practice to include the production state storage adapter in the test harness. It would be better to duplicate the adapter file. @electricmonk Thoughts?

services/statestorage/test/read_keys.go
```diff
@@ -0,0 +1,101 @@
...
+				d.write(contract, "someOtherKey", value)
+
+				output, err := d.readSingleKey(contract, key)
+
+				Expect(err).ToNot(HaveOccurred())
+				Expect(output).To(Equal(value))
+			})
+		})
+
+		When("read 5 keys some are not existing", func() {
```
electricmonk: why 5? why not 2?

test/builders/statediffs.go
electricmonk: because we need to be able to generate a valid state diff. This is only called if dev did not call `WithStringRecord()`

test/harness/network.go
electricmonk: not yet. It's better not to do that at this moment, or even move this code to be only production code.

ronnno: I'm not following, @electricmonk  do you mean the way it is implemented now is preferable or that we should move this adapter to the test harness?

test/builders/statediffs.go
ronnno: an alternate way would be to create the builder with the default record, and on the first WithStringRecord() to remove that default value before appending the new pair..
@electricmonk  what do you think?

test/harness/network.go
electricmonk: I think that currently there's only one persistence adapter and it is a production one. Don't move to test harness. If/when we end up creating a persistent implementation, this one will have to move to test harness and a contract test will be required to assert that both implementations conform to the same contract.

test/builders/statediffs.go
electricmonk: how do you know it's a default value? you'll need to have a flag that indicates that the value is a default. I'd prefer to leave it as it is.

PR #42, opened 2018-07-26
Basic block storage persistence with leveldb

netoneko:
Work in progress, should be worked on after #25 

Introduces LevelDB as basic persistence layer (which in itself is up for discussion)


.gitmodules
```diff
@@ -64,3 +64,9 @@
 [submodule "vendor/golang.org/x/crypto"]
 	path = vendor/golang.org/x/crypto
 	url = https://go.googlesource.com/crypto
+[submodule "vendor/github.com/syndtr/goleveldb"]
+	path = vendor/github.com/syndtr/goleveldb
+	url = https://github.com/syndtr/goleveldb
+[submodule "vendor/github.com/golang/snappy"]
+	path = vendor/github.com/golang/snappy
```
talkol: what data are we planning to compress?

compression of our data (blocks and such) is a big feature that may not pay off, please discuss with OdedW before introducing compression. or at least I would recommend to avoid compression until we have some realistic data samples that we can check what's the compression ration gain we can get

.gitmodules
```diff
@@ -64,3 +64,9 @@
 [submodule "vendor/golang.org/x/crypto"]
 	path = vendor/golang.org/x/crypto
 	url = https://go.googlesource.com/crypto
+[submodule "vendor/github.com/syndtr/goleveldb"]
```
talkol: what research did we do before choosing LevelDB as our production solution? the fact that we used it in the alpha is not enough. 

1. did we compare it to other solutions? we have a very distinct use case.. 99% of our block data will almost never be accessed after it's stored

2. what are other blockchain projects using for their block storage?

3. I'll be happy to see benchmarks that show how well LevelDB handles TB of data. If I compare our block storage to analytics or logging solutions, they usually use file system to save the old logs.. why not go to this type of storage solutions? If we can't find any suitable benchmarks that we can run, maybe take the time and implement one ourselves

services/blockstorage/adapter/leveldb_persistence.go
```diff
@@ -12,28 +30,138 @@ type levelDbBlockPersistence struct {
...
+		}
+
+		bp.blockWritten <- false
+		return
+	}
+
 	bp.blockWritten <- true
 }
 
 func (bp *levelDbBlockPersistence) ReadAllBlocks() []*protocol.BlockPairContainer {
```
talkol: This interface function can't stay in a persistent production implementation of real storage.. it's only applicable for memory games and short lived tests

We only wrote it as a hack to get the acceptance test working, we need to stop with it

Assuming we have 10GB of blocks, there's no way we return all of them as an array

services/blockstorage/adapter/leveldb_persistence.go
```diff
@@ -12,28 +30,138 @@ type levelDbBlockPersistence struct {
...
-	bp.blockPairs = append(bp.blockPairs, blockPair)
+	var errors []error
+	var keys []string
+
+	if !basicValidation(blockPair) {
+		bp.reporting.Info("Block is invalid", instrumentation.Stringable("txBlockHeader", blockPair.TransactionsBlock.Header))
+		return
+	}
+
+	lastBlockHeight, lastBlockHeightRetrivalError := bp.loadLastBlockHeight()
```
talkol: I couldn't understand from a brief documentation of the https://github.com/syndtr/goleveldb library what is its threading model.. is it thread safe? do you need to maintain sync? does it support transactions? what happens when several threads at the same time try to run `WriteBlock`?

services/blockstorage/adapter/leveldb_persistence.go
```diff
@@ -12,28 +30,138 @@ type levelDbBlockPersistence struct {
...
+
+type config struct {
+	name string
+}
+
+func (c *config) NodeId() string {
+	return c.name
+}
+
+func NewLevelDbBlockPersistenceConfig(name string) Config {
```
talkol: Please modify this to the standard config mechanism, we don't want to define configs this way per service or adapter. We have a global system config that works through interfaces..

PR #56, opened 2018-07-31
Feature/benchmark consensus integration

talkol:


services/consensusalgo/benchmarkconsensus/leader.go
```diff
@@ -182,7 +182,7 @@ func (s *service) leaderHandleCommittedVote(sender *gossipmessages.SenderSignatu
 
 	// count if we have enough votes to move forward
 	existingVotes := len(s.lastCommittedBlockVoters) + 1
-	s.reporting.Info("Vote arrived", instrumentation.Int("existing-votes", existingVotes), instrumentation.Int("required-votes", s.requiredQuorumSize()))
+	s.reporting.Info("Valid vote arrived", instrumentation.Int("existing-votes", existingVotes), instrumentation.Int("required-votes", s.requiredQuorumSize()))
```
netoneko: Maybe it would be beneficial to log who voted and how.

services/consensusalgo/benchmarkconsensus/non_leader.go
```diff
@@ -82,16 +81,21 @@ func (s *service) nonLeaderCommitAndReplyUnderMutex(blockPair *protocol.BlockPai
...
+		Status: status,
+		Sender: (&gossipmessages.SenderSignatureBuilder{
+			SenderPublicKey: s.config.NodePublicKey(),
+			Signature:       sig,
+		}).Build(),
+	}
+
 	// send committed back to leader via gossip
+	recipient := blockPair.ResultsBlock.BlockProof.BenchmarkConsensus().Sender().SenderPublicKey()
+	s.reporting.Info("Replying committed with last committed height", instrumentation.BlockHeight(s.lastCommittedBlockHeight()), instrumentation.String("signed-data", signedData.String()))
```
netoneko: We have `instrumentation.Bytes` for binary data and `instrumentation.Stringable` for anything implementing a `Stringer` interface.

talkol: what's the difference between `instrumentation.Stringable` and `instrumentation.String`? isn't it exactly the same thing since they're both doing `String()`?

services/consensusalgo/benchmarkconsensus/leader.go
talkol: We have a log on the sender side when the vote is sent, in the recipient side we have a log on the transport arrived (which shows the message) and then this general log that the vote was approved. I used them all to debug this and it works pretty well

services/consensusalgo/benchmarkconsensus/non_leader.go
netoneko: `Stringable` converts something to `String` but `String` itself does not implement a `Stringer`

talkol: Cool, fixed

https://github.com/orbs-network/orbs-network-go/commit/64909efe9973b1de9a80090d2cf2175c04c0198c

PR #60, opened 2018-08-01
Refactor block storage tests, introduce atomic writes

### LINKED ISSUE #50: Convert configs to standard config mechanism
Please convert the configs to the standard way config are defined in the system

Notice this:
https://github.com/orbs-network/orbs-network-go/blob/3cb919e064be64ebecf7c5c1e372c65f52abce60/services/blockstorage/service_config.go#L13

It does not appear here:
https://github.com/orbs-network/orbs-network-go/blob/3cb919e064be64ebecf7c5c1e372c65f52abce60/config/hardcoded_config.go#L11

Also notice this config:
https://github.com/orbs-network/orbs-network-go/blob/3cb919e064be64ebecf7c5c1e372c65f52abce60/main.go#L33

It's also not standard

Please have a session with @electricmonk about how he designed the configuration in the node and what's the correct way to plug into it

### LINKED ISSUE #53: Resolve sync issues
https://github.com/orbs-network/orbs-network-go/blob/3cb919e064be64ebecf7c5c1e372c65f52abce60/services/blockstorage/service.go#L59

This needs a mutex. You don't know any assumptions on which goroutines your service methods are called, they can be called concurrently from multiple ones and you have shared variables.

Access to shared variables needs to be protected.

netoneko:
Fixes #50 and fixes #53


bootstrap/node.go
```diff
@@ -45,6 +46,7 @@ func NewNode(
 		constantConsensusLeader,
 		activeConsensusAlgo,
 		benchmarkConsensusRoundRetryIntervalMillisec,
+		blockSyncCommitTimeoutMS,
```
talkol: Can we have the name consistent with the line above? Choose one format and change both to match :)

config/config.go
```diff
@@ -17,6 +18,9 @@ type NodeConfig interface {
 
 	// benchmark consensus
 	BenchmarkConsensusRoundRetryIntervalMillisec() uint32
+
+	// block storage
+	BlockSyncCommitTimeout() time.Duration
```
talkol: Let's change the function name to be consistent with the variable, like `BlockSyncCommitTimeoutMS()`

services/blockstorage/service.go
```diff
@@ -82,34 +102,27 @@ func (s *service) loadTransactionsBlockHeader(height primitives.BlockHeight) (*s
...
-		go func() {
-			const interval = 10 * NanosecondsInMillisecond
-			timeout := s.config.BlockSyncCommitTimeout().Nanoseconds()
-
-			for i := int64(0); i < timeout; i += interval {
+	currentBlockHeight := s.lastCommittedBlockHeight
+	if input.BlockHeight > currentBlockHeight && input.BlockHeight-currentBlockHeight <= 5 {
+		timeout := time.NewTimer(s.config.BlockSyncCommitTimeout())
+		defer timeout.Stop()
+		tick := time.NewTicker(10 * time.Millisecond)
```
talkol: Let's add a future TODO here to remove the polling altogether, we had several ideas of how it can be done.

My personal favorite is to create a channel for the result block height you wait for and then wait on it.

Then when we reach this block height, close the channel.

This way we'll always have a small map of 5 active channels for block heights that are still within our window.

jlevison: we have an open issue on this, i'll reference to it in the todo

PR #65, opened 2018-08-02
Feature/improve errors and logs according to Jonathan's spotlight talk

talkol:


services/consensusalgo/benchmarkconsensus/leader.go
```diff
@@ -173,7 +163,7 @@ func (s *service) leaderHandleCommittedVote(sender *gossipmessages.SenderSignatu
 	// validate the vote
 	err := s.leaderValidateVoteUnderMutex(sender, status)
 	if err != nil {
-		s.reporting.Error(err.Error()) // TODO: wrap with added context
+		s.reporting.Error("leader failed to validate vote", instrumentation.Error(err))
 		return
```
jlevison: this logs the error but then does not return it to the caller, is that the intended behavior?

services/consensusalgo/benchmarkconsensus/non_leader.go
```diff
@@ -16,26 +16,17 @@ func (s *service) nonLeaderHandleCommit(blockPair *protocol.BlockPairContainer)
 
 	err := s.nonLeaderValidateBlockUnderMutex(blockPair)
 	if err != nil {
-		s.reporting.Error(err.Error()) // TODO: wrap with added context
+		s.reporting.Error("non leader failed to validate block", instrumentation.Error(err))
 		return
```
jlevison: same as above, this means the error is handled already and no other actions will be taken

services/consensusalgo/leanhelix/service.go
```diff
@@ -88,15 +87,16 @@ func (s *service) HandleLeanHelixNewView(input *gossiptopics.LeanHelixNewViewInp
...
 	for {
-		s.reporting.Info("Entered consensus round, last committed block height is", instrumentation.BlockHeight(s.lastCommittedBlockHeight))
+		s.reporting.Info("entered consensus round with last committed block height", instrumentation.BlockHeight(s.lastCommittedBlockHeight))
 
 		// see if we need to propose a new block
 		err := s.leaderProposeNextBlockIfNeeded()
 		if err != nil {
-			s.reporting.Error(err.Error())
+			s.reporting.Error("leader failed to propose next block", instrumentation.Error(err))
 			continue
```
jlevison: these errors are not fatal?

services/publicapi/service.go
```diff
@@ -28,8 +28,8 @@ func NewPublicApi(
 }
 
 func (s *service) SendTransaction(input *services.SendTransactionInput) (*services.SendTransactionOutput, error) {
-	s.reporting.Info("enter_send_transaction")
-	defer s.reporting.Info("exit_send_transaction")
+	s.reporting.Info("enter SendTransaction")
+	defer s.reporting.Info("exit SendTransaction")
```
jlevison: feels too verbose, but not sure its important to change this now (the change is simple :) )
not sure what kind of context this gives to whoever will read these messages

services/consensusalgo/benchmarkconsensus/leader.go
talkol: yes, the caller is a gossip incoming message handler.. it doesn't have anything to do with the error. this is the top level so it should log and ignore on problem

services/publicapi/service.go
talkol: Yes, it's not my errors, I just refactored them.. I assume they will disappear eventually

PR #67, opened 2018-08-02
Feature/processor continued

talkol:


services/processor/native/call.go
```diff
@@ -52,23 +52,30 @@ func (s *service) verifyInternalMethodCall(contractInfo *types.ContractInfo, met
 	return errors.Errorf("internal method '%s' called from different service '%s' without system permissions", methodInfo.Name, callingService)
 }
 
-func (s *service) processMethodCall(contractInfo *types.ContractInfo, methodInfo *types.MethodInfo, args []*protocol.MethodArgument) ([]*protocol.MethodArgument, error, error) {
+func (s *service) processMethodCall(ctx types.Context, contractInfo *types.ContractInfo, methodInfo *types.MethodInfo, args []*protocol.MethodArgument) (outArgs []*protocol.MethodArgument, outErr error, err error) {
```
electricmonk: why do we need two error return values?

services/processor/native/call.go
```diff
@@ -77,45 +84,51 @@ func (s *service) processMethodCall(contractInfo *types.ContractInfo, methodInfo
...
 	return outArgs, outErr, err
 }
 
-func (s *service) verifyMethodArgs(methodInfo *types.MethodInfo, implementation interface{}, args []*protocol.MethodArgument) ([]reflect.Value, error) {
+func (s *service) verifyMethodArgs(ctx types.Context, methodInfo *types.MethodInfo, implementation interface{}, args []*protocol.MethodArgument) ([]reflect.Value, error) {
 	res := []reflect.Value{}
 	methodType := reflect.ValueOf(implementation).Type()
-	if methodType.NumIn()-1 != len(args) {
-		return nil, errors.Errorf("method '%s' takes %d args but received %d", methodInfo.Name, methodType.NumIn()-1, len(args))
+	if methodType.NumIn() < 2 || methodType.In(1) != reflect.TypeOf(ctx) {
```
electricmonk: `2` being a magic number might be better as extracted to a constant

services/processor/native/call.go
```diff
@@ -77,45 +84,51 @@ func (s *service) processMethodCall(contractInfo *types.ContractInfo, methodInfo
 	}
 
 	// get contract error
-	outErr, err := s.createContractError(methodInfo, outValues[len(outValues)-1])
+	outErr, err = s.createContractError(methodInfo, outValues[len(outValues)-1])
 	return outArgs, outErr, err
 }
 
-func (s *service) verifyMethodArgs(methodInfo *types.MethodInfo, implementation interface{}, args []*protocol.MethodArgument) ([]reflect.Value, error) {
+func (s *service) verifyMethodArgs(ctx types.Context, methodInfo *types.MethodInfo, implementation interface{}, args []*protocol.MethodArgument) ([]reflect.Value, error) {
```
electricmonk: The name is a bit misleading. Seems like it does more than just verify the actual arguments against the expected method params, otherwise it would've returned only a bool

services/processor/native/service.go
```diff
@@ -7,26 +7,22 @@ import (
 	"github.com/orbs-network/orbs-spec/types/go/protocol"
 	"github.com/orbs-network/orbs-spec/types/go/services"
 	"github.com/orbs-network/orbs-spec/types/go/services/handlers"
+	"github.com/pkg/errors"
 )
 
 type service struct {
```
electricmonk: why the name `service`? it's an overloaded name as it is, now you're adding more files / entities with that name. What's wrong with `processor`?

services/processor/native/service.go
```diff
@@ -70,5 +67,11 @@ func (s *service) DeployNativeService(input *services.DeployNativeServiceInput)
 }
 
 func (s *service) RegisterContractSdkCallHandler(handler handlers.ContractSdkCallHandler) {
```
electricmonk: I don't get this method. Why isn't this being done on construction?

services/processor/native/test/call_argument_test.go
```diff
@@ -78,11 +78,29 @@ func TestCallThatThrowsError(t *testing.T) {
 	assert.Equal(t, output.CallResult, protocol.EXECUTION_RESULT_ERROR_SMART_CONTRACT, "call result should be smart contract error")
 }
 
-func TestCallWithInvalidMissingErrorFails(t *testing.T) {
+func TestCallThatPanics(t *testing.T) {
+	h := newHarness()
+	call := processCallInput().WithMethod("BenchmarkContract", "panic").WithArgs().Build()
```
electricmonk: Why are you calling `WithArgs()` passing nothing?

services/processor/native/call.go
talkol: We differentiate between the contract inside returned an error to that there was an error running the contract (whether the error was in Orbs core code or the contract developer's code)

Two examples:

1. You try to read a value from StateStorage but the state DB is down so the call fails - this is an error - this is an Orbs core error

2. The contract transfer attempts to draw money from an account and it will bring the account to a negative balance, which throws an error in the contract - this is a contract developer error

services/processor/native/service.go
talkol: This method will be called on the construction of `VirtualMachine`, this part wasn't implemented yet

services/processor/native/test/call_argument_test.go
talkol: It's a contract that receives zero arguments

Since it's a variadic function, I just call it empty. Do you want me to change it to `WithNoArgs()` to make it more clear?

services/processor/native/service.go
talkol: It's a style question. Eventually, `service.go` files are just routing/boilerplate that hook up your actual service implementation to the interfaces that the spec dictates

Because it's only a routing file with almost zero logic, I don't want to give it a domain meaningful name

services/processor/native/call.go
talkol: done

services/processor/native/call.go
talkol: fixed

services/processor/native/test/call_argument_test.go
talkol: fixed

PR #72, opened 2018-08-06
Feature/statestore readpastblocks

noambergIL:
Added code for height for state storage (presistent level) + basic hash of the state.
(needed by tal for now)


bootstrap/node.go
```diff
@@ -36,6 +36,7 @@ func NewNode(
 	activeConsensusAlgo consensus.ConsensusAlgoType,
 	benchmarkConsensusRoundRetryIntervalMillisec uint32, // TODO: move all of the config from the ctor, it's a smell
 	transport gossipAdapter.Transport,
+	maxStateHistory uint64,
```
talkol: are the units here in blocks? if so, maybe improve the name or change from `uint64` to `primitives.blockHeight` to clarify

talkol: hmm.. since it's in config maybe leave as `uint64` but improve the name.. in the [config](https://github.com/orbs-network/orbs-spec/blob/master/behaviors/config/services.md) in the spec we called it `recent-block-snapshot-num`

services/statestorage/adapter/memory_persistence.go
```diff
@@ -4,41 +4,79 @@ import (
...
-	stateWritten chan bool
-	stateDiffs   map[primitives.ContractName]map[string]*protocol.StateRecord
-	config       Config
+	stateWritten 	chan bool
+	stateDiffs   	map[primitives.BlockHeight]map[primitives.ContractName]map[string]*protocol.StateRecord
 }
 
-func NewInMemoryStatePersistence(config Config) StatePersistence {
+func NewInMemoryStatePersistence() StatePersistence {
+	stateDiffsContract :=  map[primitives.ContractName]map[string]*protocol.StateRecord{primitives.ContractName("BenchmarkToken"): {}}
```
talkol: Do we have to initialize "BenchmarkToken" by name here? If we avoid initializing it, won't it be initialized on first write? or do you want to avoid that it fails on first read?

services/statestorage/adapter/memory_persistence.go
```diff
@@ -4,41 +4,79 @@ import (
...
 	return nil
 }
 
-func (sp *InMemoryStatePersistence) ReadState(contract primitives.ContractName) (map[string]*protocol.StateRecord, error){
-	if contractStateDiff, ok := sp.stateDiffs[contract]; ok {
-		return contractStateDiff, nil
+func (sp *InMemoryStatePersistence) cloneCurrentStateDiff(height primitives.BlockHeight) map[primitives.ContractName]map[string]*protocol.StateRecord {
+	prevHeight := height - primitives.BlockHeight(1)
+	if _, ok := sp.stateDiffs[prevHeight]; !ok {
+		panic ("trying to commit blocks not in order")
```
talkol: I think we decided to avoid all `panics` in production code because we can't really handle the node crashing. A better alternative would be to log an error to reporting, and recover gracefully

services/statestorage/adapter/memory_persistence.go
```diff
@@ -4,41 +4,79 @@ import (
...
+
 	sp.stateWritten <- true
 
 	return nil
 }
 
-func (sp *InMemoryStatePersistence) ReadState(contract primitives.ContractName) (map[string]*protocol.StateRecord, error){
-	if contractStateDiff, ok := sp.stateDiffs[contract]; ok {
-		return contractStateDiff, nil
+func (sp *InMemoryStatePersistence) cloneCurrentStateDiff(height primitives.BlockHeight) map[primitives.ContractName]map[string]*protocol.StateRecord {
```
talkol: maybe we should simplify `map[primitives.ContractName]map[string]*protocol.StateRecord` with a typedef?

services/statestorage/adapter/memory_persistence_test.go
```diff
@@ -8,16 +8,16 @@ import (
 var _ = Describe("Reading a Key", func() {
 	When("not providing a contract name", func() {
 		It("Returns an error", func() {
-			d := NewInMemoryStatePersistence(&struct{}{})
-			_, err := d.ReadState("")
+			d := NewInMemoryStatePersistence()
```
talkol: Recommend that you kill ginkgo as soon as you can

services/statestorage/adapter/persistence.go
```diff
@@ -6,6 +6,6 @@ import (
 )
 
 type StatePersistence interface {
-	WriteState(contract primitives.ContractName, stateDiffs *protocol.StateRecord) error // TODO: change this to an array as well since we do multiple writes in one transactions
-	ReadState(contract primitives.ContractName) (map[string]*protocol.StateRecord, error)
+	WriteState(height primitives.BlockHeight, contract primitives.ContractName, stateDiffs *protocol.StateRecord) error // TODO: change this to an array as well since we do multiple writes in one transactions
```
talkol: Since `WriteState` should always work in the granularity of a block, what happens if we start failing mid-way through a block? The current interface isn't easily transactional

services/statestorage/service.go
```diff
@@ -6,17 +6,24 @@ import (
...
 )
 
+type Config interface {
+	GetMaxStateHistory() uint64
+}
+
 type service struct {
 	persistence            adapter.StatePersistence
 	lastResultsBlockHeader *protocol.ResultsBlockHeader
+	config				   Config
```
talkol: are you running `go fmt` before committing?

test/harness/network.go
```diff
@@ -55,7 +55,7 @@ type networkNode struct {
 func NewTestNetwork(ctx context.Context, numNodes uint32, consensusAlgo consensus.ConsensusAlgoType) AcceptanceTestNetwork {
 
 	testLogger := instrumentation.GetLogger().WithFormatter(instrumentation.NewHumanReadableFormatter())
-	testLogger.Info("Creating acceptance test network", instrumentation.String("consensus", consensusAlgo.String()), instrumentation.Uint32("num-nodes", numNodes))
+	testLogger.Info("creating acceptance test network", instrumentation.String("consensus", consensusAlgo.String()), instrumentation.Uint32("num-nodes", numNodes))
```
talkol: good job catching this!! :)
I already fixed it though, it's in master.. so make sure you merge your code with master again

bootstrap/node.go
electricmonk: `stateHistoryRetentionInBlockHeights`

services/statestorage/adapter/memory_persistence.go
electricmonk: most def

electricmonk: `type ContractState` map[string]*protocol.StateRecord
`type StateVersion map[primitives.ContractName]ContractState`

services/statestorage/adapter/memory_persistence_test.go
electricmonk: before it hatches

services/statestorage/adapter/persistence.go
electricmonk: It will never be transactional. A better interface will only give us the illusion of transactionality. I do agree, though.

services/statestorage/adapter/memory_persistence.go
noambergIL: we were told that the provider will create the contract so that we have a default by name for now.

bootstrap/node.go
noambergIL: done

services/statestorage/adapter/persistence.go
noambergIL: interface changed. (internal is still not transactional for now)

PR #73, opened 2018-08-06
Feature/consensus context first component test

IdoZilberberg:
@jlevison take a look - I changed all "millisec" to "millis".
@talkol I did not change the units from Duration to uint32 yet, this will wait for another PR.


services/consensuscontext/fetch_transactions.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/orbs-network/orbs-spec/types/go/services"
+	"time"
+	"fmt"
+)
+
+func (s *service) fetchTransactions(input *services.GetTransactionsForOrderingInput, minimumTransactionsInBlock int, belowMinimalBlockDelayMillis uint32) (*services.GetTransactionsForOrderingOutput, error) {
+
+	fmt.Println("Fetch transactions")
+	var proposedTransactions *services.GetTransactionsForOrderingOutput = nil
```
jlevison: why do we need to specifically define this? the line below will handle that, no? either case, the code should run the same without that line

services/consensuscontext/fetch_transactions.go
```diff
@@ -0,0 +1,29 @@
...
+	proposedTransactions, err := s.transactionPool.GetTransactionsForOrdering(input)
+	if err != nil {
+		return nil, err
+	}
+	txCount := len(proposedTransactions.SignedTransactions)
+	if txCount >= minimumTransactionsInBlock {
+		return proposedTransactions, nil
+	}
+	// TODO: How to test that Sleep() was called
+	time.Sleep(time.Duration(belowMinimalBlockDelayMillis) * time.Millisecond)
```
jlevison: if we need to run something twice with a delay maybe we should use a probe (see slack / block storage Get..Header functions), then look at tests on how we make sure the probe was used

services/consensuscontext/fetch_transactions.go
IdoZilberberg: Right, fixed.

services/consensuscontext/fetch_transactions.go
IdoZilberberg: Note for me: 
blockstorage/service.go:104 - GetTransactionBlockHeader()
get_block_headers.go:55 - timeoutError

bootstrap/node.go
```diff
@@ -31,22 +31,27 @@ func NewNode(
...
 	nodePrivateKey primitives.Ed25519PrivateKey,
 	federationNodes map[string]config.FederationNode,
-	blockSyncCommitTimeoutMillisec uint32,
+	blockSyncCommitTimeoutMillis uint32,
 	constantConsensusLeader primitives.Ed25519PublicKey,
 	activeConsensusAlgo consensus.ConsensusAlgoType,
-	benchmarkConsensusRoundRetryIntervalMillisec uint32, // TODO: move all of the config from the ctor, it's a smell
+	benchmarkConsensusRoundRetryIntervalMillis uint32, // TODO: move all of the config from the ctor, it's a smell
 	transport gossipAdapter.Transport,
+	belowMinimalBlockDelayMillis uint32,
```
electricmonk: `NewNode` is now officially taking too many arguments. We need to decide, ASAP, how to handle passing config to `Node` from `main` and tests. @talkol any ideas?

services/consensuscontext/create_block.go
```diff
@@ -10,26 +10,25 @@ import (
 
 func (s *service) createTransactionsBlock(blockHeight primitives.BlockHeight, prevBlockHash primitives.Sha256) (*protocol.TransactionsBlockContainer, error) {
 
-	proposedTransactions, err := s.transactionPool.GetTransactionsForOrdering(&services.GetTransactionsForOrderingInput{
+	proposedTransactions, err := s.fetchTransactions(&services.GetTransactionsForOrderingInput{
```
electricmonk: since you introduced a private method, might as well move the ugly inter-service struct `services.GetTransactionsForOrderingInput` into it and have it take an `int` argument

services/consensuscontext/fetch_transactions.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/orbs-network/orbs-spec/types/go/services"
+	"time"
+	"fmt"
+)
+
+func (s *service) fetchTransactions(input *services.GetTransactionsForOrderingInput, minimumTransactionsInBlock int, belowMinimalBlockDelayMillis uint32) (*services.GetTransactionsForOrderingOutput, error) {
+
+	fmt.Println("Fetch transactions")
```
electricmonk: please remove println.
@erankirsh who's taking the linter task?

bootstrap/node.go
IdoZilberberg: Correct, but it will not be resolved as part of this PR. Opened issue #74.

services/consensuscontext/create_block.go
IdoZilberberg: done

services/consensuscontext/fetch_transactions.go
IdoZilberberg: Keeping Sleep() for now until we decide on the correct mechanism for syncing between services (e.g. context)

IdoZilberberg:
@electricmonk done, take a look.
@jlevison

bootstrap/node.go
electricmonk: Please take the time to handle #74 as well. Boy scout rule applies - leave the code better than how you found it. :)

PR #77, opened 2018-08-07
Added t.Parallel to benchmark consensus tests since they’re slow and …

talkol:
…should be parallelized


talkol:
As far as I can see it can't. `m *testing.M` exports a single function that takes no arguments - `m.Run()`

I found a proposal that asks to add this feature:
https://github.com/golang/go/issues/21214
"I could imagine permitting people to say `t.AllParallel()` in their `TestMain` function"

It was rejected

Or even just add a command line flag for this feature:
https://github.com/golang/go/issues/24335

Also rejected

Issue #74, opened 2018-08-07
introduce file based config system


IdoZilberberg:
Func `NewNode()` in node.go:29 has too many parameters and this will get worse unless an alternative mechanism for passing config is found. 
This is relevant to prod and tests.
See @electricmonk's comment on #73.
@talkol

electricmonk:
Requirements:
- every service defines its config (like today, using an interface)
- allow setting bare minimum of config for component tests
- config should be versioned by block height (V1.2)
- maybe implement by code generation (V1.1)
- config system should support rich golang types (e.g. `time.Duration`) - reading from file will automatically translate. This means setting a standard for all duration prop keys in config (for instance suffixed with `_in_millis`).
- retain config factory methods `ForAcceptanceTests` / `ForProduction`
- main() should take a mandatory arg defining path to config

yonidavidson:
You can change the NewNode(params) to a struct NodeConfig.
And create multiple "constructors"
NewNodeConfig(...regular params such as new Node )
NewNodeConfigFromFile("file_name").
NewNodeConfigFromEnviroment()

PR #88, opened 2018-08-11
Block storage transaction receipt query #1

jlevison:
this is a large feature, and will be in parts to ease up on the review process

included:
* basic bloom filter - to be changed in the next pr around this feature - the spec requires two bloom filters and after discussing with Oded this may not be the right course of action, so the implementation is on an extremely naive bloom filter right now
* added more configuration values for block storage
* added the initial logic for querying by time on the persistence storage


bootstrap/node.go
```diff
@@ -32,6 +32,9 @@ func NewNode(
 	nodePrivateKey primitives.Ed25519PrivateKey,
 	federationNodes map[string]config.FederationNode,
 	blockSyncCommitTimeoutMillis uint32,
+	blockTransactionReceiptQueryStartGraceSec uint32,
```
electricmonk: See #74. This function is taking too many parameters and it's imperative that we spend some time fixing it. Please take some time along with @IdoZilberberg and pay back this technical debt this week.

crypto/bloom/timestamp_bloom_test.go
```diff
@@ -0,0 +1,122 @@
...
+			t.Errorf("size should be %d, but is %d", test.size, x.Size())
+		}
+
+		if x.BitCount() != test.bitCount {
+			t.Errorf("bitcount should be %d, but is %d", test.bitCount, x.BitCount())
+		}
+	}
+}
+
+func TestTimestampBloomFilter_Add(t *testing.T) {
```
electricmonk: Bloom filters might be a good opportunity to play with Property Based Testing, which essentially means running your invariants against a large set of randomly-generated test data.

A quick Google finds this:
https://godoc.org/github.com/leanovate/gopter

Maybe have a look?

services/blockstorage/adapter/persistence.go
```diff
@@ -8,6 +8,7 @@ import (
 type BlockPersistence interface {
 	WriteBlock(blockPairs *protocol.BlockPairContainer) error
 	ReadAllBlocks() []*protocol.BlockPairContainer
+	ReadAllBlocksByTimeRange(start, end primitives.TimestampNano) []*protocol.BlockPairContainer
```
electricmonk: if it's by time range, it's not all blocks. Maybe `FindBlocksWithinTimeRange`? also, this should _probably_ return a cursor so as to not allow your users to kill the system

services/blockstorage/test/transaction_receipt.go
```diff
@@ -0,0 +1,82 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/orbs-network/orbs-network-go/crypto/digest"
+	"github.com/orbs-network/orbs-network-go/test/builders"
+	"github.com/orbs-network/orbs-spec/types/go/primitives"
+	"github.com/orbs-network/orbs-spec/types/go/services"
+	"time"
+)
+
+var _ = Describe("Block Storage", func() {
+	When("fetching transaction receipts", func() {
```
electricmonk: please migrate to testify

jlevison: after this is approved **not** as part of this PR
see https://tree.taiga.io/project/orbs-network/us/108 ETA is this week

services/blockstorage/adapter/persistence.go
jlevison: good point, fixing

crypto/bloom/timestamp_bloom_test.go
jlevison: good idea, the current implementation is not a real one, its more of a stub then something we should vigorously test - we should add this as a requirement for the real implementation (pending some research talks, will hopefully be implementing by the end of August)

bootstrap/node.go
jlevison: correct
but maybe we can try and do a fuel tank test on it, i.e. how low can it go before the car stops, so how many can we accumulate here until no-one can add anything because its too big? :) (talking to ido)

services/blockstorage/adapter/persistence.go
jlevison: re: cursor
system should support pagination (i will add it to the spec and open a pr) but this specific request returns a specific result (like a boolean) and not a set of results
the search 'space', start->end is implementation based and not exposed by the external api, external api which currently calls it exposes a single timestamp

electricmonk: nevertheless, someone contributing to the system or a new developer in the core team might think it's feasible (since the API supports it!) to retrieve a huge chunk of blocks this way.

bootstrap/node.go
electricmonk: I don't object to merging this PR without handling the config, as long as we *commit* to fixing the config problem this week.

services/blockstorage/adapter/persistence.go
electricmonk: at the very least, please add a `maxNumOfResults` argument to the list, and document (godoc) that it should never exceed some `k` size.

config/config.go
```diff
@@ -21,6 +21,9 @@ type NodeConfig interface {
 
 	// block storage
 	BlockSyncCommitTimeoutMillis() time.Duration
+	BlockTransactionReceiptQueryStartGraceSec() time.Duration
```
talkol: It's not consistent that some of the time related configs we have are in `uint32` (like `BenchmarkConsensusRoundRetryIntervalMillis`) and some are in `time.Duration` (like `BlockTransactionReceiptQueryStartGraceSec`).

We should choose one convention

Personally I prefer `uint32` because when we have config providers such as json files or smart contracts, primitive types are easier to work with and go-specific constructs like `time.Duration` are language-specific. In the jsons we're going to have time in `uint32` and I usually like these configs to be closer to the code

crypto/bloom/timestamp_bloom.go
```diff
@@ -0,0 +1,145 @@
+package bloom
```
talkol: maybe name the file `timestamp.go` because it's already in package `bloom`?

services/blockstorage/service.go
```diff
@@ -74,19 +77,32 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
...
-	s.lastCommittedBlockHeight = txBlockHeader.BlockHeight()
-	s.lastCommittedBlockTimestamp = txBlockHeader.Timestamp()
+	s.lastCommittedBlock = block
+}
+
+func (s *service) lastCommittedBlockHeight() primitives.BlockHeight {
+	if s.lastCommittedBlock == nil {
+		return 0
+	}
+	return s.lastCommittedBlock.TransactionsBlock.Header.BlockHeight()
```
talkol: You only protect under mutex the write without protecting the read. Is this our convention? Are we 100% sure that this is safe? isn't it an implementation detail of how the code is compiled?

If this is indeed our convention, we should communicate to everybody because lots of the code in the system is locking for both read and write.

If it's not our convention, we should fix,

services/blockstorage/service.go
```diff
@@ -74,19 +77,32 @@ func (s *service) CommitBlock(input *services.CommitBlockInput) (*services.Commi
...
+}
+
+func (s *service) lastCommittedBlockTimestamp() primitives.TimestampNano {
+	if s.lastCommittedBlock == nil {
+		return 0
+	}
+	return s.lastCommittedBlock.TransactionsBlock.Header.Timestamp()
```
talkol: It's scary to me that this code is not under mutex.. `s.lastCommittedBlock` might be updated after the `if` is ran but before the return runs. So the if result might not be relevant. Let's assume that `s.lastCommittedBlock` could be updated externally to nil, this would mean that the return will crash..

services/blockstorage/service.go
```diff
@@ -171,14 +187,49 @@ func (s *service) GetResultsBlockHeader(input *services.GetResultsBlockHeaderInp
...
+		BlockTimestamp:     s.lastCommittedBlockTimestamp(),
+	}
+}
+
 func (s *service) GetTransactionReceipt(input *services.GetTransactionReceiptInput) (*services.GetTransactionReceiptOutput, error) {
-	panic("Not implemented")
+	start := input.TransactionTimestamp - primitives.TimestampNano(s.config.BlockTransactionReceiptQueryStartGraceSec().Nanoseconds())
+	end := input.TransactionTimestamp + primitives.TimestampNano((s.config.BlockTransactionReceiptQueryEndGraceSec() + s.config.BlockTransactionReceiptQueryTransactionExpireSec()).Nanoseconds())
+
+	blocksToSearch := s.persistence.GetBlocksByTimeRange(start, end)
```
talkol: The function `GetBlocksByTimeRange` is very expensive in memory. Blocks will be very big, around 1MB each.. If the time range we search in is 30 minutes (transaction expiry), and we close a block every 5 seconds, then this function returns over 300MB of memory.. is my calculation correct?

I'm not sure we can read from our DB to memory 300MB for every query.. it will make the node vulnerable to DoS

We can optimize this significantly.. for example, we only need the `ResultsBlock.Headers` in step 1, which are quite small. And only after we find the positively matched blocks by bloom, we can query the transactions from the persistent storage (the transactions are the big memory hit - they are big).

If we decide it's too early to optimize at this stage, at least add a very clear TODO that this code cannot go to production before being changed.

services/blockstorage/service.go
jlevison: its not about convention but rather what is required. when we were looking at the code figuring out the spots requiring protection we did not find any place requiring read protection (we actually did, but refactored to avoid the lock)
by the way, we since refactored again and i am not sure this code even needs write protection, but its irrelevant to discuss this in detail as we will refactor more until we get to persistence

services/blockstorage/service.go
jlevison: not sure i see how read protect will assist here
its nil on initialization, this is just good practice, but its never really nil
and if its updated *after* the `if`, read protection will not assist with getting a more accurate answer (no interrupt control can actually, the `if` already happened)

i am going to do some research on go to be sure, but the read operation in line 105, as far as i know, will be executed atomically, the `timestamp()` call will not, but its data fetch will, and if we are going to assume that between 102->105 the lastCB will be zeroed out, then a lot more than our code is unsafe in that matter (every struct call need to be protected? and the protection is not only around the read in that case, its around the entire execution flow) and again, afaik there is no line of code that will do `s.lCB = nil`

crypto/bloom/timestamp_bloom.go
jlevison: yeah, make sense

config/config.go
jlevison: i agree on all comments, but since we *must* refactor the configuration, and next week (end of week) we will have a new version, how important (1 to 10) that i refactor this here?

bootstrap/node.go
```diff
@@ -32,6 +32,9 @@ func NewNode(
...
+	blockTransactionReceiptQueryEndGraceSec uint32,
+	blockTransactionReceiptQueryTransactionExpireSec uint32,
```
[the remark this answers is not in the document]
jlevison: i understood we do not need the granularity, but no problems to change it, see comment about the config refactor, i will couple it together

services/blockstorage/service.go
jlevison: todo was added a few minutes ago, its all true and we know this is problematic, lets wait for real persistence and then protect ourselves accordingly - this was refactored already to make it more safe regardless of the persistence

PR #89, opened 2018-08-12
TX pool - some progress

electricmonk:
* most of the logic in commit transaction receipts
* some logic of handle forwarded transactions
* almost done with add new transaction

### DIFF
--- a/services/transactionpool/add_new_transaction.go
+++ b/services/transactionpool/add_new_transaction.go
@@ -41,7 +41,7 @@ func (s *service) AddNewTransaction(input *services.AddNewTransactionInput) (*se
 	}
 
 	s.reporting.Info("adding new transaction to the pool", instrumentation.Stringable("transaction", input.SignedTransaction))
-	if _, err := s.pendingPool.add(input.SignedTransaction); err != nil {
+	if _, err := s.pendingPool.add(input.SignedTransaction, s.config.NodePublicKey()); err != nil {
 		return nil, err
 
 	}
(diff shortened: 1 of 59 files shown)


services/transactionpool/pools.go
```diff
@@ -0,0 +1,97 @@
...
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	size := uint32(len(transaction.Raw()))
+
+	if p.currentSizeInBytes+size > p.config.PendingPoolSizeInBytes() {
+		return nil, &ErrTransactionRejected{protocol.TRANSACTION_STATUS_RESERVED} //TODO change to TRANSACTION_STATUS_REJECTED_CONGESTION
+	}
+
+	p.currentSizeInBytes += size
+	p.transactions[key.KeyForMap()] = &pendingTransaction{size: size, gatewayPublicKey: gatewayPublicKey}
```
jlevison: &pendingTransaction{size: size
name it sizeInBytes: size

services/transactionpool/add_new_transaction.go
```diff
@@ -40,20 +40,38 @@ func (s *service) AddNewTransaction(input *services.AddNewTransactionInput) (*se
...
+func (s *service) forwardTransaction(tx *protocol.SignedTransaction) error {
+	// TODO sign
+	//sig, err := signature.SignEd25519(s.config.NodePrivateKey(), signedData)
+	//if err != nil {
+	//	return nil, err
+	//}
+
+	_, err := s.gossip.BroadcastForwardedTransactions(&gossiptopics.ForwardedTransactionsInput{
+		Message: &gossipmessages.ForwardedTransactionsMessage{
+			SignedTransactions: []*protocol.SignedTransaction{tx},
```
jlevison: should it be an array here? (this is the copy i talked about, its not a copy, missed that it is an array)

PR #91, opened 2018-08-13
Feature/logger improvements

jlevison:
four changes as described by the commit messages
@talkol @IdoZilberberg , you were right about the params order, was bad code that is now fixed

re: StringableSlice, it uses reflection, while its fast with go1.9+, its ugly (albeit safe, will not panic in the way its used) - the problem is that accepting 'any slice' requires reflection to work with it (or at least i couldn't find another way) - let me know what you think


instrumentation/basic_logger.go
```diff
@@ -86,6 +86,29 @@ func Stringable(key string, value fmt.Stringer) *Field {
 	return &Field{Key: key, String: value.String(), Type: StringType}
 }
 
+func StringableSlice(key string, values interface{}) *Field {
+	builder := strings.Builder{}
+	builder.WriteString("[")
+
+	switch reflect.TypeOf(values).Kind() {
```
electricmonk: this is a performance hit. We need to think if we can defer execution of this function, so that in production we can choose "debug levels" or whatnot, thereby saving on the performance hit, or at least executing the expensive logging-related stuff in a background goroutine. I suggest to open an issue on that and make sure to decide if we want to handle it before V1 goes live.

instrumentation/basic_logger_test.go
```diff
@@ -80,13 +83,41 @@ func TestNestedLogger(t *testing.T) {
...
+	jsonMap := parseStdout(stdout)
+
+	Expect(jsonMap["level"]).To(Equal("info"))
+	Expect(jsonMap["node"]).To(Equal("node1"))
+	Expect(jsonMap["service"]).To(Equal("public-api"))
+	Expect(jsonMap["function"]).To(Equal("instrumentation_test.TestStringableSlice.func1"))
+	Expect(jsonMap["message"]).To(Equal("StringableSlice test"))
+	Expect(jsonMap["source"]).NotTo(BeEmpty())
+	Expect(jsonMap["timestamp"]).NotTo(BeNil())
+	Expect(jsonMap["a collection"]).ToNot(Equal("[]"))
```
electricmonk: seems like you're not testing the actual serialization of the slice here...

instrumentation/basic_logger.go
```diff
@@ -86,6 +86,29 @@ func Stringable(key string, value fmt.Stringer) *Field {
 	return &Field{Key: key, String: value.String(), Type: StringType}
 }
 
+func StringableSlice(key string, values interface{}) *Field {
```
netoneko: I suggest we replace it with something like this to avoid performance hit:

```
func StringableSlice(key string, values ...fmt.Stringer) *Field {
```

instrumentation/basic_logger.go
```diff
@@ -86,6 +86,29 @@ func Stringable(key string, value fmt.Stringer) *Field {
...
+	builder := strings.Builder{}
```
netoneko: Here lies a problem: I think we need to store the whole string slice and not form a string, because we are going to format the slice differently in both human readable formatter and JSON formatter.

Please consider introducing a `[]string` array to the `Field` type and storing it there and implementing support on `Formatter` level as well.

test/eventually.go
```diff
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-const iterations = 20
+const iterations = 100
```
talkol: Let's split to 2 constants, one for `Eventually` and one for `Consistently` and not change the one for `Consistently`

The reason is that we pay the full time for `Consistently` every time so if you increase the iterations times 5, you make all tests that use `Consistently` 5 times longer.. this is painful

instrumentation/basic_logger.go
jlevison: i think the problem with it is that you cannot pass the slice in its original type, so a []*SignedTransactions cannot be passed to that function signature, this is because the slice itself ([]) does not implement string(), but the objects inside.. (i.e. you have no generics, everything must implicitly implement the interfaces)

```
func StringableSlice1(key string, values ...fmt.Stringer) *Field {
	builder := strings.Builder{}
	builder.WriteString("[")

	for i, v := range values {
		builder.WriteString(v.String())
		if i < len(values)-1 {
			builder.WriteString(",")
		}
	}

	builder.WriteString("]")
	return &Field{Key: key, String: builder.String(), Type: StringType}
}
```

instrumentation/basic_logger.go
jlevison: got it, so for humans we can do sort of a json-lint like format and for the machine logger just compressed json - will change

jlevison:
@talkol @electricmonk @netoneko please approve or give more comments, need to merge this already

PR #92, opened 2018-08-13
Feature/statestore latch

ronnno:
Add sync grace to ReadKeys. 

Using generic BlockTracker ready for reuse in other sync grace flows.

Add crossServiceConfig to config objects with queryGraceTimeoutMillis


bootstrap/node.go
```diff
@@ -37,6 +37,8 @@ func NewNode(
 	benchmarkConsensusRoundRetryIntervalMillis uint32, // TODO: move all of the config from the ctor, it's a smell
 	transport gossipAdapter.Transport,
 	stateHistoryRetentionInBlockHeights uint64,
+	querySyncGraceBlockDist uint64,
```
electricmonk: as discussed, these should probably be defined as a `Config` interface in the `BlockTracker`. Please open an issue for that and fix it ASAP.

PR #93, opened 2018-08-13
Feature/statestore merkle

ronnno:
Implemented GetStateHash. 
new package merkle. 
supports simple Merkle Trie implementation with verifiable proofs.


services/statestorage/merkle/merkel_test.go
```diff
@@ -0,0 +1,12 @@
+package merkle
```
talkol: 1. Notice that the name "merkle" is spelled in different two ways in this path

2. If the Merkle libarary will be usable by other code in the project (like Merkle root of the transactions in a block), maybe we should move it from /services/statestorage to /crypto

services/statestorage/merkle/merkle_forest.go
```diff
@@ -0,0 +1,236 @@
...
+
+type RootId uint64
+type Proof []*Node
+
+var zeroValueHash = hash.CalcSha256([]byte{})
+
+type Node struct {
+	path     string
+	value    primitives.Sha256
+	branches [256]primitives.MerkleSha256
```
talkol: 256 I would move to a const to make more readable where this "magic" number comes from

services/statestorage/merkle/merkle_forest.go
```diff
@@ -0,0 +1,236 @@
...
+		path:     path,
+		value:    valueHash,
+		branches: [256]primitives.MerkleSha256{},
+	}
+}
+func (n *Node) hasValue() bool {
+	return !zeroValueHash.Equal(n.value)
+}
+func (n *Node) hash() primitives.MerkleSha256 {
+	serializedNode := fmt.Sprintf("%+v", n)
```
talkol: Recommend not to use `Sprintf` in production as a serialization mechanism, its behavior is not guaranteed to remain constant across Go versions. We should implement simple serialization ourselves

services/statestorage/merkle/merkle_forest.go
```diff
@@ -0,0 +1,236 @@
...
+	"strings"
+)
+
+type RootId uint64
+type Proof []*Node
+
+var zeroValueHash = hash.CalcSha256([]byte{})
+
+type Node struct {
+	path     string
```
talkol: I'm not sure I understand how path works, would it be possible to add a comment on the path format? Also, why is it a string?

services/statestorage/merkle/merkle_forest.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func NewForest() *Forest {
+	return &Forest{
+		roots: map[RootId]primitives.MerkleSha256{0: emptyNodeHash},
+		nodes: map[string]*Node{emptyNodeHash.KeyForMap(): emptyNode},
+	}
+}
+
+func (f *Forest) GetRoot(height RootId) (primitives.MerkleSha256, error) {
```
talkol: To someone unfamiliar with the exact terminology of merkle trees, it's very difficult to understand the API.. maybe a comment above every function like `GetRoot` that explains what it does with an example will help somebody understand merkle trees from this code

services/statestorage/service.go
```diff
@@ -52,7 +56,10 @@ func (s *service) CommitStateDiff(input *services.CommitStateDiffInput) (*servic
 		return &services.CommitStateDiffOutput{NextDesiredBlockHeight: lastCommittedBlock + 1}, nil
 	}
 
+	// if updating state records fails downstream the merkle tree entries will not bother us
+	s.merkle.Update(merkle.RootId(committedBlock), input.ContractStateDiffs)
```
talkol: a bit confusing to all the variable `committedBlock` and not `committedBlockHeight`.. took me a while to figure this out

services/statestorage/service.go
```diff
@@ -108,5 +115,9 @@ func (s *service) GetStateStorageBlockHeight(input *services.GetStateStorageBloc
 }
 
 func (s *service) GetStateHash(input *services.GetStateHashInput) (*services.GetStateHashOutput, error) {
-	panic("Not implemented")
+	value, _ := s.merkle.GetRoot(merkle.RootId(input.BlockHeight))
```
talkol: Can't this fail? What happens if the block height is out of the forrest? like I give a block height of 1 million when the system has 10 blocks

services/statestorage/test/get_state_hash.go
```diff
@@ -0,0 +1,39 @@
...
+	When("changing state", func() {
+		When("adding first key/value", func() {
+			It("merkle root changes", func() {
+				d := newStateStorageDriver(1)
+
+				root1, err := d.service.GetStateHash(&services.GetStateHashInput{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(len(root1.StateRootHash)).ToNot(BeZero())
+
+				d.commitValuePairs("foo", "bar", "baz")
```
talkol: Doesn't commit need to specify the block height? How will we commit in a test the keys for the next block?

services/statestorage/merkle/merkle_forest.go
ronnno: Agreed. Final serialization is pending updates to the spec. 
Adding a //TODO to reflect this

services/statestorage/merkle/merkle_forest.go
ronnno: temporary implementation. path will eventually be a []byte with an additional parity indication. This too is pending final spec updates.

Currently using simple string as substitute. This should not reach production as is. adding //TODO

services/statestorage/service.go
ronnno: In the current implementation it will return an empty byte array.

Now after BlockTracker is in master we can add the latch here and return an error in case we are out of range.

services/statestorage/test/get_state_hash.go
ronnno: this is a test driver. it conveniently commits at the next block height. 
there are corresponding methods for committing at specific block heights

services/statestorage/merkle/merkle_forest.go
```diff
@@ -9,15 +9,17 @@ import (
 	"strings"
 )
 
-type RootId uint64
+type TrieId uint64
 type Proof []*Node
 
+const trieRadix = 256 // base of the merkle trie. TODO change to 16
```
jlevison: const should be PascalCase

services/statestorage/merkle/merkle_forest.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/orbs-network/orbs-spec/types/go/primitives"
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+	"github.com/pkg/errors"
+	"strings"
+)
+
+type TrieId uint64
+type Proof []*Node
+
+const trieRadix = 256 // base of the merkle trie. TODO change to 16
```
jlevison: const should be PascalCase

PR #94, opened 2018-08-14
Feature/migrate block storage tests to testify

netoneko:
Depends on #88 

I remove gingko and gomega from block storage tests and introduced a 2s timeout per test which is applied during build (our slowest test is under 600ms).


test.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash -xe
 
-go test ./...
+go test -timeout 100ms ./...
```
talkol: This timeout is too low

We have tests that have `Eventually` or `Consistently` that take longer to run.. let's change this to 1 second

services/blockstorage/test/transaction_receipt_test.go
```diff
@@ -1,82 +1,65 @@
...
-				Expect(err).ToNot(HaveOccurred())
-				Expect(out.TransactionReceipt).ToNot(BeNil())
-				Expect(out.TransactionReceipt.Txhash()).To(Equal(txHash))
-				Expect(out.BlockHeight).To(BeEquivalentTo(1))
-				Expect(out.BlockTimestamp).To(Equal(block.ResultsBlock.Header.Timestamp()))
-			})
+	require.NoError(t, err)
+	require.Nil(t, out.TransactionReceipt)
+	require.EqualValues(t, out.BlockHeight, 1)
+	require.EqualValues(t, out.BlockTimestamp, block.ResultsBlock.Header.Timestamp())
```
talkol: The expectation and actual are mixed, require takes the expectation first and the actual second. This needs to be changed in many other places because otherwise the failure messages of these tests will confuse the developer getting them

services/blockstorage/test/transaction_receipt_test.go
```diff
@@ -1,82 +1,65 @@
...
 
-				Expect(err).ToNot(HaveOccurred())
-				Expect(out.TransactionReceipt).ToNot(BeNil())
-				Expect(out.TransactionReceipt.Txhash()).To(Equal(txHash))
-				Expect(out.BlockHeight).To(BeEquivalentTo(1))
-				Expect(out.BlockTimestamp).To(Equal(block.ResultsBlock.Header.Timestamp()))
-			})
+	require.NoError(t, err)
+	require.Nil(t, out.TransactionReceipt)
+	require.EqualValues(t, out.BlockHeight, 1)
```
talkol: We usually like to add a message for every assertion to give context to this comparison so the error will not be "1 is not equal 2" - but "blockheight should match" instead. I don't know if you have the energy to add these to all existing tests, but let's make a habit to add them on new tests

services/blockstorage/test/harness.go
```diff
@@ -9,24 +9,26 @@ import (
...
 	csdOut := &services.CommitStateDiffOutput{}
 
 	d.stateStorage.When("CommitStateDiff", mock.Any).Return(csdOut, nil).Times(1)
-
 }
 
 func (d *driver) verifyMocks() {
 	_, err := d.stateStorage.Verify()
-	Expect(err).ToNot(HaveOccurred())
+	require.NoError(d.t, err)
```
talkol: Let's not store `t` on your driver. 

The convention we use is to pass `t` to `verifyMocks()` as an argument

talkol:
Notice that in lots of the code that the expectation and actual are mixed, require takes the expectation first and the actual second. This needs to be changed in many places because otherwise the failure messages of these tests will confuse the developer getting them

talkol:
Another general comment.. We usually like to add a message for every assertion to give context to this comparison so the error will not be "1 is not equal 2" - but "blockheight should match" instead. I don't know if you have the energy to add these to all existing tests, but let's make a habit to add them on new tests. All assertions of testify take a message as their last arg

services/blockstorage/test/harness.go
netoneko: We have multiple calls to t in many driver methods, that's why it's contained in there. We did not have to do it before because we used to have gingko/gomega.

test.sh
netoneko: I upped it to 2s and forgot to push. I agree that 100ms is way too low.

services/blockstorage/test/transaction_receipt_test.go
netoneko: Done.

services/blockstorage/test/transaction_receipt_test.go
netoneko: 👍

PR #97, opened 2018-08-14
implemented most of the functionality of GetTransactionsForOrdering

electricmonk:
left to implement:
- block and wait for block height (grace sync)
- omit committed transactions from batch


electricmonk:
- implemented wait for block height
- decided not to omit committed transactions from batch as a tx can never be in both pools at the same time

synchronization/block_tracker.go
```diff
@@ -48,21 +48,21 @@ func (t *BlockTracker) readAtomicHeightAndLatch() (uint64, chan struct{}) {
 
 func (t *BlockTracker) WaitForBlock(requestedHeight primitives.BlockHeight) error {
 
-	requestedHeightUint := uint64(requestedHeight)
+	requestedHeightInt := int64(requestedHeight)
```
ronnno: This will break when requestedHight is larger than the max value of int64. suggest to avoid this cast by using addition instead of subtraction below in line 58.

synchronization/block_tracker.go
```diff
@@ -48,21 +48,21 @@ func (t *BlockTracker) readAtomicHeightAndLatch() (uint64, chan struct{}) {
...
 	currentHeight, currentLatch := t.readAtomicHeightAndLatch()
 
-	if currentHeight >= requestedHeightUint { // requested block already committed
+	if currentHeight >= requestedHeightInt { // requested block already committed
 		return nil
 	}
 
-	if currentHeight < requestedHeightUint-t.graceDistance { // requested block too far ahead, no grace
+	if currentHeight < requestedHeightInt-int64(t.graceDistance) { // requested block too far ahead, no grace
```
ronnno: to avoid underflow: currentHeight + t.graceDistance < requestedHeightInt

PR #100, opened 2018-08-14
run acceptance on ci multiple times

talkol:


talkol:
CircleCI issues: `Too long with no output (exceeded 10m0s)`

netoneko:
It seems that locally sometimes 100 is not enough. It's hard to predict at which rate it will fail consistently.

PR #103, opened 2018-08-15
Added reporting to consensus context

IdoZilberberg:
Also replaced Printf()'s with reporting.Info().


services/consensuscontext/service.go
```diff
@@ -38,8 +41,7 @@ func (s *service) RequestNewTransactionsBlock(input *services.RequestNewTransact
 		return nil, err
 	}
 
-	// TODO: add reporting mechanism to this service and convert this print to it
-	fmt.Printf("created Transactions block num-transactions=%d block=%s\n", len(txBlock.SignedTransactions), txBlock.String())
+	s.reporting.Info("created Transactions block", instrumentation.Int("num-transactions", len(txBlock.SignedTransactions)), instrumentation.String("block", txBlock.String()))
```
netoneko: Why `txBlock.String()`? We have `instrumentation.Stringable`.

services/consensuscontext/service.go
```diff
@@ -52,8 +54,7 @@ func (s *service) RequestNewResultsBlock(input *services.RequestNewResultsBlockI
 		return nil, err
 	}
 
-	// TODO: add reporting mechanism to this service and convert this print to it
-	fmt.Printf("created Results block: %s\n", rxBlock.String())
+	s.reporting.Info("created Results block", instrumentation.String("results-block", rxBlock.String()))
```
netoneko: Same here.

services/consensuscontext/service.go
IdoZilberberg: Thanks, fixed.

IdoZilberberg: Thanks, fixed.

services/consensuscontext/service.go
```diff
@@ -38,8 +41,7 @@ func (s *service) RequestNewTransactionsBlock(input *services.RequestNewTransact
 		return nil, err
 	}
 
-	// TODO: add reporting mechanism to this service and convert this print to it
-	fmt.Printf("created Transactions block num-transactions=%d block=%s\n", len(txBlock.SignedTransactions), txBlock.String())
+	s.reporting.Info("created Transactions block", instrumentation.Int("num-transactions", len(txBlock.SignedTransactions)), instrumentation.Stringable("block", txBlock))
```
netoneko: I suggest you use `transactions-block` instead of `block`.

PR #113, opened 2018-08-17
using different belowMinimalBlockDelayMillis values for accaptance an…

electricmonk:
…d e2e suites


talkol:
`queryGraceTimeoutMillis` in the acceptance should be `1`.
see master prior to your merge: https://github.com/orbs-network/orbs-network-go/blob/33fbcdf70f584a4abd7048d40d5211e6ffb0550b/test/harness/network.go#L101

electricmonk:
made some further changes to indicate that different configs are needed for acceptance tests vs production

PR #131, opened 2018-08-21
Feature/statestore remove zero values

ronnno:
depends on #125

- when setting values to zero value (empty byte array) in state storage remove keys from main key value store and from merkle tree.

- avoid storing zero values completely

- any unique state snapshot will always lead to the same markle root.


ronnno:
this request is closed until #125 is accepted

PR #134, opened 2018-08-22
Add LogFailedExpectation

### LINKED ISSUE #57: Support log.Diff to represent expectation mismatch
Expected API:

`logger.Error("protocol version mismatch", log.Diff(protocol, expectedProtocol)`

### LINKED ISSUE #121: ResultsBlockHeader in CommitTransactionReceipts is too wide in scope
`transactionpool.CommitTransactionReceipts` provides an instance of `ResultsBlockHeader` - out of which transaction pool only needs last block height and timestamp.

This creates unnecessary coupling and provides too wide a scope for transaction pool.

Please introduce changes in proto to *only* pass the block height and timestamp.

yonidavidson:
This function allows you to log a diff between 2 fields.
Can provide the base to fix #57

### DIFF
--- a/instrumentation/log/basic_logger.go
+++ b/instrumentation/log/basic_logger.go
@@ -13,6 +13,7 @@ import (
 
 type BasicLogger interface {
 	Log(level string, message string, params ...*Field)
+	LogFailedExpectation(message string, expected *Field, actual *Field, params ...*Field)
 	Info(message string, params ...*Field)
 	Error(message string, params ...*Field)
 	Metric(name string, params ...*Field)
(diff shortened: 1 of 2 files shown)


instrumentation/log/basic_logger.go
```diff
@@ -264,6 +265,13 @@ func (b *basicLogger) Error(message string, params ...*Field) {
 	b.Log("error", message, params...)
 }
 
+func (b *basicLogger) LogFailedExpectation(message string, expected *Field, actual *Field, params ...*Field) {
+	actual.Key = "actual-" + actual.Key
+	expected.Key = "expected-" + expected.Key
+	newParams:= append(params,expected,actual)
```
jlevison: we should also support the case where params is nil (so only the expected/actual is provided)

netoneko: It will work because `params` are always present even if they are not populated with anything.

electricmonk:
What’s the use case for this? Meaning - what problem are we solving with logging diffs? Or failed expectations?

netoneko:
@electricmonk we have many checks and validations in multiple places and we would like to have universal way to debug if they fail. This pr provides exactly that.

PR #142, opened 2018-08-26
move map of block height->merkle root from lib service

noambergIL:


services/statestorage/service.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/pkg/errors"
 	"sync"
 	"time"
-)
+		)
```
ronnno: go fmt ./...

services/statestorage/service.go
```diff
@@ -55,12 +56,18 @@ func (s *service) CommitStateDiff(input *services.CommitStateDiffInput) (*servic
 
 	s.reporting.Info("trying to commit state diff", log.BlockHeight(commitBlockHeight), log.Int("number-of-state-diffs", len(input.ContractStateDiffs)))
 
-	if lastCommittedBlock := s.lastCommittedBlockHeader.BlockHeight(); lastCommittedBlock+1 != commitBlockHeight {
+	lastCommittedBlock := s.lastCommittedBlockHeader.BlockHeight()
```
ronnno: this was originally inline and was broken only to facilitate debugging. normally i like to keep things easy to debug but we have also tend to prefer inline where a variable would not be used more than once... what do you think  (?)

services/statestorage/service.go
```diff
@@ -31,6 +31,7 @@ type service struct {
 }
 
 func NewStateStorage(config Config, persistence adapter.StatePersistence, reporting log.BasicLogger) services.StateStorage {
+	persistence.WriteMerkleRoot(0, merkle.GetEmptyNodeHash()) // TODO genesis block merkle here
```
ronnno: if in `merkle_forest.go` we interpreted a `nil` merkle root as if it was the empty trie hash we could avoid this initialization. we would gain:
* no need for a public `GetEmptyNodeHash()` (or rename it). it can remain encapsulated in `merkle_forest.go`
* we would not need to hard code the genesis block merkle root here. we could let the init flow just call `CommitStateDiff` for block height 0 and it would work as if the merkle root was hard coded.

the down side of this idea is that wherever we currently have `f.nodes[rootHash.KeyForMap()]` in `merkle_forest.go` we would need to replace with a call to a function that would check for `rootHash == nil` and replace it with the empty hash accordingly. this is less elegant on the merkle lib side. 

what do you think?

services/statestorage/service.go
```diff
@@ -55,12 +56,18 @@ func (s *service) CommitStateDiff(input *services.CommitStateDiffInput) (*servic
...
+	lastCommittedBlock := s.lastCommittedBlockHeader.BlockHeight()
+	if  lastCommittedBlock+1 != commitBlockHeight {
 		return &services.CommitStateDiffOutput{NextDesiredBlockHeight: lastCommittedBlock + 1}, nil
 	}
 
 	// if updating state records fails downstream the merkle tree entries will not bother us
-	s.merkle.Update(input.ContractStateDiffs)
+	if root, err := s.persistence.ReadMerkleRoot(commitBlockHeight-1); err != nil { // TODO use input.resultheader.preexecutuion
+		return nil, errors.Errorf("cannot find previous block merkle root. current block %d", commitBlockHeight)
+	} else {
```
ronnno: no need for else here. but it would require declaring root and err before the if.

services/statestorage/service.go
noambergIL: found a common ground solution that is acceptable by majority of consensus

PR #148, opened 2018-08-28
Feature/statestore refactor tests

ronnno:
* create TamperingStatePersistence with blocktracker functionality
* TamperingStatePersistence uses InMemoryStatePersistence internally
* refactor merkle tests to remove redundant code


services/statestorage/merkle/merkle_forest_test.go
```diff
@@ -29,137 +27,68 @@ import (
...
 func TestAddSingleEntryToEmptyTree(t *testing.T) {
-	f := newForestDriver()
-	rootId := updateStringEntries(f, "bar", "baz")
-	require.EqualValues(t, 1, rootId, "unexpected root id")
+	f, root := NewForest()
+	topRoot := updateStringEntries(f, root, "bar", "baz")
 
-	getProofRequireHeight(t, f, rootId, "", "bar", 1)
+	getProofRequireHeight(t, f, topRoot, "", "bar", 1)
 }
```
noambergIL: i think this test should be first one, and add require that topRoot != root

services/statestorage/merkle/merkle_forest_test.go
```diff
@@ -29,137 +27,68 @@ import (
...
-	topRootOf1, err1 := f.getTopRootHash()
-	updateStringEntries(f, "second", "val")
-	rootOfOneAfterSecondUpdate, err2 := f.getRootHash(1)
-
-	require.NoError(t, err1, "GetTopHash failed in first call")
-	require.NoError(t, err2, "GetTopHash failed in second call")
-	require.Equal(t, rootOfOneAfterSecondUpdate, topRootOf1, "getRootHash did not return expected hash")
-}
-
 func TestRootChangeAfterStateChange(t *testing.T) {
```
noambergIL: second test can be adding to tree with root that is not exists ... and check that fails.

services/statestorage/merkle/merkle_forest_test.go
ronnno: moved to top

services/statestorage/merkle/merkle_forest_test.go
ronnno: added a test called TestUpdateTrieFailsForMissingBaseNode for this purpose.
the test required moving the allocation of the empty node hash again to inside the scope of NewForest() because inside the new test I was editing the value of the root hash. When using a global variable this caused this corruption to leak to other test.

PR #158, opened 2018-08-30
Block Tracker - refactor unit tests

ronnno:
This is an attempt to remove all test related code from production code of BlockTracker.

This is work in progress, PR is opened to get review.

The first commit moves the implementation of the test related logic from BlockTracker to unit tests, and leaves a minimal footprint inside production code in the form of a private callback function member - used by unit tests alone.

This is cleaner than before but we still have a member in production code. 
A major disadvantage is that it leaves the unit test code more complicated than it should be.

The main advantage of this implementation is that it keeps the tests as fast as possible. we don't waste any time on waiting or sleeping.

An alternative solution was proposed by @talkol  
Tal suggested to use test.Consistently() and test.Eventually() instead of the callback. 
This will remove all traces of testing code from production code, and will make the tests much more explicit, concise and readable.

The downside of @talkol 's suggestion is that it will make the tests longer by a few 100 Millis.

@erankirsh, @electricmonk what do you think?

### DIFF
--- a/synchronization/block_tracker.go
+++ b/synchronization/block_tracker.go
@@ -15,9 +15,7 @@ type BlockTracker struct {
 	currentHeight uint64 // this is not primitives.BlockHeight so as to avoid unnecessary casts
 	latch         chan struct{}
 
-	// following fields are for tests only
-	enteredSelectSignalForTests chan int
-	selectIterationsForTests    int
+	fireOnWait func() // used by unit test
 }
 
 func NewBlockTracker(startingHeight uint64, graceDist uint16, timeout time.Duration) *BlockTracker {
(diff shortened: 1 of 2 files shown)


synchronization/block_tracker.go
```diff
@@ -15,9 +15,7 @@ type BlockTracker struct {
 	currentHeight uint64 // this is not primitives.BlockHeight so as to avoid unnecessary casts
 	latch         chan struct{}
 
-	// following fields are for tests only
-	enteredSelectSignalForTests chan int
-	selectIterationsForTests    int
+	fireOnWait func() // used by unit test
```
electricmonk: this is a non-exported field; is this feature *only* used by the block tracker's own unit tests?

PR #164, opened 2018-09-02
Feature/better timer

jlevison:


synchronization/timer.go
```diff
@@ -0,0 +1,31 @@
+package synchronization
+
+import "time"
+
+type Timer struct {
+	timer *time.Timer
+	C     <-chan time.Time
```
erankirsh: Why is C public?

talkol: It's also public in the original Timer, so I think to maintain the same interface

jlevison: correct, the idea would be that for the buggy operations, you can use the built-in time or our synchronization package without any API changes (or any code changes, just replace the constructing method with ours)

erankirsh: So why don't we use the their channel like the documentation days:
```
if !t.Stop() {
	<-t.C
}
t.Reset(d)
```

synchronization/timer.go
```diff
@@ -0,0 +1,31 @@
+package synchronization
```
electricmonk: I'd document the motivation for this type, including a reference to the golang issue with the standard timer, so that if that issue goes away we can kill our own impl

synchronization/timer_test.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"github.com/orbs-network/orbs-network-go/synchronization"
+	"testing"
+	"time"
+)
+
+func TestTimer_Reset(t *testing.T) {
+	start := time.Now()
+
+	timer := synchronization.NewTimer(10 * time.Millisecond)
```
electricmonk: consider using microseconds; if we have a lot of tests sleeping 10ms each, we're dead by a thousand paper cuts

synchronization/timer_test.go
```diff
@@ -0,0 +1,79 @@
...
+	select {
+	case <-timer.C:
+		t.Error("timer pinged, but it was stopped already")
+	case <-releaseC:
+	}
+}
+
+func TestTimer_GetTimer(t *testing.T) {
+	x := synchronization.NewTimer(time.Second)
+	if x == nil {
```
electricmonk: why aren't you using testify?

synchronization/timer_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	timer := x.GetTimer()
+	if timer == nil {
+		t.Error("timer inner object was not created, something is really wrong")
+	}
+}
+
+func TestValidTimerStopOp(t *testing.T) {
+	// testing that a timer stopped before it was expired does not later ping
+	timer := synchronization.NewTimer(5 * time.Millisecond)
```
electricmonk: again, microseconds if possible

synchronization/timer.go
jlevison: sure, good idea

synchronization/timer_test.go
jlevison: correct, becomes flaky at microseconds, but i reduced everything to what seems stable (all the tests now take 10ms to complete

synchronization/timer_test.go
jlevison: correct, becomes flaky at microseconds, but i reduced everything to what seems stable (all the tests now take 10ms to complete

synchronization/timer_test.go
jlevison: to be honest i did not think of it, it seems a simple test, changed it to testify

PR #170, opened 2018-09-03
Improved error prints from cli tool running in external process

talkol:


electricmonk:
@talkol tests fail in CI

PR #183, opened 2018-09-06
Feature/direct gossip adapter

talkol:
See spec of the network protocol here:

https://github.com/orbs-network/orbs-spec/blob/master/encoding/gossip/membuffers-over-tcp.md

### DIFF
--- a/services/gossip/adapter/direct_transport.go
+++ b/services/gossip/adapter/direct_transport.go
@@ -3,9 +3,11 @@ package adapter
 import (
 	"context"
 	"fmt"
+	"github.com/orbs-network/membuffers/go"
 	"github.com/orbs-network/orbs-network-go/config"
 	"github.com/orbs-network/orbs-network-go/instrumentation/log"
 	"github.com/orbs-network/orbs-spec/types/go/primitives"
+	"github.com/orbs-network/orbs-spec/types/go/protocol/gossipmessages"
 	"github.com/pkg/errors"
 	"net"
 	"sync"
@@ -22,10 +24,11 @@ type directTransport struct {
 	config    Config
 	reporting log.BasicLogger
 
+	peerQueues map[string]chan *TransportData // does not require mutex to read
+
 	mutex             *sync.RWMutex
 	transportListener TransportListener
 	serverReady       bool
-	peerQueues        map[string]chan *TransportData
 }
 
 func NewDirectTransport(ctx context.Context, config Config, reporting log.BasicLogger) Transport {
@@ -33,19 +36,24 @@ func NewDirectTransport(ctx context.Context, config Config, reporting log.BasicL
 		config:    config,
 		reporting: reporting.For(log.String("adapter", "gossip")),
 
-		mutex:      &sync.RWMutex{},
 		peerQueues: make(map[string]chan *TransportData),
+
+		mutex: &sync.RWMutex{},
+	}
+
+	// client channels (not under mutex, before all goroutines)
+	for peerNodeKey, peer := range t.config.FederationNodes(0) {
+		if !peer.NodePublicKey().Equal(t.config.NodePublicKey()) {
+			t.peerQueues[peerNodeKey] = make(chan *TransportData)
+		}
 	}
 
 	// server goroutine
 	go t.serverMainLoop(ctx, t.getListenPort())
 
 	// client goroutines
-	t.mutex.Lock()
-	defer t.mutex.Unlock()
 	for peerNodeKey, peer := range t.config.FederationNodes(0) {
 		if !peer.NodePublicKey().Equal(t.config.NodePublicKey()) {
-			t.peerQueues[peerNodeKey] = make(chan *TransportData)
 			peerAddress := fmt.Sprintf("%s:%d", peer.GossipEndpoint(), peer.GossipPort())
 			go t.clientMainLoop(ctx, peerAddress, t.peerQueues[peerNodeKey])
 		}
@@ -62,7 +70,23 @@ func (t *directTransport) RegisterListener(listener TransportListener, listenerP
 }
 
 func (t *directTransport) Send(data *TransportData) error {
-	panic("not implemented")
+	switch data.RecipientMode {
+	case gossipmessages.RECIPIENT_LIST_MODE_BROADCAST:
+		for _, peerQueue := range t.peerQueues {
+			peerQueue <- data
+		}
+	case gossipmessages.RECIPIENT_LIST_MODE_LIST:
+		for _, recipientPublicKey := range data.RecipientPublicKeys {
+			if peerQueue, found := t.peerQueues[recipientPublicKey.KeyForMap()]; found {
+				peerQueue <- data
+			} else {
+				return errors.Errorf("unknown recepient public key: %s", recipientPublicKey.String())
+			}
+		}
+	case gossipmessages.RECIPIENT_LIST_MODE_ALL_BUT_LIST:
+		panic("Not implemented")
+	}
+	return errors.Errorf("unknown recipient mode: %s", data.RecipientMode.String())
 }
 
 func (t *directTransport) getListenPort() uint16 {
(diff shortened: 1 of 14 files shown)


services/gossip/adapter/direct_transport_test.go
```diff
@@ -0,0 +1,278 @@
...
+	test.WithContext(func(ctx context.Context) {
+
+		h := newDirectHarness().start(ctx)
+
+		listener, err := net.Listen("tcp", fmt.Sprintf(":%d", h.portForPeer(0)))
+		defer listener.Close()
+		require.NoError(t, err, "test peer server could not listen")
+
+		connection, err := listener.Accept()
+		defer connection.Close()
```
electricmonk: I'm not sure about this defer. Does it close the connection pointed from `connection` or the specific instance?  if so, then the connection it points to is closed in the first forceful disconnect...

services/gossip/adapter/direct_transport_test.go
```diff
@@ -0,0 +1,278 @@
...
+			RecipientMode:       gossipmessages.RECIPIENT_LIST_MODE_BROADCAST,
+			RecipientPublicKeys: nil,
+			Payloads:            [][]byte{{0x11}, {0x22, 0x33}},
+		})
+		require.NoError(t, err, "adapter Send should not fail")
+
+		for i := 0; i < networkSize-1; i++ {
+			data, err := h.peerListenerReadTotal(i, 20)
+			require.NoError(t, err, "test peer server could not read from local transport")
+			require.Equal(t, []byte{0x02, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x11, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x22, 0x33, 0x00, 0x00}, data)
```
electricmonk: It's really unclear why we're expecting this specific byte buffer here. I'm assuming this has to do with the header, so please introduce a factory method that constructs the expected byte array and takes the payload array:
```
payloads := [][]byte{{0x11}, {0x22, 0x33}}
err := h.transport.Send(&TransportData{
	SenderPublicKey:     h.config.NodePublicKey(),
	RecipientMode:       gossipmessages.RECIPIENT_LIST_MODE_BROADCAST,
	RecipientPublicKeys: nil,
	Payloads:            payloads,
})
require.NoError(t, err, "adapter Send should not fail")

for i := 0; i < networkSize-1; i++ {
	data, err := h.peerListenerReadTotal(i, 20)
	require.NoError(t, err, "test peer server could not read from local transport")
	require.Equal(t, rawMessageFor(payloads), data)
}
```

services/gossip/adapter/direct_transport_test.go
```diff
@@ -0,0 +1,278 @@
...
+func TestDirectOutgoing_SendsKeepAliveWhenNothingToSend(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+
+		h := newDirectHarnessWithConnectedPeers(t, ctx)
+		defer h.cleanupConnectedPeers()
+
+		for numKeepAliveSent := 0; numKeepAliveSent < 2; numKeepAliveSent++ {
+			data, err := h.peerListenerReadTotal(1, 4)
+			require.NoError(t, err, "test peer server could not read keepalive from local transport")
+			require.Equal(t, []byte{0x00, 0x00, 0x00, 0x00}, data) // keepalive content (zero payloads)
```
electricmonk: instead of commenting, just extract to a variable `KeepAliveContent`

services/gossip/adapter/direct_transport_test.go
```diff
@@ -0,0 +1,278 @@
...
+		h.expectTransportListenerCalled([][]byte{{0x11}, {0x22, 0x33}})
+
+		for numKeepAliveReceived := 0; numKeepAliveReceived < 2; numKeepAliveReceived++ {
+			buffer := []byte{0x00, 0x00, 0x00, 0x00} // keepalive content (zero payloads)
+			written, err := h.peerTalkerConnection.Write(buffer)
+			require.NoError(t, err, "test peer could not write to local transport")
+			require.Equal(t, len(buffer), written)
+		}
+
+		buffer := []byte{0x02, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x11, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x22, 0x33, 0x00, 0x00}
```
electricmonk: why do you send a non-keepalive message here?

services/gossip/adapter/direct_transport_test.go
```diff
@@ -0,0 +1,278 @@
...
+		written, err := h.peerTalkerConnection.Write(buffer)
+		require.NoError(t, err, "test peer could not write to local transport")
+		require.Equal(t, len(buffer), written)
+
+		h.verifyTransportListenerCalled(t)
+	})
+}
+
+// wanted to simulate Timeout on Send instead of Error but was unable to
+func TestDirectOutgoing_ErrorDuringSendCausesReconnect(t *testing.T) {
```
electricmonk: I did not understand what you're testing here

test/harness/services/gossip/adapter/transport_contract_test.go
```diff
@@ -66,28 +72,55 @@ func aTamperingTransport() *transportContractContext {
...
 	}
 	res.listeners = []*mockListener{
 		listenTo(res.transports[0], res.publicKeys[0]),
 		listenTo(res.transports[1], res.publicKeys[1]),
 		listenTo(res.transports[2], res.publicKeys[2]),
 		listenTo(res.transports[3], res.publicKeys[3]),
 	}
+
+	// TODO: improve this, we need some time until everybody connects to everybody else
+	// TODO: maybe add an adapter function to check how many active outgoing connections we have
```
electricmonk: Adapter could take a `ConnectionListener` that gets notified on connects/disconnects, and the test could provide such a listener to block until the desired number of connections has been reached

services/gossip/adapter/direct_transport.go
```diff
@@ -0,0 +1,399 @@
...
+		panic("Not implemented")
+	}
+	return errors.Errorf("unknown recipient mode: %s", data.RecipientMode.String())
+}
+
+func (t *directTransport) getListenPort() uint16 {
+	nodePublicKey := t.config.NodePublicKey()
+	nodeConfig, found := t.config.FederationNodes(0)[nodePublicKey.KeyForMap()]
+	if !found {
+		err := errors.Errorf("fatal error - gossip configuration (port and endpoint) not found for my public key: %s", nodePublicKey.String())
```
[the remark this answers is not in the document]
talkol: Split federation nodes and gossip peers config - https://github.com/orbs-network/orbs-network-go/commit/cc29d916e9b275be6db206ae824e35e192b50992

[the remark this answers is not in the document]
talkol: Gave gossip listen port its own config key - https://github.com/orbs-network/orbs-network-go/commit/9d525b5d38b284507415ec907222a490a5e02fb3

services/gossip/adapter/direct_transport.go
```diff
@@ -0,0 +1,399 @@
...
+		t.reporting.Error(err.Error())
+		panic(err)
+	}
+	t.reporting.Info("gossip transport server listening", log.Uint32("port", uint32(listenPort)))
+
+	for {
+		conn, err := listener.Accept()
+		if err != nil {
+			if !t.isServerReady() {
+				t.reporting.Info("incoming connection accept stopped since server is shutting down")
```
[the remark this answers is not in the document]
talkol: No, the server is started and listening (ready) after the line `serverListenForIncomingConnections()` which is before the for loop. I renamed `isServerReady` to `isServerListening` to make this clearer

services/gossip/adapter/direct_transport_test.go
talkol: There's already a separate issue for this problem:
https://github.com/orbs-network/orbs-network-go/issues/187

I added your issue there as well.

We should resolve this separately because I invested significant efforts to make sure all resources are closed yet this still happens.. require further investigation and I don't want to block this PR

services/gossip/adapter/direct_transport_test.go
talkol: that when the adapter is sending data to a peer and the send fails due to a socket error (or a socket timeout) - the adapter disconnects and reconnects the outgoing socket

test/harness/services/gossip/adapter/transport_contract_test.go
talkol: Added your proposal to the TODO, will resolve outside this PR

PR #208, opened 2018-09-12
Feature/add time to test sh

IdoZilberberg:
Use this to run specific component tests within Docker.


netoneko:
We decided to move acceptance to a different stage of the build (see #215). If we want to reuse any of this code, it should be cleaned up and refactored.

IdoZilberberg:
@netoneko This version of test.sh allows you to run component tests on Docker on local machine. We use them to identify flaky tests in specific components. How do you suggest to run them differently?

IdoZilberberg:
After talking to @netoneko, decided to open another PR instead of this one.

PR #217, opened 2018-09-13
Add benchmark tests for logger

netoneko:


electricmonk:
seems like `os.Stdout` is much, much slower than `ioutil.Discard` - which makes sense, but then it begs the question of how much does the logger slow down our tests...

PR #222, opened 2018-09-17
Added component test script under Docker

IdoZilberberg:


test.sh
```diff
@@ -4,3 +4,6 @@ source ./test.common.sh
 
 go test -timeout 3m ./... -failfast > test.out
 check_exit_code_and_report
+
+# Uncomment to run component tests
+# ./test.components.sh
```
electricmonk: do we want this to run as yet another parallel workflow? @netoneko wdyt?

netoneko: Sure, we can do that as well.

PR #226, opened 2018-09-17
Feature/deployable go contracts

talkol:
We (@netoneko and I) are aware that this is red in CI and doesn't pass E2E

After a long investigation today, our best suspicion that some problem in memberlist transport causes the transport to start misbehaving and swallow some network messages from some point in the test

We debugged it until we reached memberlist transport implementation and then gave up

Then @netoneko had the great idea to merge my new transport and see if the problem disappears.. it did.. which reduced our motivation to investigate further

Our recommendation is this:
1. Submit this PR (deployable go) and start review
2. Address the feedback from my transport PR review
3. Merge my transport
4. Address this PR feedback
5. Merge this PR after my transport was merged to master so we can see it's green (fingers crossed)


docker-compose.yml
```diff
@@ -67,7 +67,7 @@ services:
   orbs-e2e:
     image: orbs:e2e
     environment:
-      API_ENDPOINT: http://orbs-network-node-3:8080/api/
+      API_ENDPOINT: http://orbs-network-node-1:8080/api/
```
netoneko: We need to revert this change so we could test that non-leader node gets updated.

talkol: yes, I remember

talkol:
closing since this branch was corrupted with extra sambusac commits, will open a new PR with the reverted branch

PR #237, opened 2018-09-20
Refactor file based config

netoneko:
JSON configuration is now supported in main executable, refactored hardcoded config initialization.

I need to do minor refactor including separation of federation nodes and gossip peers.


config/factory.go
```diff
@@ -101,6 +82,61 @@ func EmptyConfig() MutableNodeConfig {
...
+	cfg.SetDuration(TRANSACTION_POOL_COMMITTED_POOL_CLEAR_EXPIRED_INTERVAL, 30*time.Second)
+
+	cfg.SetUint32(GOSSIP_LISTEN_PORT, 4400)
+
+	cfg.SetString(PROCESSOR_ARTIFACT_PATH, filepath.Join(GetProjectSourceTmpPath(), "processor-artifacts"))
+
+	return cfg
+}
+
+func (c *config) OverrideNodeSpecificValues(
```
electricmonk: This method signature is confusing. When you return an argument, I assume the method has no side effects (is referentially transparent), which isn't the case. Either make this a copying method (leaving the original receiver unchanged, or change the signature to a void

config/file_config_test.go
```diff
@@ -95,12 +96,35 @@ func TestSetFederationNodes(t *testing.T) {
...
 	node1 := &hardCodedFederationNode{
-		nodePublicKey:  keyPair.PublicKey(),
+		nodePublicKey: keyPair.PublicKey(),
+	}
+
+	require.EqualValues(t, node1, cfg.FederationNodes(0)[keyPair.PublicKey().KeyForMap()])
+}
+
+func TestSetGossipPeers(t *testing.T) {
+	cfg, err := NewEmptyFileConfig(FILE_CONFIG_CONTENTS)
```
electricmonk: You have several tests all sharing the same config stub. What might happen is that all tests fail on parse error if any of the parse method is broken. I'd prefer creating an inline stub for the minimal config required by each individual test

example-config/node1.json
```diff
@@ -0,0 +1,10 @@
+{
```
electricmonk: is this an example config or is this the config used by the docker tests?

test/e2e/harness.go
```diff
@@ -73,18 +73,18 @@ func newHarness() *harness {
...
 				uint16(firstRandomPort+i),
+				nodeKeyPair.PublicKey(),
+				nodeKeyPair.PrivateKey(),
 				leaderKeyPair.PublicKey(),
-				consensus.CONSENSUS_ALGO_TYPE_BENCHMARK_CONSENSUS,
-				logger,
-				processorArtifactPath,
-			)
+				consensus.CONSENSUS_ALGO_TYPE_BENCHMARK_CONSENSUS)
+			cfg.SetString(config.PROCESSOR_ARTIFACT_PATH, processorArtifactPath)
```
electricmonk: I don't like this. It should be part of the factory method somehow. Please try to find a better way, if we leave it here it's broken windows syndrome.

PR #248, opened 2018-09-20
Feature/processor native adapter flakiness

talkol:
Hopefully this fixes some issues (go 1.11 for example)


talkol:
Not sure about this anymore, for some reason this is increasing flakiness?

services/processor/native/adapter/native_compiler.go
```diff
@@ -71,7 +71,7 @@ func getHashOfCode(code string) string {
 
 func writeSourceCodeToDisk(filenamePrefix string, code string, artifactsPath string) (string, error) {
 	dir := filepath.Join(artifactsPath, SOURCE_CODE_PATH)
-	err := os.MkdirAll(dir, 0700)
+	err := os.MkdirAll(dir, 0755)
```
netoneko: Why do we need this change? I don't see a reason that anyone other than the caller should have access to the directory.

netoneko:
@talkol there is still a problem with compilation:

```
error 2018-09-25T15:29:08.599741Z warm up compilation on init failed node=dfc06c5be24a67adee80b35ab4f147bb1a35c55ff85eda69f40ef827bddec173 adapter=processor-native error=error building go source:  function=adapter.NewNativeCompiler source=/go/src/github.com/orbs-network/orbs-network-go/services/processor/native/adapter/native_compiler.go:44 
```

I see it in the logs from time to time. I think it's a good approach to pass external configuration limit for compiling warm up contract.

PR #251, opened 2018-09-22
Bugfix/fix merkel tree memory leak

yonidavidson:
Naive fix for memory leak.
the manual testing shows memory leak is gone in the merkel tree


electricmonk:
@noambergIL @ronnno IIRC you're making a full clone of the state on purpose. Is that so?

noambergIL:
Yes. This is the early simple version.
Easy version is to delete all older than config clones

On 22 Sep 2018 13:02, "Shai Yallin" <notifications@github.com> wrote:

@noambergIL <https://github.com/noambergIL> @ronnno
<https://github.com/ronnno> IIRC you're making a full clone of the state on
purpose. Is that so?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/orbs-network/orbs-network-go/pull/251#issuecomment-423732033>,
or mute the thread

talkol:
Further investigation regarding the memory issues today revealed that the problem in the state storage is smaller than we thought. The bigger problem was that on `-count 100` testa, the node itself was leaking and wasn't freed. This meant that we were accumulating memory during the test. It's true that most of this memory is related to Merkle state store, but the big issue wasn't the deep clone, it was the fact that the entire node was leaking.

What's the source of the leak? During graceful shutdown of the node, some goroutines leaked and weren't closed, causing references to the node to be kept which meant the node itself would not be freed on GC. Since the node holds reference to the state storage (and this is most of the consumed memory) - it looked like the leak in the state storage is the main cause.

We fixed the leaking node today in a series of PRs and the memory issue disappeared.

@noambergIL and @ronnno have a different plan on how to improve the state store implementation.. so I'm closing this PR without merging.

yonidavidson:
@talkol thanks for the detailed explanation.

PR #255, opened 2018-09-25
Refactor tampering transport mutex not to call other services when mu…

netoneko:
…tex is locked


talkol:
@electricmonk please go over this after the fact just to see what we changed

electricmonk:
@netoneko @talkol not sure I understand the reasoning for this change. How did this fix the leaking goroutine? why was it leaking?

PR #258, opened 2018-09-25
HTTP grafecul shutdown + blocking socket binding

netoneko:


electricmonk:
@noambergIL what was the motivation behind this change?

noambergIL:
Kiril & Tal saw that there is still a bit of flakiness when sometimes the http server was not up when the first e2e messages where used and they saw online that it was suggested to separate the listen and serve part as the listen is synchronous and thus blocks until its ready. (also they saw a todo graceful shutdown i forgot to add).

PR #268, opened 2018-09-26
Feature/public api mutex

noambergIL:
replace waiter with event loop on channels to a map under mutex of channels.
please don't review the waiter.go and waiter_test.go as diff but read code itself


services/publicapi/service.go
```diff
@@ -49,12 +52,27 @@ func NewPublicApi(
...
+				transactionReceipt: txReceipt,
+				blockHeight:        input.BlockHeight,
+				blockTimestamp:     input.Timestamp,
+			}})
 	}
 	return &handlers.HandleTransactionResultsOutput{}, nil
 }
 
 func (s *service) HandleTransactionError(input *handlers.HandleTransactionErrorInput) (*handlers.HandleTransactionErrorOutput, error) {
+	//s.waiter.complete(input.Txhash.KeyForMap(),
```
erankirsh: Why don't you implement this?

services/publicapi/waiter_test.go
```diff
@@ -3,105 +3,209 @@ package publicapi
...
+		go waitTillCancelled(wc1)
+		go waitTillCancelled(wc2)
 	})
-	<-done
-	<-done
+	<-c
+	<-c
 }
+
+// TODO  MUTEX TESTS ? and/or timing tests
```
erankirsh: Please review your tests - I guess this wasn't TDD, but I see that delete by key is never tested (For instance.)

services/publicapi/waiter_test.go
```diff
@@ -3,105 +3,209 @@ package publicapi
...
 	})
 }
 
-func TestWaitForTransaction_GracefulShutdownFreesAllWaitingGoroutines(t *testing.T) {
-	var w *txWaiter
-	done := make(chan struct{})
-	test.WithContext(func(ctx context.Context) {
-		w = newTxWaiter(ctx)
+func TestPublicApiWaiter_WaitGracefulShutdown(t *testing.T) {
+	var waiter *waiter
```
erankirsh: I think these tests can be parallel, would be nice to add.

services/publicapi/service.go
noambergIL: shai's code is not in yet - there is no tx hash so i cannot actually use my code.

Issue #266, opened 2018-09-26
Monitoring system phase 1


electricmonk:
Establish a monitoring system (New Relic? Graphana? something else?). Create separate accounts (tenants):
- testnet
- jepsen
- CI

Only implement CI for now. Let's collect metrics for all HTTP calls for starters, and discuss other KPIs with the team.

electricmonk:
add metrics for running a contract - metric per contract name + method name

electricmonk:
https://godoc.org/github.com/go-kit/kit/metrics

PR #269, opened 2018-09-26
More metrics

netoneko:
Related to #266

Updating default configuration in 83a9bf6 was blessed by @talkol


services/consensuscontext/create_block.go
```diff
@@ -9,6 +9,9 @@ import (
 )
 
 func (s *service) createTransactionsBlock(blockHeight primitives.BlockHeight, prevBlockHash primitives.Sha256) (*protocol.TransactionsBlockContainer, error) {
+	meter := s.reporting.Meter("tx-block-creation")
```
electricmonk: could be a one-liner

services/publicapi/service.go
```diff
@@ -65,6 +65,9 @@ func (s *service) SendTransaction(input *services.SendTransactionInput) (*servic
 	s.reporting.Info("transaction received via public api", log.String("flow", "checkpoint"), log.Stringable("txHash", txHash))
 	defer s.reporting.Info("transaction status returned by public api", log.String("flow", "checkpoint"), log.Stringable("txHash", txHash))
 
+	meter := s.reporting.Meter("tx-processing-time", log.Stringable("txHash", txHash))
+	defer meter.Done()
```
electricmonk: I'm not sure this is good enough. Do we want to collect different metrics, or to qualify at least, the difference between a tx that has finished successfully, a tx that timed out (i.e. result did not return within timeout interval) and a tx that failed? if so, how would we do that?

services/publicapi/test/public_api_test.go
```diff
@@ -86,6 +86,45 @@ func TestSendTransaction_BlocksUntilTransactionCompletes(t *testing.T) {
 	})
 }
 
+func TestSendTransaction_TimesOut(t *testing.T) {
```
electricmonk: why did you add this test?

netoneko: There wasn't one and I needed to check if the timeout works as expected.

services/publicapi/service.go
netoneko: This is a good point

electricmonk: we can move forward with this, but document the downsides fo this approach in a new issue

services/consensusalgo/benchmarkconsensus/leader.go
```diff
@@ -26,6 +26,8 @@ func (s *service) leaderConsensusRoundRunLoop(ctx context.Context) {
 
 	s.lastCommittedBlock = s.leaderGenerateGenesisBlock()
 	for {
+		meter := s.logger.Meter("consensus-round-tick")
```
electricmonk: here as well, a better approach would be to measure distinct branches of behavior: one for successful round, one for timeout and one for error. The API for `Meter` might need to change - can you think of a better approach?

services/publicapi/test/public_api_test.go
```diff
@@ -110,13 +112,54 @@ func newPublicApiConfig(txTimeout time.Duration) publicapi.Config {
...
+		vmMock := &services.MockVirtualMachine{}
+
+		timeoutDuration := 1 * time.Millisecond
+
+		cfg := newPublicApiConfig(timeoutDuration)
+		papi := publicapi.NewPublicApi(ctx, cfg, txpMock, vmMock, logger)
+
+		blockTime := primitives.TimestampNano(time.Now().Nanosecond())
+		txb := builders.Transaction().Builder()
+		txpMock.When("AddNewTransaction", mock.Any).Times(1).
```
electricmonk: take a look at master for this code, @noambergIL introduced a test harness, make sure you adapt to it

services/publicapi/test/public_api_test.go
```diff
@@ -110,13 +112,54 @@ func newPublicApiConfig(txTimeout time.Duration) publicapi.Config {
...
+		tx, err := papi.SendTransaction(&services.SendTransactionInput{
+			ClientRequest: (&client.SendTransactionRequestBuilder{
+				SignedTransaction: txb,
+			}).Build(),
+		})
+
+		txHash := digest.CalcTxHash(txb.Build().Transaction())
+
+		require.EqualError(t, err, fmt.Sprintf("timed out waiting for transaction result %s", txHash.String()))
+		require.WithinDuration(t, time.Now(), start, timeoutDuration*2, "timeout duration exceeded")
```
electricmonk: very nice way to enforce timeouts! can you share this approach on Slack? I for one didn't think of this

test/e2e/harness.go
```diff
@@ -65,7 +65,13 @@ func newHarness() *harness {
 			gossipPeers[publicKey.KeyForMap()] = config.NewHardCodedGossipPeer(uint16(firstRandomPort+i), "127.0.0.1")
 		}
 
-		logger := log.GetLogger().WithOutput(log.NewOutput(os.Stdout).WithFormatter(log.NewHumanReadableFormatter()))
+		os.MkdirAll(config.GetProjectSourceRootPath()+"/logs", 0755)
+
+		logger := log.GetLogger().WithTags(
+			log.String("_test", "e2e"),
+			log.String("_branch", os.Getenv("GIT_BRANCH")),
```
electricmonk: these two tags can be added automatically during enrichment, where you add the caller info

netoneko: I prefer this code to be explicitly outside for now. If we see that we use it in multiple places I will move it.

services/publicapi/test/public_api_test.go
netoneko: Done.

services/consensusalgo/benchmarkconsensus/leader.go
netoneko: I think it makes sense, we should rethink the API.

services/publicapi/test/public_api_test.go
netoneko: Done.

netoneko:
test

PR #275, opened 2018-09-26
Issue/contract errors

talkol:
Resolves https://github.com/orbs-network/orbs-contract-sdk/issues/2

### DIFF
--- a/services/processor/native/service.go
+++ b/services/processor/native/service.go
@@ -50,7 +50,7 @@ func (s *service) ProcessCall(input *services.ProcessCallInput) (*services.Proce
 	contractInfo, methodInfo, err := s.retrieveContractAndMethodInfoFromRepository(executionContextId, string(input.ContractName), string(input.MethodName))
 	if err != nil {
 		return &services.ProcessCallOutput{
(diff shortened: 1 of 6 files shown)


services/processor/native/service.go
```diff
@@ -50,7 +50,8 @@ func (s *service) ProcessCall(input *services.ProcessCallInput) (*services.Proce
 	contractInfo, methodInfo, err := s.retrieveContractAndMethodInfoFromRepository(executionContextId, string(input.ContractName), string(input.MethodName))
 	if err != nil {
 		return &services.ProcessCallOutput{
-			OutputArgumentArray: (&protocol.MethodArgumentArrayBuilder{}).Build(),
+			// TODO: discussion, showing the error string in call output will store this on the transactions on-chain but may be difficult to reproduce during audit if code changes
```
netoneko: Errors might leak sensitive information, so I am not sure we should store it on chain.

talkol: This is a spec decision, opened an issue there:
https://github.com/orbs-network/orbs-spec/issues/97

PR #278, opened 2018-09-27
Log acceptance tests to files

netoneko:
This PR drastically decreases memory usage of the acceptance suite, something that we initially believed was caused by a memory leak.

The actual problem was `go test` process capturing stdout and holding it in memory.

### DIFF
--- a/test.acceptance.sh
+++ b/test.acceptance.sh
@@ -2,5 +2,5 @@
 
 source ./test.common.sh
 
-STANDALONE=true go test ./test/acceptance -count 100 -timeout 6m -failfast > test.out
+NO_LOG_STDOUT=true go test ./test/acceptance -count 100 -timeout 6m -failfast > test.out
 check_exit_code_and_report
--- a/test.acceptance.sh
+++ b/test.acceptance.sh
@@ -2,5 +2,5 @@
 
 source ./test.common.sh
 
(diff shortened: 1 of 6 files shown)


netoneko:
Please don't merge this, there are some permissions issues while uploading acceptance logs.

test.acceptance.sh
```diff
@@ -2,5 +2,5 @@
 
 source ./test.common.sh
 
-go test ./test/acceptance -count 100 -timeout 6m -failfast > test.out
+STANDALONE=true go test ./test/acceptance -count 100 -timeout 6m -failfast > test.out
```
talkol: `STANDALONE` is a very confusing name for this env variable. 
Maybe rename it to `NO_LOG_STDOUT`? It's a much clearer name that says exactly what it does..

talkol:
In `test.acceptance.sh` we have this code:

```
go test ./test/acceptance -count 100 -timeout 6m -failfast > test.out
```

and in `test.common.sh` the function `check_exit_code_and_report` does:

```
grep -B 150 -A 15 -- "FAIL:" test.out > fail.out
        cat fail.out

        grep -B 150 -A 150 -- "test timed out" test.out > timed.out
        cat timed.out
```

This code greps on `test.out` which is redirection of STDOUT but we no longer log to STDOUT

test.acceptance.sh
netoneko: Fixed

PR #279, opened 2018-09-27
Issue/276 concurrent map access in tx pool

### LINKED ISSUE #276: Concurrent map access in tx pool
When removing logging from the system, we fail on concurrent map access for this line (read while map is being written):

```
func (p *committedTxPool) get(txHash primitives.Sha256) *committedTransaction {
	key := txHash.KeyForMap()

	tx := p.transactions[key]

	return tx
}
```

Please make sure you use a `RWMutex` and add `RLock` in this function

electricmonk:
fixes #276


electricmonk:
@noambergIL please approve PR

PR #287, opened 2018-09-28
More comprehensive log filters

electricmonk:
1) Introduced polymorphic log filters
2) Renamed `reporting` to `logger` across project
3) Added filters to remove expected log messages from acceptance tests so that real errors are more visible

Note that filters might be too restrictive, so after merging this PR we might need to tweak them a bit more

### DIFF
--- a/instrumentation/log/basic_logger.go
+++ b/instrumentation/log/basic_logger.go
@@ -14,19 +14,19 @@ type BasicLogger interface {
 	Info(message string, params ...*Field)
 	Error(message string, params ...*Field)
 	Metric(name string, params ...*Field)
-	For(params ...*Field) BasicLogger
+	WithTag(params ...*Field) BasicLogger
 	Meter(name string, params ...*Field) BasicMeter
-	Prefixes() []*Field
+	Tags() []*Field
 	WithOutput(writer ...Output) BasicLogger
-	WithFilter(filter *Field) BasicLogger
+	WithFilter(filter Filter) BasicLogger
 }
 
 type basicLogger struct {
 	outputs               []Output
-	prefixes              []*Field
+	tags                  []*Field
 	nestingLevel          int
 	sourceRootPrefixIndex int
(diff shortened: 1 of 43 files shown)


instrumentation/log/filters.go
```diff
@@ -0,0 +1,92 @@
...
+func ExcludeField(field *Field) Filter {
+	return &excludeField{field: field}
+}
+
+func IncludeFieldWithKey(key string) Filter {
+	return &includeFieldWithKey{key: key}
+}
+
+func OnlyErrors() Filter {
+	return &levelMatch{level: "error"}
```
netoneko: It shouldn't discard any line that has field `error` populated through `log.Error`.

instrumentation/log/basic_logger.go
```diff
@@ -14,19 +14,19 @@ type BasicLogger interface {
 	Info(message string, params ...*Field)
 	Error(message string, params ...*Field)
 	Metric(name string, params ...*Field)
-	For(params ...*Field) BasicLogger
+	WithTag(params ...*Field) BasicLogger
```
netoneko: Should be `WithTags` because it takes multiple arguments.

netoneko:
As part of this effort, would you like to introduce some kind of configuration in `main.go` or do you feel it's unnecessary at this point?

electricmonk:
No, I think we might want to have error only for mainnet, full tracing on testnet and whatever the tester wants in tests

instrumentation/log/filters.go
electricmonk: does the log level automatically become "error" if there's an error log? actually, if we use this logic, maybe we can drop the info/error distinction and use the error field heuristic? wdyt?

netoneko: The level does not change because sometimes we want to log the error even in an `info` log because the error is generally not significant. But if we are logging only errors we should still display it.

PR #291, opened 2018-10-02
handleerror uncomment + test

noambergIL:


services/publicapi/test/public_api_test.go
```diff
@@ -86,6 +72,64 @@ func TestSendTransaction_BlocksUntilTransactionCompletes(t *testing.T) {
...
 
+func TestSendTransaction_BlocksUntilTransactionErrors(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		harness := newPublicApiHarness(ctx, 1*time.Second)
+
+		blockTime := primitives.TimestampNano(time.Now().Nanosecond())
+		txb := builders.Transaction().Builder()
+		txHash := digest.CalcTxHash(txb.Build().Transaction())
+
+		harness.txpMock.When("AddNewTransaction", mock.Any).Times(1).
```
electricmonk: you can move the mock setup into the harness, so that the test only needs to supply the callback function. Everything else, including the sleep (yuck) can be hidden in the harness:
```golang
harness.OnAddNewTransaction(func() {
  harness.papi.HandleTransactionError(...)
})
```

services/publicapi/test/public_api_test.go
```diff
@@ -86,6 +72,64 @@ func TestSendTransaction_BlocksUntilTransactionCompletes(t *testing.T) {
...
+						Txhash:            txHash,
+						TransactionStatus: protocol.TRANSACTION_STATUS_REJECTED_TIMESTAMP_WINDOW_EXCEEDED,
+						BlockHeight:       8,
+						BlockTimestamp:    blockTime,
+					})
+				}()
+				return &services.AddNewTransactionOutput{TransactionStatus: protocol.TRANSACTION_STATUS_PENDING}, nil
+			})
+
+		tx, err := harness.papi.SendTransaction(&services.SendTransactionInput{
```
electricmonk: this is not `tx`, it's `result`

services/publicapi/test/public_api_test.go
```diff
@@ -86,6 +72,64 @@ func TestSendTransaction_BlocksUntilTransactionCompletes(t *testing.T) {
...
+		txHash := digest.CalcTxHash(txb.Build().Transaction())
+
+		harness.txpMock.When("AddNewTransaction", mock.Any).Times(1).
+			Call(func(input *services.AddNewTransactionInput) (*services.AddNewTransactionOutput, error) {
+				go func() {
+					time.Sleep(1 * time.Millisecond)
+					harness.papi.HandleTransactionError(&handlers.HandleTransactionErrorInput{
+						Txhash:            txHash,
+						TransactionStatus: protocol.TRANSACTION_STATUS_REJECTED_TIMESTAMP_WINDOW_EXCEEDED,
+						BlockHeight:       8,
```
electricmonk: please either assert that blockheight is the same as what you got from `HandleTransactionError` (and `HandleTransactionResults`) or drop these values from the struct

PR #294, opened 2018-10-03
increased max transactions per block in acceptance suite to 30

electricmonk:
shortened acceptance test 100x run time by half


config/factory.go
```diff
@@ -18,6 +18,8 @@ func ForProduction(processorArtifactPath string) MutableNodeConfig {
 	cfg.SetDuration(GOSSIP_NETWORK_TIMEOUT, 30*time.Second)
 	cfg.SetDuration(PUBLIC_API_SEND_TRANSACTION_TIMEOUT, 30*time.Second)
 	cfg.SetDuration(PUBLIC_API_TRANSACTION_STATUS_GRACE, 5*time.Second)
+	cfg.SetUint32(CONSENSUS_CONTEXT_MINIMUM_TRANSACTION_IN_BLOCK, 10)
```
netoneko: I think before we do this we need to add some logic that still produces a block if certain timeout was reached so the system won't be unresponsive under low load.

Please consider setting it to `1` for now.

electricmonk: I think there is such a logic:
https://github.com/orbs-network/orbs-network-go/blob/76aafbcdcb2ce7c3fb31745e563c010a49ac991c/services/consensuscontext/fetch_transactions.go#L25
Am I missing something?

netoneko: You are right, we don't check for minimal transactions for the second time.

PR #295, opened 2018-10-03
Add the get tx status functionality

noambergIL:
update the spec and config
add the get status function
add and order tests
update the http server


bootstrap/httpserver/server_handlers_test.go
```diff
@@ -107,3 +107,45 @@ func TestHttpServerCallMethod_Error(t *testing.T) {
...
+
+	s := NewHttpServer("", logger, papiMock)
+
+	request := (&client.GetTransactionStatusRequestBuilder{}).Build()
+
+	req, _ := http.NewRequest("POST", "", bytes.NewReader(request.Raw()))
+	rec := httptest.NewRecorder()
+	s.(*server).getTransactionStatusHandler(rec, req)
+
+	require.Equal(t, http.StatusInternalServerError, rec.Code, "should fail with 500")
```
electricmonk: did you look at `require.HttpError`?

services/publicapi/service.go
```diff
@@ -145,6 +156,110 @@ func toTxOutput(transactionOutput *txResponse) *services.SendTransactionOutput {
 	return &services.SendTransactionOutput{ClientResponse: response.Build()}
 }
 
+func (s *service) CallMethod(input *services.CallMethodInput) (*services.CallMethodOutput, error) {
```
electricmonk: consider splitting separate methods to separate files

services/publicapi/test/get_transaction_test.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func TestGetTransactionStatus_GetPendingStatusFromTxPool(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		harness := newPublicApiHarness(ctx, 1*time.Second)
+
+		txb := builders.Transaction().Builder()
+		txHash := digest.CalcTxHash(txb.Build().Transaction())
+
+		harness.OnGetTransactionSetPending()
```
electricmonk: these methods don't need to be exported
also, maybe name the setup method according to the scenarios they describe:
```
harness.transactionIsInPool(txHash)
harness.transactionNotInPool(txHash)
```

bootstrap/httpserver/server_handlers_test.go
noambergIL: i looked ... it doesn't seem to have somehting that is binary post in the body.

services/publicapi/service.go
noambergIL: will consider when i do call method

PR #296, opened 2018-10-03
Add more log filters including composition filter Or

netoneko:


instrumentation/log/filters.go
```diff
@@ -97,5 +110,30 @@ func (f *excludeField) Allows(level string, message string, fields []*Field) boo
 	return true
 }
 
+type matchField struct {
```
electricmonk: We have `includeFieldWithKey`. Can you think of a nice way to generalize a type to handle both this case and the other case?

test/harness/acceptance_network.go
```diff
@@ -31,7 +31,7 @@ func NewAcceptanceTestNetwork(numNodes uint32, logFilters []log.Filter, consensu
 		WithOutput(log.NewOutput(output).
 		WithFormatter(log.NewHumanReadableFormatter())).
 		WithFilters(logFilters...).
-		WithFilters(log.OnlyErrors())
+		WithFilters(log.Or(log.OnlyErrors(), log.OnlyCheckpoints()))
```
electricmonk: I'm not sure this is a good enough approach. What if the `logFilters` param above has some rejecting filters? it will make the `Or` filter irrelevant. We can keep this for the mean time, but note that the entire filter concept might require some overhaul

PR #311, opened 2018-10-08
refactored the port in the gamma tests to be in 1 place per test and

itamararjuan:
routed it through the cli calls instead of using the default port


noambergIL:
itamar please don't choose 8090 or 8089 ... maybe try in tests for 9000 ... the e2e has 8090 as default now.

devtools/gamma-server/test/gamma_server_and_cli_test.go
```diff
@@ -148,15 +153,15 @@ func transferAmountToAddress(t *testing.T, keyPair *keys.Ed25519KeyPair, targetA
 	require.NotNil(t, response.TransactionReceipt.Txhash, "got empty txhash")
 }
 
-func getBalanceOfAddress(t *testing.T, targetAddress primitives.Ripmd160Sha256, expectedAmount uint64) {
+func getBalanceOfAddress(t *testing.T, targetAddress primitives.Ripmd160Sha256, expectedAmount uint64, port uint64) {
```
talkol: Why is the port `uint64` and not `uint16`?

devtools/gamma-server/test/gamma_server_and_cli_test.go
```diff
@@ -126,7 +129,11 @@ func generateAddCounterJSON(amount uint64) []byte {
 	return addJSONBytes
 }
 
-func transferAmountToAddress(t *testing.T, keyPair *keys.Ed25519KeyPair, targetAddress primitives.Ripmd160Sha256, amount uint64) {
+func getNodeUrl(port uint64) string {
+	return "http://localhost:" + strconv.FormatUint(port, 10)
```
talkol: Might be simpler to do `fmt.Sprintf("http://localhost:%d", port)` and then you don't have all the `uint64` issues

PR #314, opened 2018-10-10
Transaction batching

### LINKED ISSUE #119: Signing a batch of transactions
TransactionPool is supposed to sign a batch of transactions before forwarding them via gossip.
Our signature API currently takes `[]byte`.
This would require copying all transactions `.Raw()` into a new `[]byte` prior to signing, which defeats the purpose of membuffers.

What do we do?

netoneko:
Fixes #119


electricmonk:
this is work @netoneko and I did together, a third-party reviewer is required here

services/transactionpool/forward_transactions.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+type TransactionForwarderConfig interface {
+	NodePublicKey() primitives.Ed25519PublicKey
+	NodePrivateKey() primitives.Ed25519PrivateKey
+	TransactionPoolPropagationBatchSize() uint16
+	TransactionPoolPropagationBatchingTimeout() time.Duration
+}
+
+func (s *service) RegisterTransactionResultsHandler(handler handlers.TransactionResultsHandler) {
```
noambergIL: why did you move the registration of txresults to the forward tx file ? it doesn't really belong here

services/transactionpool/forward_transactions.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, tx := range input.Message.SignedTransactions {
+		txHash := digest.CalcTxHash(tx.Transaction())
+		s.logger.Info("adding forwarded transaction to the pool", log.String("flow", "checkpoint"), log.Stringable("transaction", tx), log.Stringable("txHash", txHash))
+		if _, err := s.pendingPool.add(tx, sender.SenderPublicKey()); err != nil {
+			s.logger.Error("error adding forwarded transaction to pending pool", log.Error(err), log.Stringable("transaction", tx), log.Stringable("txHash", txHash))
+		}
+	}
+	return nil, nil
+}
+
```
noambergIL: also about this .. the service does the handling even if it is forwarded ... i'd consider leaving it in the service or maybe joining the two funcs to a new file called tx_handlers.go ?

services/transactionpool/forward_transactions.go
electricmonk: the thinking was that this file deals with transaction propagation, but I'm not sure about this anymore

services/transactionpool/forward_transactions.go
netoneko: We decided to move everything related to transaction forwarding in the same file.

services/transactionpool/forward_transactions.go
noambergIL: is transaction result same as forwarding ?

netoneko:
Before

screen shot 2018-10-10 at 17 14 15

After

screen shot 2018-10-10 at 17 14 07

Looks like we have huge bump (36% **up**) in 95% and we yet have to understand why does it take longer to process the transactions. Of course it's related to propagation but maybe we need to play around with it to find optimal values for acceptance tests.

services/transactionpool/forward_transactions.go
electricmonk: what do you mean by that?

netoneko:
I am merging this because before @talkol finishes his work on contexts there is no way for us to properly debug why the hike has occurred.

PR #317, opened 2018-10-10
New block sync

jlevison:
new state machine, connected to the block storage
missing:
1. the source implementation (sync server) is still running with the old logic, need to remove that
2. can refactor the sync state factory to use a BlockSyncGossipClient (which is the correct abstraction)
3. some more invalid responses logic, such as:
- receiving an availability request with blocks older than what we have latest (need to dump it)
- receiving unexpected blocks in a chunk (need to dump the entire chunk)


services/blockstorage/sync/state_finished_car.go
```diff
@@ -0,0 +1,47 @@
...
+	logger    log.BasicLogger
+	sf        *stateFactory
+}
+
+func (s *finishedCARState) name() string {
+	return "finished-collecting-availability-requests-state"
+}
+
+func (s *finishedCARState) processState(ctx context.Context) syncState {
+	if ctx.Err() == context.Canceled {
```
electricmonk: eh?

services/blockstorage/sync/state_finished_car_test.go
```diff
@@ -0,0 +1,54 @@
...
+	h := newBlockSyncHarness()
+	finishedState := h.sf.CreateFinishedCARState([]*gossipmessages.BlockAvailabilityResponseMessage{response})
+	shouldBeWaitingState := finishedState.processState(h.ctx)
+
+	require.IsType(t, &waitingForChunksState{}, shouldBeWaitingState, "next state should be waiting for chunk")
+}
+
+func TestFinishedWithInvalidResponsesMovesToIdle(t *testing.T) {
+	h := newBlockSyncHarness()
+	finishedState := h.sf.CreateFinishedCARState(nil)
```
electricmonk: again, I think nil slices and empty slices are the same, am I wrong?

services/blockstorage/sync/state_idle.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil
+	}
+}
+
+func (s *idleState) blockCommitted() {
+	select {
+	case s.restartIdle <- struct{}{}:
+		s.logger.Info("sync got new block commit")
+	default:
+		s.logger.Info("channel was not ready, skipping notification")
```
electricmonk: please document why this default clause is required here

services/blockstorage/sync/state_idle_test.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestIdleStateStaysIdleOnCommit(t *testing.T) {
+	h := newBlockSyncHarness()
+	idle := h.sf.CreateIdleState()
+	var next syncState = nil
+	latch := make(chan struct{})
```
electricmonk: I wonder; this latch-goroutine-processState mechanism repeats in some tests - is there a common utility function that could be extracted or would that be premature?

services/blockstorage/sync/state_idle_test.go
```diff
@@ -0,0 +1,45 @@
...
+	<-latch
+	require.True(t, next != idle, "processState state should be a different idle state (which was restarted)")
+}
+
+func TestIdleStateMovesToCollectingOnNoCommitTimeout(t *testing.T) {
+	h := newBlockSyncHarness()
+	idle := h.sf.CreateIdleState()
+	next := idle.processState(h.ctx)
+	_, ok := next.(*collectingAvailabilityResponsesState)
+	require.True(t, ok, "processState state should be collecting availability responses")
```
electricmonk: `require.IsType`?

services/blockstorage/sync/state_idle_test.go
```diff
@@ -0,0 +1,45 @@
...
+	idle := h.sf.CreateIdleState()
+	var next syncState = nil
+	latch := make(chan struct{})
+	go func() {
+		next = idle.processState(h.ctx)
+		latch <- struct{}{}
+	}()
+	idle.blockCommitted()
+	<-latch
+	require.True(t, next != idle, "processState state should be a different idle state (which was restarted)")
```
electricmonk: you're actually not checking that the next state *is* idle; should add a `require.IsType` before.

services/blockstorage/sync/state_processing_blocks.go
```diff
@@ -0,0 +1,69 @@
...
+	return "processing-blocks-state"
+}
+
+func (s *processingBlocksState) processState(ctx context.Context) syncState {
+	if ctx.Err() == context.Canceled {
+		return nil
+	}
+
+	if s.blocks == nil {
+		s.logger.Error("fatal error in block sync, received no blocks to processing blocks state")
```
electricmonk: well, *fatal* might be a bit of an exaggeration here, no? I think this is one of those `throw new ThisShouldNeverHappenException()` scenarios

electricmonk: by which I mean, "unexpected error in block sync"

services/blockstorage/sync/state_processing_blocks.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"context"
+	"github.com/orbs-network/orbs-network-go/instrumentation/log"
+	"github.com/orbs-network/orbs-spec/types/go/protocol/gossipmessages"
+	"github.com/orbs-network/orbs-spec/types/go/services"
+)
+
+type processingBlocksState struct {
+	blocks  *gossipmessages.BlockSyncResponseMessage
```
electricmonk: it's like just a *bit* ugly that you hold the gossip message here, but is it better to duplicate the struct? I truly don't know

services/blockstorage/sync/state_processing_blocks.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			s.logger.Error("failed to validate block received via sync", log.Error(err))
+			break
+		}
+
+		_, err = s.storage.CommitBlock(&services.CommitBlockInput{BlockPair: blockPair})
+
+		if err != nil {
+			s.logger.Error("failed to commit block received via sync", log.Error(err))
+			break
```
electricmonk: *should* we break? think about the scenario: I'm at block 4, I just got a chunk with blocks 5 to 10, but just before I got to this current state, I managed to commit block 5 via the consensus route. Should I drop blocks 6-10?

electricmonk: conversely, if *any* block fails to validate I think it's fair to treat the entire chunk as byzantine and break

electricmonk: anyway this is an optimization so maybe open an issue and move on rather than delay merging this PR

services/blockstorage/sync/state_waiting_for_chunks.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (s *waitingForChunksState) gotBlocks(message *gossipmessages.BlockSyncResponseMessage) {
+	if !message.Sender.SenderPublicKey().Equal(s.sourceKey) {
+		s.logger.Info("byzantine message detected, expected source key does not match incoming",
+			log.Stringable("source-key", s.sourceKey),
+			log.Stringable("message-key", message.Sender.SenderPublicKey()))
+		s.abort <- struct{}{}
+	} else {
+		s.blocks = message
```
electricmonk: why not pass `blocks` in the channel?

services/blockstorage/sync/state_waiting_for_chunks_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	h.storage.When("LastCommittedBlockHeight").Return(primitives.BlockHeight(0)).Times(1)
+	h.gossip.When("SendBlockSyncRequest", mock.Any).Return(nil, errors.New("gossip failure")).Times(1)
+
+	waitingState := h.sf.CreateWaitingForChunksState(h.config.NodePublicKey())
+	nextState := waitingState.processState(h.ctx)
+
+	_, isIdle := nextState.(*idleState)
+
+	require.True(t, isIdle, "expecting back to idle on transport error")
```
electricmonk: `require.IsType`

services/blockstorage/sync/state_waiting_for_chunks_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	h.storage.When("LastCommittedBlockHeight").Return(primitives.BlockHeight(0)).Times(1)
+	h.gossip.When("SendBlockSyncRequest", mock.Any).Return(nil, nil).Times(1)
+
+	waitingState := h.sf.CreateWaitingForChunksState(h.config.NodePublicKey())
+	nextState := waitingState.processState(h.ctx)
+
+	_, isIdle := nextState.(*idleState)
+
+	require.True(t, isIdle, "expecting back to idle on transport error")
```
electricmonk: ditto

services/blockstorage/sync/state_waiting_for_chunks_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+func TestWaitingAcceptsNewBlockAndMovesToProcessing(t *testing.T) {
+	blocksMessage := builders.BlockSyncResponseInput().Build().Message
+	h := newBlockSyncHarness().withNodeKey(blocksMessage.Sender.SenderPublicKey()) //.withWaitForChunksTimeout(10 * time.Millisecond)
+
+	h.storage.When("LastCommittedBlockHeight").Return(primitives.BlockHeight(10)).Times(1)
+	h.gossip.When("SendBlockSyncRequest", mock.Any).Return(nil, nil).Times(1)
+
+	waitingState := h.sf.CreateWaitingForChunksState(h.config.NodePublicKey())
+	var nextState syncState
```
electricmonk: same comment from before about a shared mechanism; more than DRYing up the code, it will force you to give a name to this synchronization mechanism you built here

services/blockstorage/sync/state_waiting_for_chunks_test.go
```diff
@@ -0,0 +1,116 @@
...
+	go func() {
+		nextState = waitingState.processState(h.ctx)
+		latch <- struct{}{}
+	}()
+	waitingState.gotBlocks(blocksMessage)
+	<-latch
+
+	require.IsType(t, &processingBlocksState{}, nextState, "expecting to be at processing state after blocks arrived")
+	pbs := nextState.(*processingBlocksState)
+	require.NotNil(t, pbs.blocks, "blocks payload initialized in processing stage")
```
electricmonk: same comment from before re: nil slices

services/blockstorage/test/block_sync_test.go
```diff
@@ -196,8 +208,43 @@ func TestSyncCompletePetitionerSyncFlow(t *testing.T) {
...
 
+		// verify that we committed the blocks
 		harness.verifyMocks(t, 4)
 	})
 }
+
+func TestSyncNeverStartsWhenBlocksAreCommitted(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		// let the sync time to start
+		time.Sleep(1 * time.Millisecond)
```
electricmonk: `*ugh*`

services/blockstorage/test/block_sync_test.go
```diff
@@ -196,8 +208,43 @@ func TestSyncCompletePetitionerSyncFlow(t *testing.T) {
...
 
+		// fake the response
 		harness.blockStorage.HandleBlockSyncResponse(blockSyncResponse)
 
+		// verify that we committed the blocks
 		harness.verifyMocks(t, 4)
 	})
 }
+
+func TestSyncNeverStartsWhenBlocksAreCommitted(t *testing.T) {
```
electricmonk: so how do I know what the sync timeout is? maybe it didn't start because the timeout is 100ms and you only waited 10ms?

services/blockstorage/test/commit_block_test.go
```diff
@@ -14,6 +15,8 @@ import (
 func TestCommitBlockSavesToPersistentStorage(t *testing.T) {
 	test.WithContext(func(ctx context.Context) {
 		harness := newHarness(ctx)
+		// adding the broadcast as it might hit because of timeout, its not required for the test specifically
```
electricmonk: WHAAAAAT?
why? 
please explain

services/blockstorage/test/get_block_headers_test.go
```diff
@@ -16,6 +17,8 @@ func TestReturnTransactionBlockHeader(t *testing.T) {
 		harness := newHarness(ctx)
 		harness.expectCommitStateDiff()
 		harness.expectValidateWithConsensusAlgosTimes(1)
+		// adding the broadcast as it might hit because of timeout, its not required for the test specifically
```
electricmonk: same WHAAAAT as above

services/blockstorage/test/commit_block_test.go
jlevison: i think this is something we can discuss at the weekly:
originally the harness of block storage (which was written 3 months ago?) had a lot of bad assumptions, one of which was just having that global broadcast mock - while most tests do need it, until the sync process no-one really cared about it.
so the sync is part of the 'global' harness and then if a test is taking too long to trigger a sync process, it will begin calling the broadcast, so unless the test is specifically not expecting the sync to start (because of no commits), it probably needs this broadcast mock or it will flake out - this is true now of any test that is starting block storage (and not a mock, but a mock gossip)

services/blockstorage/test/block_sync_test.go
jlevison: config, right? i never really fixed the manual setting of config values here :/ lets discuss tomorrow

services/blockstorage/test/block_sync_test.go
jlevison: tell me about it..

services/blockstorage/sync/state_collecting_availability_requests_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	h.storage.When("LastCommittedBlockHeight").Return(primitives.BlockHeight(10)).Times(1)
+	h.gossip.When("BroadcastBlockAvailabilityRequest", mock.Any).Return(nil, errors.New("gossip failure")).Times(1)
+
+	collectingState := h.sf.CreateCollectingAvailabilityResponseState()
+	nextShouldBeIdle := collectingState.processState(h.ctx)
+
+	_, isIdle := nextShouldBeIdle.(*idleState)
+
+	require.True(t, isIdle, "should be idle on gossip error")
```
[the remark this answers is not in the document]
jlevison: yup

services/blockstorage/sync/state_idle_test.go
jlevison: correct, but this entire test is wrong, i added what was missing, but i think it should be rewritten as it assumes that the implementation will return a new state - maybe it should just commit some blocks in a rate faster than the timeout and then check that the state is still idle

services/blockstorage/sync/block_sync.go
```diff
@@ -0,0 +1,121 @@
...
+type blockSyncConfig interface {
+	NodePublicKey() primitives.Ed25519PublicKey
+	BlockSyncBatchSize() uint32
+	BlockSyncNoCommitInterval() time.Duration
+	BlockSyncCollectResponseTimeout() time.Duration
+	BlockSyncCollectChunksTimeout() time.Duration
+}
+
+type BlockSyncStorage interface {
+	GetBlocks(first primitives.BlockHeight, last primitives.BlockHeight) (blocks []*protocol.BlockPairContainer, firstAvailableBlockHeight primitives.BlockHeight, lastAvailableBlockHeight primitives.BlockHeight)
```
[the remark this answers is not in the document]
jlevison: yeah, just copied the entire thing, removing it

services/blockstorage/sync/block_sync.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Stringable("collect-responses-timeout", bs.config.BlockSyncCollectResponseTimeout()),
+		log.Stringable("collect-chunks-timeout", bs.config.BlockSyncCollectChunksTimeout()),
+		log.Uint32("batch-size", bs.config.BlockSyncBatchSize()))
+
+	go bs.syncLoop(ctx)
+	return bs
+}
+
+func (bs *BlockSync) syncLoop(ctx context.Context) {
+	meter := bs.logger.Meter("inter-sync-main-loop")
```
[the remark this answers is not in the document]
jlevison: counting if the main sync loop was restarted somehow - this reminded me i wanted to add a measure to each state, but wanted to ask if you want a naming convention, i'll add the timers, names can be easily changed later

services/blockstorage/sync/block_sync.go
```diff
@@ -0,0 +1,121 @@
...
+	for bs.currentState = bs.sf.CreateIdleState(); bs.currentState != nil; {
+		bs.logger.Info("state transitioning", log.String("current-state", bs.currentState.name()))
```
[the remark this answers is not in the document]
jlevison: true, just String() instead of that silly name()

services/blockstorage/sync/block_sync_test.go
```diff
@@ -0,0 +1,45 @@
...
+	h.storage.When("LastCommittedBlockHeight").Return(primitives.BlockHeight(10)).Times(1)
+	h.gossip.When("BroadcastBlockAvailabilityRequest", mock.Any).Return(nil, nil).Times(1)
+	sync := NewBlockSync(h.ctx, h.config, h.gossip, h.storage, h.logger)
+	h.cancel()
+	time.Sleep(time.Millisecond)
+	// TODO: refactor this once more logic is added, this is not really checking the the goroutine stopped
+	require.True(t, sync.terminated, "expecting the stop flag up")
+}
+
+func TestBlockSyncStaysInIdleOnBlockCommitExternalMessage(t *testing.T) {
```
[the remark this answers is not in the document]
jlevison: it covers HandleBlockCommitted()
there is another unit test inside idleState that covers the code there and then another one at a higher level covering the flow from gossip - felt that instead of relying on the gossip test that will find problem anywhere in the pipeline, to have all of these tests so its easier to pinpoint the problem

services/blockstorage/sync/state_collecting_availability_requests.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+func (s *collectingAvailabilityResponsesState) blockCommitted() {
+	return
+}
+
+func (s *collectingAvailabilityResponsesState) gotAvailabilityResponse(message *gossipmessages.BlockAvailabilityResponseMessage) {
+	s.logger.Info("got a new availability response", log.Stringable("response-source", message.Sender.SenderPublicKey()))
+	s.responses = append(s.responses, message)
```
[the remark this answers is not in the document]
jlevison: the mutex is handled at the entry point gossip starts - the question is valid, should we lock at the gossip entrypoint or one level down, which can be called internally (but never should, although that may be famous last words..)

services/blockstorage/sync/state_collecting_availability_requests_test.go
```diff
@@ -0,0 +1,91 @@
...
+		nextShouldBeFinished = collectingState.processState(h.ctx)
+		latch <- struct{}{}
+	}()
+	require.NoError(t, test.EventuallyVerify(10*time.Millisecond, h.gossip), "broadcast was not sent out")
+	collectingState.gotAvailabilityResponse(message)
+	<-latch
+
+	require.IsType(t, &finishedCARState{}, nextShouldBeFinished, "state should transition to finished CAR")
+	fcar := nextShouldBeFinished.(*finishedCARState)
+	require.NotNil(t, fcar.responses, "finished should have the responses initialized")
```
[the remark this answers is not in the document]
jlevison: yeah, not intuitive imo, but that is a redundant line

services/blockstorage/sync/state_finished_car_test.go
jlevison: yes, so can it ever be "invalid"? i think i should leave it as sanity

services/blockstorage/sync/state_processing_blocks.go
jlevison: correct, too much drama

services/blockstorage/sync/state_waiting_for_chunks_test.go
jlevison: not a slice (but maybe it will be if i refactor as you suggested, we'll discuss that tomorrow)

services/blockstorage/sync/state_finished_car.go
jlevison: checking if context was terminated (thats the way to do it without blocking on the channel)

services/blockstorage/sync/state_idle.go
jlevison: good idea, bled over that one

services/blockstorage/sync/block_sync.go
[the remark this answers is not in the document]
electricmonk: I want to adopt a dot-notation-based convention: system.subsystem.component.state.whatever

how, exactly, would it restart? also, you're measuring the time it took to complete, not the *times* it ran

services/blockstorage/sync/block_sync.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (bs *BlockSync) HandleBlockCommitted() {
+	bs.eventLock.Lock()
+	defer bs.eventLock.Unlock()
+	if bs.currentState != nil {
+		bs.currentState.blockCommitted()
+	}
+}
+
+func (bs *BlockSync) HandleBlockAvailabilityRequest(input *gossiptopics.BlockAvailabilityRequestInput) (*gossiptopics.EmptyOutput, error) {
```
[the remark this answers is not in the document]
electricmonk: so, again - the *service* implements the handler - since it must do that as per protos. It then delegates the *client* stuff to the state machine and the *server* stuff to another unit, but they don't have to implement the handler interface.

services/blockstorage/sync/block_sync_test.go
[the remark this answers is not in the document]
electricmonk: my point was - since this test duplicates other tests at lower levels of the system, you should document why this test is needed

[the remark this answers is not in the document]
electricmonk: my point was - since this test duplicates other tests at lower levels of the system, you should document why this test is needed

services/blockstorage/sync/state_finished_car_test.go
electricmonk: I think not. It's a property of golang that slices are never nil

services/blockstorage/test/block_sync_test.go
electricmonk: can you think of a way to make this deterministic?

services/blockstorage/test/commit_block_test.go
electricmonk: so maybe you don't start block sync in component tests of the block storage?

jlevison: putting aside the technical 'how-to' on doing that, i think the sync should be running in the background of the component tests, as it will in production - the alternative is to redesign the sync as its own service

services/blockstorage/test/block_sync_test.go
jlevison: if i had a way i would have done it by now. 
there is one thing in the block-sync design that 'troubled' me, is that when i shut it down, in order to know its actually shut down i am holding a boolean, it feels like something is missing in that area, like a letch saying 'good to go', thats my only idea around that.

services/blockstorage/sync/block_sync.go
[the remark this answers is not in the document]
jlevison: yeah, i agree, will be fixed, i will keep the code under the sync package still, but the service will implement the handlers and delegate accordingly (not sure as part of this pr)

services/blockstorage/sync/block_sync_test.go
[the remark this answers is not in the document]
jlevison: missed that point :) i'll document it

services/blockstorage/sync/state_collecting_availability_requests.go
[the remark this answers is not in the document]
jlevison: change to a channel

services/blockstorage/sync/state_collecting_availability_requests_test.go
```diff
@@ -0,0 +1,91 @@
...
+	h := newBlockSyncHarness()
+
+	h.storage.When("LastCommittedBlockHeight").Return(primitives.BlockHeight(10)).Times(1)
+	h.gossip.When("BroadcastBlockAvailabilityRequest", mock.Any).Return(nil, nil).Times(1)
+
+	message := builders.BlockAvailabilityResponseInput().Build().Message
+	collectingState := h.sf.CreateCollectingAvailabilityResponseState()
+	var nextShouldBeFinished syncState = nil
+	latch := make(chan struct{})
+	go func() {
```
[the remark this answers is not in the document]
jlevison: will change to a function in the harness that takes nextState+function (payload)

services/blockstorage/sync/state_collecting_availability_requests_test.go
```diff
@@ -0,0 +1,91 @@
...
+	fcar := nextShouldBeFinished.(*finishedCARState)
+	require.NotNil(t, fcar.responses, "finished should have the responses initialized")
+	require.Equal(t, 1, len(fcar.responses), "there should be one response")
+	require.Equal(t, message.Sender, fcar.responses[0].Sender, "state sender should match message sender")
+	require.Equal(t, message.SignedBatchRange, fcar.responses[0].SignedBatchRange, "state payload should match message")
+
+	h.verifyMocks(t)
+}
+
+func TestCollectingAvailabilityResponsesAddsAResponse(t *testing.T) {
```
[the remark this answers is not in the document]
jlevison: removed

services/blockstorage/sync/state_processing_blocks.go
jlevison: will open an issue, commitblock failure does not have to be fatal

services/blockstorage/sync/state_waiting_for_chunks.go
jlevison: will move to channel

services/blockstorage/test/commit_block_test.go
jlevison: change the harness to configure sync to a very long timeout

services/blockstorage/sync/state_processing_blocks.go
jlevison: opened issue #320

PR #319, opened 2018-10-11
state storage implementation notes

ronnno:


noambergIL:
General question: how many contract names are in one virtual chain ? because hashing it doubles key size (and larger proofs) and in some cases where there are one or two it seems a waste ...

ronnno:
imho the original reasoning was that since we use a compact trie the length of a proof will not be affected. if you have just a few contracts there will typically not be more than one or two nodes to represent the entire contract name. 

There is a good point in the fact that it requires twice the space to store each key, and this will affect state storage in more than just the merkle trie. The benefit of isolating contract names at the beginning of the variable identifier is that we can isolate subtrees in the merkle trie corresponding to specific contract names and possibly later on isolate them physically. this may be useful for backup etc. but this is arguably YANGI

I'm for opening this up for discussion. @OdedWx what do you think?

PR #322, opened 2018-10-13
New metric system

electricmonk:
Added a new metric system (side by side with existing one)

Left to do:
- move all services to the new one (I moved some)
- remove old system
- come up with additional KPIs
- rotate bucketing meters (histogram / rate)
- figure out exporting mechanism (logger / udp / homing pigeon...)
- more?

@netoneko - this doesn't build currently due to a problem I'm having adding a package via Manul:
```
$ manul -I golang.org/x/time
2018-10-13 19:26:51 [FATAL]: unknown dependency: golang.org/x/time
```
Can you please take a look and tell me what I'm doing wrong?


netoneko:
@electricmonk you should use `manul -It golang.org/x/time` because it's a test dependency. `manul -Irt` would find any uninstalled dependencies and install them automatically.

instrumentation/log/filters.go
```diff
@@ -39,6 +39,10 @@ func IgnoreErrorsMatching(pattern string) Filter {
 	return &errorRegexp{pattern: pattern}
 }
 
+func Nothing() Filter {
```
netoneko: I would call it `DiscardAll` because if you _filter nothing_, you're not supposed to filter at all.

instrumentation/metric/registry.go
```diff
@@ -0,0 +1,98 @@
...
+
+	var s string
+	for _, m := range r.mu.metrics {
+		s += m.String()
+	}
+
+	return s
+}
+
+func (r *inMemoryRegistry) ExportAll() interface{} {
```
netoneko: Why is it an `interface{}` and not `map[string]interface{}`?

services/consensuscontext/service.go
```diff
@@ -31,6 +41,10 @@ func NewConsensusContext(
 		stateStorage:    stateStorage,
 		config:          config,
 		logger:          logger.WithTags(LogTag),
+		metrics: metrics{
```
netoneko: It feels a bit wrong to initialize it here, maybe we need a factory method per service like `consensuscontext.NewMetrics()`?

bootstrap/httpserver/server.go
```diff
@@ -100,6 +105,12 @@ func report(reporting log.BasicLogger, h http.Handler) http.Handler {
 	})
 }
 
+func (s *server) dumpMetrics(w http.ResponseWriter, r *http.Request) {
```
netoneko: Couldn't find any tests for this.

netoneko:
@electricmonk and I have agreed that `RecordNanosSince` and `RecordMillisSince` are going to be replaced with `RecordSince` for the sake of uniformity.

instrumentation/metric/registry.go
electricmonk: why do we care? either it's strictly typed - if we want to convert to some other system's domain model - or it's going to become JSONs anyway, in which case we could care less. No?

services/consensuscontext/service.go
electricmonk: why?

bootstrap/httpserver/server.go
electricmonk: no tests :)
I don't see any value in testing this piece of code at this point

PR #324, opened 2018-10-14
Export metrics in a scheduled manner

netoneko:


bootstrap/node_logic.go
```diff
@@ -63,6 +63,8 @@ func NewNodeLogic(
 	//consensusAlgos = append(consensusAlgos, leanhelix.NewLeanHelixConsensusAlgo(gossipService, blockStorageService, transactionPoolService, consensusContextService, logger, nodeConfig))
 	consensusAlgos = append(consensusAlgos, benchmarkconsensus.NewBenchmarkConsensusAlgo(ctx, gossipService, blockStorageService, consensusContextService, logger, nodeConfig))
 
+	metricFactory.ReportEvery(ctx, nodeConfig.MetricsReportInterval(), logger)
```
electricmonk: 1) rename variable `metricRegistry`
2) do we really need the report metric interval as a config value?

instrumentation/metric/gauge.go
```diff
@@ -48,6 +51,14 @@ func (g *Gauge) Value() int64 {
 	return g.value
 }
 
+func (g *Gauge) Reset() {
```
electricmonk: I think it doesn't make sense to reset `Gauge`s at all. They are counters of present state, for instance we have a gauge that counts num of transactions in pool

instrumentation/metric/histogram.go
```diff
@@ -69,3 +72,21 @@ func (h *Histogram) Export() interface{} {
 		h.histo.TotalCount(),
 	}
 }
+
+func (h *Histogram) Reset() {
+	h.histo.Reset()
```
electricmonk: check whether this makes sense - the library has a `WindowedHistogram` for this purpose

instrumentation/metric/rate.go
```diff
@@ -53,3 +56,19 @@ func (r *Rate) Measure(eventCount int64) {
 
 	r.runningSum += eventCount
 }
+
+func (r *Rate) Reset() {
```
electricmonk: we need to understand the relationship between the tick interval (in ` func Measure`) and the polling interval. Does it really make sense to reset the rate? I'm not sure anymore - maybe look at how other people use the `MovingAverage` library.

also, if we DO reset, do we want to reset `runningSum` as well?

instrumentation/metric/registry.go
```diff
@@ -15,13 +18,20 @@ type Factory interface {
 type Registry interface {
 	Factory
 	String() string
-	ExportAll() map[string]interface{}
+	ExportAll() map[string]exportedMetric
```
electricmonk: do we need `ExportAll` if we have `ExportAndRotate`?

instrumentation/metric/registry.go
```diff
@@ -79,14 +89,46 @@ func (r *inMemoryRegistry) String() string {
...
 		all[m.Name()] = m.Export()
 	}
 
 	return all
 }
+
+func (r *inMemoryRegistry) ExportAndRotate() map[string]exportedMetric {
+	results := r.ExportAll()
+
+	for _, m := range r.mu.metrics {
```
electricmonk: what happens if we export all and then before we reset all, some more metrics get updated? do we merely lose some precision, or does it result in metric corruption?
maybe export and rotate should be an atomic operation per metric?

instrumentation/metric/registry.go
```diff
@@ -79,14 +89,46 @@ func (r *inMemoryRegistry) String() string {
...
+func (r *inMemoryRegistry) ReportEvery(ctx context.Context, interval time.Duration, logger log.BasicLogger) {
+	go func() {
+		periodicalTrigger := synchronization.NewPeriodicalTrigger(interval, func() {
+			r.report(logger)
+		})
+
+		periodicalTrigger.Start()
+
+		select {
+		case <-ctx.Done():
```
electricmonk: this is ugly - either add the context to `NewPeriodicalTrigger` - which would make sense anyway, or do both polling and cancellation inline in the `select` statement.

I think after @jlevison's refactor of BlockSync, the new trigger might be irrelevant - if so, don't use it

instrumentation/metric/registry_test.go
```diff
@@ -0,0 +1,27 @@
...
+func TestInMemoryRegistry_ExportAndRotate(t *testing.T) {
+	registry := NewRegistry()
+	gauge := registry.NewGauge("hello")
+	gauge.Add(1)
+
+	gaugeValue := registry.ExportAndRotate()["hello"].(gaugeExport)
+	require.EqualValues(t, gaugeValue.Value, 1)
+
+	gaugeValue = registry.ExportAll()["hello"].(gaugeExport)
+	require.EqualValues(t, gaugeValue.Value, 0)
```
electricmonk: you could also read the value strait from the `gauge`, no? do you really need to call `ExportAll`?

electricmonk: also maybe add a test per metric for the `Reset` operation (if it stays separate from `Rotate`) or for `ExportAndMaybeRotate` if we end up having an atomic operation

bootstrap/node_logic.go
netoneko: 2. I don't see any reason why not. Everything is configurable, and this should be too.

instrumentation/metric/registry.go
netoneko: `ExportAll` is for http responses, `ExportAndRotate` is for reporting to logs. We don't want to rotate on every http request.

instrumentation/metric/registry.go
netoneko: @jlevison promised to refactor `NewPeriodicalTrigger` to be context-aware, so I would prefer to stick with the trigger until it happens.

instrumentation/metric/registry.go
electricmonk: so maybe rename it to reflect that it is a "peek" operation

instrumentation/metric/registry.go
electricmonk: ETA? if he's not doing it this week, just do it yourself :)

netoneko: This week.

instrumentation/metric/rate.go
netoneko: I decided not to reset at all.

instrumentation/metric/registry.go
netoneko: I have removed `ExportAndRotate` all together.

instrumentation/metric/histogram.go
```diff
@@ -28,55 +28,55 @@ type histogramExport struct {
 func newHistogram(name string, max int64) *Histogram {
 	return &Histogram{
 		namedMetric: namedMetric{name: name},
-		histo:       hdrhistogram.New(0, max, 1),
+		histo:       hdrhistogram.NewWindowed(5, max, 1, 3),
```
electricmonk: are you rotating it anywhere?

netoneko: My understanding is that it is rotated automatically because it's rolling, so it uses 5 underlying histograms that constantly rotate by themselves.

PR #327, opened 2018-10-15
new block sync server

jlevison:
* removed the last of the old state machine
* added BlockSyncGossipClient
* state factor does not send the entire config object to states, just relevant config values
* small fix for one of the tests that was flaky (if it will still be flaky in the future will rethink it then)


services/blockstorage/block_sync_server.go
```diff
@@ -0,0 +1,86 @@
...
+func (s *service) sourceHandleBlockAvailabilityRequest(message *gossipmessages.BlockAvailabilityRequestMessage) error {
+	s.logger.Info("received block availability request", log.Stringable("sender", message.Sender))
+
+	lastCommittedBlockHeight := s.LastCommittedBlockHeight()
+
+	if lastCommittedBlockHeight <= message.SignedBatchRange.LastCommittedBlockHeight() {
+		return nil
+	}
+
+	firstAvailableBlockHeight := primitives.BlockHeight(1)
```
electricmonk: will this always be true? if not, can this test somehow fail when the assumption changes?

services/blockstorage/block_sync_server.go
```diff
@@ -0,0 +1,86 @@
...
+	blockType := message.SignedBatchRange.BlockType()
```
electricmonk: I missed this before - does this mean we're asking for results blocks and transaction blocks separately?

services/blockstorage/block_sync_server.go
```diff
@@ -0,0 +1,86 @@
...
+			}).Build(),
+		},
+	}
+	_, err := s.gossip.SendBlockAvailabilityResponse(response)
+	return err
+}
+
+func (s *service) sourceHandleBlockSyncRequest(message *gossipmessages.BlockSyncRequestMessage) error {
+	senderPublicKey := message.Sender.SenderPublicKey()
+	blockType := message.SignedChunkRange.BlockType()
```
electricmonk: so now I see that we're reading block type but not doing anything with it?

services/blockstorage/sync/block_sync_gossip_client.go
```diff
@@ -0,0 +1,84 @@
...
+	batchSize uint32
+	nodeKey   primitives.Ed25519PublicKey
+}
+
+func newBlockSyncGossipClient(
+	g gossiptopics.BlockSync,
+	s BlockSyncStorage,
+	l log.BasicLogger,
+	batchSize uint32,
+	pk primitives.Ed25519PublicKey) *blockSyncGossipClient {
```
electricmonk: what you can do is instead of sending the node pk, you can send a `Func() primitives.Ed25519PublicKey`, and reference the function from the config; it's more future-proof

services/blockstorage/sync/factory.go
```diff
@@ -53,14 +52,13 @@ func (f *stateFactory) CreateFinishedCARState(responses []*gossipmessages.BlockA
...
-		gossip:    f.gossip,
-		storage:   f.storage,
-		config:    f.config,
-		logger:    f.logger,
-		blocksC:   make(chan *gossipmessages.BlockSyncResponseMessage),
-		abort:     make(chan struct{}),
+		sourceKey:      sourceKey,
+		sf:             f,
+		gossipClient:   newBlockSyncGossipClient(f.gossip, f.storage, f.logger, f.config.BlockSyncBatchSize(), f.config.NodePublicKey()),
+		collectTimeout: f.config.BlockSyncCollectChunksTimeout(),
```
electricmonk: here too, just send the function reference rather than invoke the function

services/blockstorage/sync/state_collecting_availability_requests.go
```diff
@@ -4,18 +4,16 @@ import (
...
 type collectingAvailabilityResponsesState struct {
-	sf         *stateFactory
-	gossip     gossiptopics.BlockSync
-	storage    BlockSyncStorage
-	config     blockSyncConfig
-	logger     log.BasicLogger
-	responsesC chan *gossipmessages.BlockAvailabilityResponseMessage
+	sf             *stateFactory
+	gossipClient   *blockSyncGossipClient
+	collectTimeout time.Duration
```
electricmonk: and here too

services/blockstorage/sync/state_idle.go
```diff
@@ -5,10 +5,11 @@ import (
 	"github.com/orbs-network/orbs-network-go/instrumentation/log"
 	"github.com/orbs-network/orbs-network-go/synchronization"
 	"github.com/orbs-network/orbs-spec/types/go/protocol/gossipmessages"
+	"time"
 )
 
 type idleState struct {
-	config      blockSyncConfig
+	idleTimeout time.Duration
```
electricmonk: and here

services/blockstorage/sync/state_idle_test.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func TestIdleStateStaysIdleOnCommit(t *testing.T) {
-	h := newBlockSyncHarness()
+	h := newBlockSyncHarness().withNoCommitTimeout(time.Second) // we are checking for a newly created state, the timeout here is irrelevant
```
electricmonk: why does `withNoCommitTimeout` take a `Duration`?

services/blockstorage/test/block_sync_server_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/orbs-network/orbs-spec/types/go/services/gossiptopics"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestSourceRespondToAvailabilityRequests(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		harness := newHarness(ctx)
+		harness.setupSomeBlocks(3)
+		harness.gossip.When("SendBlockAvailabilityResponse", mock.Any).Return(nil, nil).Times(1)
```
electricmonk: are you intentionally not making any assertion on the expected response?

services/blockstorage/test/block_sync_server_test.go
```diff
@@ -0,0 +1,129 @@
...
+			Build()
+		_, err := harness.blockStorage.HandleBlockAvailabilityRequest(msg)
+
+		require.NoError(t, err, "expecting a happy flow")
+
+		harness.verifyMocks(t, 1)
+	})
+}
+
+func TestSourceRespondsNothingToAvailabilityRequestIfSourceIsBehindPetitioner(t *testing.T) {
```
electricmonk: is the name appropriate? looks like the behavior you expect is `DoesNotRespondToAvailabilityRequest`

services/blockstorage/test/block_sync_server_test.go
```diff
@@ -0,0 +1,129 @@
...
+			WithSenderPublicKey(keys.Ed25519KeyPairForTests(1).PublicKey()).
+			WithFirstBlockHeight(firstHeight).
+			Build()
+
+		chunksResponseVerifier := func(i interface{}) bool {
+			response, ok := i.(*gossiptopics.BlockSyncResponseInput)
+			if !ok {
+				require.Failf(t, "response type does not match", "", i)
+			}
+			require.EqualValues(t, batchSize, len(response.Message.BlockPairs), "batch size does not match config")
```
electricmonk: `require.Len`?

services/blockstorage/test/block_sync_server_test.go
```diff
@@ -0,0 +1,129 @@
...
+			return true
+		}
+
+		harness.gossip.When("SendBlockSyncResponse", mock.AnyIf("reponse should hold correct blocks", chunksResponseVerifier)).Return(nil, nil).Times(1)
+		harness.blockStorage.HandleBlockSyncRequest(msg)
+		harness.verifyMocks(t, 1)
+	})
+}
+
+func TestSourceIgnoresBlockSyncRequestIfSourceIsBehindOrInSync(t *testing.T) {
```
electricmonk: `OrInSync`? what does that mean?

services/blockstorage/test/block_sync_test.go
```diff
@@ -211,8 +211,11 @@ func TestSyncCompletePetitionerSyncFlow(t *testing.T) {
 }
 
 func TestSyncNeverStartsWhenBlocksAreCommitted(t *testing.T) {
+	// this test may still be flaky, it runs commits in a busy wait loop that should take longer than the timeout,
```
electricmonk: how can you make it completely deterministic?

services/blockstorage/test/block_sync_server_test.go
netoneko: It means that the source has block height less or equal to block height of the petitioner.

services/blockstorage/test/block_sync_test.go
jlevison: raise the timeouts - if we go into 100 and 50ms  scenario the probability of it being flaky is low enough (we see 5-15ms of freezes sometimes, so 50ms should be safe)
but then its 'slow', not sure how to refactor it

services/blockstorage/sync/block_sync_gossip_client.go
jlevison: makes sense, will fix everywhere

services/blockstorage/sync/state_idle_test.go
jlevison: it is the timeout interval for when the sync is supposed to transition to the next state
default for tests is 3ms i think
i added this to solve a flaky case where the next state function will deprioritize and then it would take more than 3ms to complete, and the state would change

services/blockstorage/test/block_sync_server_test.go
jlevison: correct

services/blockstorage/test/block_sync_server_test.go
jlevison: need to get myself used to look at require.X

services/blockstorage/test/block_sync_server_test.go
jlevison: so because source is behind it *may* be in sync, might be TMI because we do not really have a way to check if the source is syncing right now, and we do not care either, right? we do not 'deny syncing' as a potential source because we are syncing right now on newer blocks

services/blockstorage/test/block_sync_test.go
electricmonk: nope, this is just reducing probability of failure
let's do this - merge it, but open an issue and we can maybe hold a discussion tomorrow at Weekly about it

services/blockstorage/test/block_sync_server_test.go
electricmonk: so can you suggest a better name for this test?

services/blockstorage/test/block_sync_server_test.go
jlevison: no, it was missing in the test, added it

services/blockstorage/test/block_sync_server_test.go
jlevison: yeah, just without the OrInSync
already renamed, didn't push yet

services/blockstorage/test/block_sync_test.go
jlevison: sounds good, will do

services/blockstorage/block_sync_server.go
jlevison: I will discuss this with Tal, its his logic and i agree with you that there might be problems here - in either way i will need to make a small change here to make it less error prone. this is relevant to the other two comments about the sync server

PR #335, opened 2018-10-18
Block sync acceptance test

netoneko:


electricmonk:
@jlevison @netoneko please consider adding a stress test that fucks up with block sync messages and makes sure the system can handle it

PR #343, opened 2018-10-21
Maintain a single state snapshot with state deltas increments

ronnno:


services/statestorage/adapter/memory_persistence.go
```diff
@@ -3,132 +3,109 @@ package adapter
...
+
+	// TODO - this is our hard coded Genesis block. Move this to a more dignified place or load from a file
+
+	_, root := merkle.NewForest()
 	return &InMemoryStatePersistence{
-		// TODO remove this hard coded init of genesis block state once init flow syncs state storage with block storage
-		snapshots: map[primitives.BlockHeight]StateVersion{primitives.BlockHeight(0): map[primitives.ContractName]ContractState{}},
-		roots:     map[primitives.BlockHeight]primitives.MerkleSha256{},
+		mu:          sync.RWMutex{},
+		snapshot:    ChainDiff{},
```
noambergIL: the name snapshot is misleading ... this is the db itself

services/statestorage/service.go
```diff
@@ -16,31 +17,49 @@ import (
...
 type service struct {
 	config       config.StateStorageConfig
-	merkle       *merkle.Forest
 	blockTracker *synchronization.BlockTracker
 	logger       log.BasicLogger
 
-	mutex                    *sync.RWMutex
-	persistence              adapter.StatePersistence
-	lastCommittedBlockHeader *protocol.ResultsBlockHeader
+	mutex       sync.RWMutex
```
noambergIL: i would move the state incement, the incache and the height and ts to a seperate file with all the _methods so that it will be in charge of dealing with the history + persistent ....

question as merkle needs to be synced ... do we also have that there ?

services/statestorage/service.go
```diff
@@ -121,13 +176,38 @@ func (s *service) ReadKeys(ctx context.Context, input *services.ReadKeysInput) (
...
+}
+
 func (s *service) GetStateStorageBlockHeight(ctx context.Context, input *services.GetStateStorageBlockHeightInput) (*services.GetStateStorageBlockHeightOutput, error) {
 	s.mutex.RLock()
 	defer s.mutex.RUnlock()
 
 	result := &services.GetStateStorageBlockHeightOutput{
-		LastCommittedBlockHeight:    s.lastCommittedBlockHeader.BlockHeight(),
-		LastCommittedBlockTimestamp: s.lastCommittedBlockHeader.Timestamp(),
+		LastCommittedBlockHeight:    s.height,
```
noambergIL: i prefer keeing lastCommited header instead of height and ts seperately
maybe even keep one copy of each lastCommitted for each history  diff.

services/statestorage/adapter/memory_persistence.go
```diff
@@ -3,132 +3,109 @@ package adapter
...
+	blockHeight primitives.BlockHeight
+	timestamp   primitives.TimestampNano
+	merkleRoot  primitives.MerkleSha256
 }
 
 func NewInMemoryStatePersistence() *InMemoryStatePersistence {
+
+	// TODO - this is our hard coded Genesis block. Move this to a more dignified place or load from a file
+
+	_, root := merkle.NewForest()
```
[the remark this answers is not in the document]
ronnno: I agree this does not belong here. the merkle root should come from our genesis block representation. as the TODO states, this should move from here to a proper place. I have left the remark to make sure it's dealt with later on

services/statestorage/adapter/memory_persistence.go
```diff
@@ -3,132 +3,109 @@ package adapter
...
 )
 
-type ContractState map[string]*protocol.StateRecord
-type StateVersion map[primitives.ContractName]ContractState
-
 type InMemoryStatePersistence struct {
-	snapshots map[primitives.BlockHeight]StateVersion
-	roots     map[primitives.BlockHeight]primitives.MerkleSha256
+	mutex       sync.RWMutex
+	bState      ChainState
```
electricmonk: why the `b` prefix? please use something more meaningful

services/statestorage/layered_state.go
```diff
@@ -0,0 +1,107 @@
...
+func (ls *layeredState) Write(height primitives.BlockHeight, ts primitives.TimestampNano, root primitives.MerkleSha256, diff adapter.ChainState) error {
+	ls.incCache = append(ls.incCache, &stateIncrement{
+		diff:       diff,
+		merkleRoot: root,
+		height:     height,
+		ts:         ts,
+	})
+	ls.recentHeight = height
+	// TODO - move this loop for merging and persisting snapshots to a separate goroutine. merely append here with a safety array size limit
+	for len(ls.incCache) > ls.maxLayers {
```
electricmonk: either way this should probably move to an unexported method named `flush`, no?

services/statestorage/layered_state.go
```diff
@@ -0,0 +1,107 @@
...
+		pHeight:      pHeight,
+		pTs:          pTs,
+		pRoot:        pRoot,
+	}
+
+	return result
+}
+
+func (ls *layeredState) Write(height primitives.BlockHeight, ts primitives.TimestampNano, root primitives.MerkleSha256, diff adapter.ChainState) error {
+	ls.incCache = append(ls.incCache, &stateIncrement{
```
electricmonk: is this thread safe?

services/statestorage/layered_state.go
```diff
@@ -0,0 +1,107 @@
...
+	height     primitives.BlockHeight
+	ts         primitives.TimestampNano
+}
+
+type layeredState struct {
+	persistence  adapter.StatePersistence
+	maxLayers    int
+	incCache     []*stateIncrement
+	recentHeight primitives.BlockHeight
+	pHeight      primitives.BlockHeight
```
electricmonk: what does `p` stand for? please expand

services/statestorage/layered_state_test.go
```diff
@@ -0,0 +1,203 @@
...
+	v, exists, err := d.read(0, "c", "k1")
+	require.NoError(t, err)
+	require.EqualValues(t, false, exists)
+
+	v, exists, err = d.read(1, "c", "k1")
+	require.NoError(t, err)
+	require.EqualValues(t, true, exists)
+	require.EqualValues(t, "v1", v)
+
+	v, exists, err = d.read(200, "c", "k1")
```
electricmonk: replace `v` with `_`

services/statestorage/layered_state_test.go
```diff
@@ -0,0 +1,203 @@
...
+func TestWriteAtHeight(t *testing.T) {
+	persistenceMock := &StatePersistenceMock{}
+	persistenceMock.
+		When("Read", primitives.ContractName("c"), "k1").
+		Return((*protocol.StateRecord)(nil), false, nil).
+		Times(1)
+	d := newDriver(persistenceMock, 5)
+	d.write(1, "c", "k1", "v1")
+
+	v, exists, err := d.read(0, "c", "k1")
```
electricmonk: your naming makes it hard to understand what the test does. Try this here:
`_, exists, err := d.read(0, "c", "k1")`

services/statestorage/layered_state_test.go
```diff
@@ -0,0 +1,203 @@
...
+	require.NoError(t, err)
+	require.EqualValues(t, false, exists)
+
+	v, exists, err = d.read(1, "c", "k1")
```
electricmonk: and here:
`valueAtHeight1, exists, err = d.read(1, "c", "k1")`

services/statestorage/layered_state_test.go
```diff
@@ -0,0 +1,203 @@
...
+	require.NoError(t, err)
+	require.EqualValues(t, true, exists)
+	require.EqualValues(t, "v1", v)
+
+	v, exists, err = d.read(200, "c", "k1")
+	require.Error(t, err)
+
+	ok, errCalled := persistenceMock.Verify()
```
electricmonk: no need to assert both boolean and error result of `Verify`, just assert no error.

services/statestorage/service.go
```diff
@@ -18,29 +18,38 @@ var LogTag = log.Service("state-storage")
...
 		blockTracker: synchronization.NewBlockTracker(0, uint16(config.BlockTrackerGraceDistance()), config.BlockTrackerGraceTimeout()),
 		logger:       logger.WithTags(LogTag),
 
-		mutex:                    &sync.RWMutex{},
-		persistence:              persistence,
-		lastCommittedBlockHeader: (&protocol.ResultsBlockHeaderBuilder{}).Build(), // TODO change when system inits genesis block and saves it will need to be read from db
+		mutex:        sync.RWMutex{},
+		merkle:       merkle,
+		layeredState: newLayeredState(persistence, int(config.StateStorageHistoryRetentionDistance())),
+		height:       height,
```
electricmonk: but shouldn't `height` be a property of `layeredState`?

services/statestorage/service.go
```diff
@@ -137,15 +144,39 @@ func (s *service) GetStateHash(ctx context.Context, input *services.GetStateHash
...
 	if err != nil {
-		return nil, errors.Wrapf(err, "could not merkle root for block height %d", input.BlockHeight)
+		return nil, errors.Wrapf(err, "could not find a merkle root for block height %d", input.BlockHeight)
 	}
 	output := &services.GetStateHashOutput{StateRootHash: value}
 
 	return output, nil
 }
 
+func inflateChainState(csd []*protocol.ContractStateDiff) adapter.ChainState {
```
electricmonk: why do you need this method? you have what seems to be a very similar method inside the `layeredState`

services/statestorage/layered_state.go
ronnno: this object is not thread safe. everything here is called under a mutex from the call site

services/statestorage/adapter/memory_persistence.go
ronnno: replaced with prst, as in "persisted"

services/statestorage/layered_state.go
ronnno: done

services/statestorage/layered_state.go
ronnno: p measn persisted. replaced with prst

services/statestorage/layered_state_test.go
ronnno: ignoring ok everywhere. but if it's never important why is it included.. ?

services/statestorage/service.go
ronnno: there was at one point but i dont think its there anymore..

services/statestorage/service.go
ronnno: moved this from service to layeredState (now renamed to rollingRevisions)

PR #351, opened 2018-10-23
Add blockHeight metric to block storage

netoneko:
Also fixes a bug that erased metric name from periodic log.


services/blockstorage/service.go
```diff
@@ -112,6 +126,7 @@ func (s *service) updateLastCommittedBlock(block *protocol.BlockPairContainer) {
 	s.lastBlockLock.Lock()
 	defer s.lastBlockLock.Unlock()
 
+	s.metrics.blockHeight.Inc()
```
electricmonk: I'd do something different, since as soon as we're persisting block heights, this Gauge will lose sync with the block storage. Instead, add an `Update` method on the `Gauge` type and read the block height from `block`

netoneko: Done.

PR #353, opened 2018-10-23
benchmark consensus flakiness - mutex granularity

talkol:


electricmonk:
A neat trick I learned recently is that you can do this to better signify which fields are under mutex:
```
	mu struct {
		sync.RWMutex
		lastCommittedBlockUnderMutex                    *protocol.BlockPairContainer
		lastSuccessfullyVotedBlock                      primitives.BlockHeight // leader only
		lastCommittedBlockVotersUnderMutex              map[string]bool        // leader only
		lastCommittedBlockVotersReachedQuorumUnderMutex bool                   // leader only
	}
```

it saves on the need to suffix everything with `underMutex` and it wraps everything nicely

the only downside is that the call to `make` (or any initialization logic) needs to happen outside the (parent) struct init, as I couldn't find a way to initialize `mu` without repeating its definition:

```
	s.mu.lastSuccessfullyVotedBlock = blockHeightNone
	s.mu.lastCommittedBlockVotersUnderMutex = make(map[string]bool)
```

jlevison:
dont think its a big downside, i like that suggestion for organizing 'locked' members

PR #354, opened 2018-10-24
Fixes #285 (Flakiness in TestBenchmarkConsensusLeaderGetsVotesBeforeN…

### LINKED ISSUE #285: Flakiness in TestBenchmarkConsensusLeaderGetsVotesBeforeNextBlock
https://circleci.com/gh/orbs-network/orbs-network-go/1471

netoneko:
…extBlock)

### DIFF
--- a/test/acceptance/consensus_test.go
+++ b/test/acceptance/consensus_test.go
@@ -40,7 +40,7 @@ func TestLeanHelixLeaderGetsValidationsBeforeCommit(t *testing.T) {
 }
 
 func TestBenchmarkConsensusLeaderGetsVotesBeforeNextBlock(t *testing.T) {
-	harness.Network(t).Start(func(ctx context.Context, network harness.InProcessNetwork) {
+	harness.Network(t).WithMaxTxPerBlock(1).Start(func(ctx context.Context, network harness.InProcessNetwork) {
 
 		network.DeployBenchmarkToken(ctx, 5)
 
--- a/config/factory.go
+++ b/config/factory.go
@@ -38,6 +38,7 @@ func ForAcceptanceTests(
 	nodePrivateKey primitives.Ed25519PrivateKey,
 	constantConsensusLeader primitives.Ed25519PublicKey,
 	activeConsensusAlgo consensus.ConsensusAlgoType,
+	maxTxPerBlock uint32,
 ) mutableNodeConfig {
 	cfg := defaultConfig()
 	cfg.OverrideNodeSpecificValues(federationNodes,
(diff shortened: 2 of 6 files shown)


test/acceptance/consensus_test.go
```diff
@@ -40,7 +40,7 @@ func TestLeanHelixLeaderGetsValidationsBeforeCommit(t *testing.T) {
 }
 
 func TestBenchmarkConsensusLeaderGetsVotesBeforeNextBlock(t *testing.T) {
-	harness.Network(t).Start(func(ctx context.Context, network harness.InProcessNetwork) {
+	harness.Network(t).WithMaxTxPerBlock(1).Start(func(ctx context.Context, network harness.InProcessNetwork) {
```
electricmonk: not sure I agree with this fix
I think it's better to rethink the test than to assume anything about num of txs in block

electricmonk:
Please don't merge this yet, I'm pretty sure this is wrong
If we can avoid this test relying on 1 tx per block, it's better

netoneko:
It was @talkol's decision and he was very much against rethinking the test.

electricmonk:
it actually makes even less sense as the system *will* eventually close a block even with a single transaction, so I doubt that raising the max tx per block config has affected this test - please correct me if I'm wrong @talkol

electricmonk:
again, a block *will* be closed with a single transaction even with a  high max tx per block value, as long as the timeout for waiting for extra transactions has passed, and the test *should* behave deterministically regardless of tx/block value

talkol:
yes.. but adding manual timeouts with methods like `network.WaitForTransactionInState` doesn't work in this case.. I played with it quite a lot. It fails the test..

The flakiness results IMO when the first and second tx are placed in the same block.

This is the current code:
```
network.SendTransferInBackground(ctx, 0, 0, 5, 6) // send a transaction so that network advances to block 1. the tamper prevents COMMITTED messages from reaching leader, so it doesn't move to block 2
		committedLatch.Wait()                             // wait for validator to try acknowledge that it reached block 1 (and fail)
		committedLatch.Wait()                             // wait for another consensus round (to make sure transaction(0) does not arrive after transaction(17) due to scheduling flakiness

		txHash := network.SendTransferInBackground(ctx, 0, 17, 5, 6) // this should be included in block 2 which will not be closed until leader knows network is at block 2
```

The dual `committedLatch.Wait()` should add some timeout to make sure this doesn't happen, but it's flaky - it's not strong enough. Adding `network.WaitForTransactionInState` in the middle fails the test because the timeout is too big.

The only other alternative I can think of is adding another `committedLatch.Wait()` in the middle, but this is voodoo..

electricmonk:
I’d really prefer to make the test deterministic without assuming block size, but we can merge this and then fix the test

Shai

PR #359, opened 2018-10-24
Use milliseconds for metrics

netoneko:


instrumentation/metric/histogram.go
```diff
@@ -16,24 +16,24 @@ type Histogram struct {
 
 type histogramExport struct {
 	Name    string
-	Min     int64
-	P50     int64
-	P95     int64
-	P99     int64
-	Max     int64
+	Min     float64
```
electricmonk: The conversion to millis should happen only in `LogRow`

instrumentation/metric/histogram.go
```diff
@@ -16,24 +16,24 @@ type Histogram struct {
...
+	Max     float64
 	Avg     float64
 	Samples int64
 }
 
 func newHistogram(name string, max int64) *Histogram {
 	return &Histogram{
 		namedMetric: namedMetric{name: name},
-		histo:       hdrhistogram.NewWindowed(5, max, 1, 3),
+		histo:       hdrhistogram.NewWindowed(5, 0, max, 1),
```
electricmonk: ok, I see that you fixed a bug where min was set to max, but why change sigfigs to 1?

instrumentation/metric/histogram.go
```diff
@@ -16,24 +16,24 @@ type Histogram struct {
...
 	}
 }
 
 func (h *Histogram) RecordSince(t time.Time) {
-	d := time.Since(t)
+	d := time.Since(t).Nanoseconds()
```
electricmonk: does this do anything other than signify that we're measuring in millis?

instrumentation/metric/histogram.go
```diff
@@ -65,14 +65,14 @@ func (h *Histogram) String() string {
...
-	return histogramExport{
+	return &histogramExport{
 		h.name,
-		histo.Min(),
-		histo.ValueAtQuantile(50),
-		histo.ValueAtQuantile(95),
-		histo.ValueAtQuantile(99),
-		histo.Max(),
-		histo.Mean(),
+		float64(histo.Min()) / 1e+6,
```
electricmonk: so as I said above, this shouldn't be here

instrumentation/metric/histogram.go
```diff
@@ -81,15 +81,19 @@ func (h *Histogram) Rotate() {
...
+
 	return []*log.Field{
 		log.String("metric", h.Name),
 		log.String("metric-type", "histogram"),
-		log.Int64("min", h.Min),
-		log.Int64("p50", h.P50),
-		log.Int64("p95", h.P95),
-		log.Int64("p99", h.P99),
-		log.Int64("max", h.Max),
+		log.Float64("min", h.Min),
```
electricmonk: convert to float millis here; you can create a function `toMillisFloat` that takes an in64 and handles both casting and division by 1000000

netoneko: It can't take int64 because it also has to take h.Mean() which returns float.

instrumentation/metric/histogram.go
netoneko: no

instrumentation/metric/histogram.go
netoneko: I disagree because these values are going to be exported through http api. They should be uniform.

PR #360, opened 2018-10-24
test to simulate state-storage startup time after 6 months of 100 tps on average

ronnno:
test is run manually as the file starts with _. this is because it takes about 30 minutes to run on a typical dev env


talkol:
Since this is a manual test, I prefer the following filename convention which we already use elsewhere in the system:

`services/statestorage/test/_manual/stress_test.go`

[Example](https://github.com/orbs-network/orbs-network-go/tree/master/test/_manual)

and add a small readme.md next to it in the folder explaining how to run it and why

services/statestorage/test/_stress_test.go
```diff
@@ -0,0 +1,79 @@
...
+			d.commitStateDiff(ctx, commit)
+
+			// print progress every 100000 commits
+			if commits++; commits % 100000 == 0 {
+				ms := runtime.MemStats{}
+				runtime.ReadMemStats(&ms)
+				elapsedTick := time.Now().Sub(tickStart)
+				elapsed := time.Now().Sub(start)
+				commitDuration = elapsed - createInputDuration
+				fmt.Printf("tick: %v, commit: %v, elapsed: %v, progress: %d, HeapSys: %dMB, HeapAlloc: %dMB\n", elapsedTick, commitDuration, elapsed, 100*txCount/TxCount6Months100Tps, ms.HeapSys/ (1024 * 1024), ms.HeapAlloc/ (1024 * 1024))
```
talkol: Consider `t.Log` / `t.Logf` instead of `fmt.Printf`. If the system has its own stdout then these outputs are hard to find whereas `t.Log` is designed for this purpose

services/statestorage/test/_stress_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/orbs-network/orbs-network-go/test/builders"
+	"github.com/orbs-network/orbs-spec/types/go/services"
+	"github.com/stretchr/testify/require"
+	"math/rand"
+	"runtime"
+	"testing"
+	"time"
+)
+
+const TpxAvg = 100
```
talkol: I think our convention for consts (which is different than go's default) is: `TPX_AVG`

services/statestorage/test/_stress_test.go
```diff
@@ -0,0 +1,79 @@
+package test
```
talkol: Since this is a manual test, I prefer the following filename convention which we already use elsewhere in the system:

`services/statestorage/test/_manual/stress_test.go`

[Example](https://github.com/orbs-network/orbs-network-go/tree/master/test/_manual)

and add a small README.md next to it in the folder explaining how to run it and why

services/statestorage/test/_stress_test.go
ronnno: In the summary log I changed to `t.Logf` but in this line I left in as is because t.Logf does not print until the test is done and this is a progress prompt

services/statestorage/test/_stress_test.go
ronnno: conformed

services/statestorage/test/_stress_test.go
ronnno: moved file and added README.md

talkol:
Why does acceptance fail now? It seems to me like you didn't touch anything there

PR #361, opened 2018-10-25
trigger support for context

jlevison:
should make a lot more sense now to use it


synchronization/trigger.go
```diff
@@ -1,19 +1,20 @@
...
 	"sync"
 	"sync/atomic"
 	"time"
 )
 
 type Trigger interface {
-	Start()
-	Reset(duration time.Duration)
+	Start(ctx context.Context)
+	Reset(ctx context.Context, duration time.Duration)
```
electricmonk: does this make sense in terms of API? when I have to reset, do I have a long-living context? I'm afraid people might mistakenly pass the ephemeral context they currently have (http-bound, for instance) when calling reset

come to think about it, why do we need `Start` at all? why not pass the context in the factory function and start the goroutine with the context immediately?

synchronization/trigger.go
```diff
@@ -1,19 +1,20 @@
...
 	TimesTriggered() uint64
 	TimesReset() uint64
 	TimesTriggeredManually() uint64
 	IsRunning() bool
-	FireNow()
+	FireNow(ctx context.Context)
```
electricmonk: same error prone issue as above; furthermore - why do we need `FireNow` at all?

synchronization/trigger.go
jlevison: this is to avoid keeping an internal reference to a context, which is considered a bad idea.
i can remove start and of course start the go func as a factory ends (i.e rename Start() to run() )
but this does not absolve us from passing a new context on Reset and FireNow - i think that the caller should have a relevant context for the long-lived trigger at hand - as if he was not using the trigger he would have had some internal implementation of it either way, no?

synchronization/trigger.go
jlevison: context issue is addressed above, but yes, right now i do not think anyone calls FireNow(), not really sure why we still need it other then 'its there and tested...'
the story ofc is 'i need the action to happen right now and not periodically, and i would like to begin waiting for the periodic time again'

synchronization/trigger.go
electricmonk: I think you should kill the `Reset` and `FireNow` features

PR #365, opened 2018-10-25
E2E stress test

netoneko:
I have ran this test locally and my measurements give us `1000` tps locally, which is pretty nice, and `300`+ in docker.

Running on actual testnet, I also got `330` tps.

### DIFF
--- a/services/transactionpool/committed_pool.go
+++ b/services/transactionpool/committed_pool.go
@@ -2,6 +2,7 @@ package transactionpool
 
 import (
 	"context"
+	"github.com/orbs-network/orbs-network-go/instrumentation/metric"
 	"github.com/orbs-network/orbs-spec/types/go/primitives"
 	"github.com/orbs-network/orbs-spec/types/go/protocol"
 	"sync"
@@ -11,12 +12,27 @@ import (
 type committedTxPool struct {
 	transactions map[string]*committedTransaction
 	lock         *sync.RWMutex
+
+	metrics *committedPoolMetrics
 }
 
-func NewCommittedPool() *committedTxPool {
+type committedPoolMetrics struct {
+	transactionCountGauge *metric.Gauge
+	poolSizeInBytesGauge  *metric.Gauge // TODO use this metric
+}
+
+func newCommittedPoolMetrics(factory metric.Factory) *committedPoolMetrics {
+	return &committedPoolMetrics{
+		transactionCountGauge: factory.NewGauge("TransactionPool.CommittedPool.TransactionCount"),
+		poolSizeInBytesGauge:  factory.NewGauge("TransactionPool.CommittedPool.PoolSizeInBytes"),
+	}
+}
+
+func NewCommittedPool(metricFactory metric.Factory) *committedTxPool {
 	return &committedTxPool{
 		transactions: make(map[string]*committedTransaction),
 		lock:         &sync.RWMutex{},
+		metrics:      newCommittedPoolMetrics(metricFactory),
 	}
 }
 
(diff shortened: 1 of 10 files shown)


services/transactionpool/committed_pool.go
```diff
@@ -11,12 +12,27 @@ import (
 type committedTxPool struct {
 	transactions map[string]*committedTransaction
 	lock         *sync.RWMutex
+
+	metrics *committedPoolMetrics
+}
+
+type committedPoolMetrics struct {
+	transactionCountGauge *metric.Gauge
```
electricmonk: why do you call this field `fooGauge`? we know it's a `Gauge`

services/transactionpool/committed_pool.go
```diff
@@ -11,12 +12,27 @@ import (
...
+	poolSizeInBytesGauge  *metric.Gauge // TODO use this metric
 }
 
-func NewCommittedPool() *committedTxPool {
+func newCommittedPoolMetrics(factory metric.Factory) *committedPoolMetrics {
+	return &committedPoolMetrics{
+		transactionCountGauge: factory.NewGauge("TransactionPool.CommittedPool.TransactionCount"),
```
electricmonk: what does this Gauge represent? is it number of transactions seen since startup, or current size of pool?

test/e2e/network_benchmark_test.go
```diff
@@ -31,21 +36,40 @@ func BenchmarkE2ENetwork(b *testing.B) {
...
+	require.True(t, test.Eventually(1*time.Minute, func() bool {
+		m = h.getMetrics()
+		return m != nil
+	}), "could not retrieve metrics")
+
+	txCount := m["TransactionPool.CommittedPool.TransactionCount"]["Value"].(float64)
+
+	expectedNumberOfTx := float64((100 - config.acceptableFailureRate) / 100 * config.numberOfTransactions)
+
+	require.Condition(t, func() (success bool) {
```
electricmonk: why do you need a named return argument?

test/e2e/harness.go
```diff
@@ -17,33 +18,58 @@ import (
...
+	stressTestNumberOfTransactions := int64(10000)
+	stressTestFailureRate := int64(2)
+	stressTestTargetTPS := float64(700)
+
+	if !bootstrap {
+		apiEndpoint = os.Getenv("API_ENDPOINT")
+		url, _ := url.Parse(apiEndpoint)
+		baseUrl = url.Scheme + "://" + url.Host
+
+		stressTestNumberOfTransactions, _ = strconv.ParseInt(os.Getenv("STRESS_TEST_NUMBER_OF_TRANSACTIONS"), 10, 0)
```
electricmonk: I don't really see why we need to propagate these values from the outside

talkol:
Are we planning to run the e2e stress test in CI? won't this add a bunch of flakiness?

electricmonk:
separate step, against deployed testnet

services/transactionpool/committed_pool.go
netoneko: Tx currently in the pool.

test/e2e/network_benchmark_test.go
netoneko: I doesn't have to have a name, but I think it improves readability a bit.

test/e2e/harness.go
netoneko: Because they are different for different conditions. We might want to check how system behaves under 100, 1000, 10000, 100000 tx and so on.

netoneko:
Currently I don't see why not run it on CI, only adds 20 seconds to e2e docker test. I think it's worth it.

services/transactionpool/committed_pool.go
netoneko: Fixed.

PR #368, opened 2018-10-26
Development network is for gamma (like acceptance but with contract compilation), it needs special timeouts

talkol:
`cfg.SetDuration(BLOCK_SYNC_INTERVAL, 100*time.Millisecond)` should be about x10 of `cfg.SetDuration(CONSENSUS_CONTEXT_MINIMAL_BLOCK_DELAY, 10*time.Millisecond)`


netoneko:
Maybe we should rename it to `config.ForGamma` so no one would ever get confused?

PR #371, opened 2018-10-28
Issue/316 use checksum for batch signature

### LINKED ISSUE #316: TransactionForwarder should not aggregate txhashes
Rather, use an aggregative checksum as stated in #119

electricmonk:
fixes #316


talkol:
Please merge master and rerun tests so this can be merged

electricmonk:
done

services/transactionpool/forward_transactions.go
```diff
@@ -140,12 +141,15 @@ func (f *transactionForwarder) drainQueue() []*protocol.SignedTransaction {
 }
 
 func HashTransactions(txs ...*protocol.SignedTransaction) (oneBigHash []byte, hashes []primitives.Sha256) {
+	checksum := adler32.New()
```
talkol: where is this algorithm defined in the spec?

this is incorrect here:
https://github.com/orbs-network/orbs-spec/blob/master/interfaces/protocol/gossipmessages/transaction_relay.proto

electricmonk: it's not in the spec. Shall I update the spec with a PR? crc64 is slow, adler32 is faster and was used as per @noamoded's suggestion. Alternately, we can use FNV which seems to be faster, but is not cryptographic hash - not sure if it matters, as we only need it for checksum purposes.

odednoam: Neither CRC or Adler are cryptographic hashes, so that shouldn't bother us. We need here a checksum, the difference between checksums and hashes is usually that checksums ensure even single bit errors are reflected in the result (for error correction) and hashes focus more on uniform distribution of the results (for higher performance hash tables). What we need here is a checksum; as far as I know FNV is good enough as a checksum so either is fine here.

electricmonk: @OdedWx @talkol please decide, I have no problem switching to either, FNV is easier to use and much faster so it's the natural choice IMO

PR #373, opened 2018-10-28
fixed two incorrect test configs that make one of the block sync tests flaky

jlevison:


services/blockstorage/sync/state_waiting_for_chunks_test.go
```diff
@@ -41,14 +41,14 @@ func TestWaitingMovesToIdleOnTimeout(t *testing.T) {
...
-	h := newBlockSyncHarness().withNodeKey(blocksMessage.Sender.SenderPublicKey())
+	h := newBlockSyncHarness().withNodeKey(blocksMessage.Sender.SenderPublicKey()).withWaitForChunksTimeout(time.Second)
 
 	h.storage.When("LastCommittedBlockHeight").Return(primitives.BlockHeight(10)).Times(1)
 	h.gossip.When("SendBlockSyncRequest", mock.Any, mock.Any).Return(nil, nil).Times(1)
 
 	waitingState := h.sf.CreateWaitingForChunksState(h.config.NodePublicKey())
 	nextState := h.nextState(waitingState, func() {
-		time.Sleep(1 * time.Millisecond) // yield to make sure the state is fully 'running', cannot latch the internal goroutine
+		time.Sleep(10 * time.Millisecond) // yield to make sure the state is fully 'running', cannot latch the internal goroutine
```
electricmonk: can you think of a way to latch to the goroutine? the trigger func shouldn't run before the state has started...

Issue #377, opened 2018-10-28
test name does not show when a panic happens in a goroutine


jlevison:
example test output for **reference**: https://circleci.com/gh/orbs-network/orbs-network-go/2457

when a panic happens inside a goroutine, the go test output will not hold the test name which failed
easy way to reproduce in any test:
```go
	go func() { panic("a") }()
```

not really sure if its a golang issue or something we are doing wrong

electricmonk:
Ok, so it's more difficult than it seems;
What happens when the test runs is that the test actually finishes *before* the goroutine panic, and most likely (as @talkol already discovered) what happens is that the panicing goroutine kills the test runner; a deferred recovery method manages to print that the suite (the package) ran but no one recovered the failing goroutine so we have no idea where the panic came from.

An immediate solution is to add a deferred function calling `recover()` inside the gossip goroutine. A more systematic approach would be to start using a supervision strategy for goroutines.

PR #378, opened 2018-10-28
will help with debugging until #377 is resolved

jlevison:
see #377


jlevison:
rebase this after #385 is merged, it should work fine then

Issue #352, opened 2018-10-23
Acceptance changes required


jlevison:
some of the problems we have with acceptance right now are:
1. it does not output logs, very difficult to debug it
2. the timeout of 20 minutes is too long in CI, it needs to be much lower (takes around 70secs on my machine with 100 count, but maybe the CI is running something more complex)
3. the network harness has a waiting mechanism which is flawed conceptually - if something blocks, the test should die on timeout - and then we get all the goroutines of the system in the log and we can debug - there should not be a 'wait-at-most' function there that panics manually, or if it panics manually, it should output proper logs so we understand why the system froze
4. the waiting 'issue' may not be limited to the network harness but may also be coming from the block tracker

talkol:
More clarification regarding issue 3:

Currently in the system we have various protections against deadlocks with timeouts.. if an operation takes obscene amount of time - we timeout/panic

This behavior is right for production (e2e) but IMO is wrong for acceptance and component tests

If we reach a deadlock in these tests, I prefer to let the system deadlock and reach the test timeout panic which will print stack traces of all goroutines. These stack traces are the main thing that helps with debugging deadlocks. By panicking manually first we miss out on the chance to easily get these stack traces which are essential to debugging

electricmonk:
Re: logs - there's a tradeoff, as in most cases you don't want all of the logs. Is it problematic to manually change (and not commit) the logger definition? if so, maybe we can add an external log config file (but I really hate this idea!) or - better - pass logger override directives via environment variables?

Re: timeout - we need to split the acceptance suite; there are the essential tests, that check logic, which need to have a short timeout; then there are stress tests that fuck up gossip so see how the system behaves; these tests run 100 times, so they need a longer timeout; I think that what we need is to run these tests in a separate build phase than normal acceptance. Thoughts?

Re: 3 - please explain what's the current flaw? I'm not sure I follow what exactly you would like to see. I'm not sure we can use the public api's blocking behavior in all use cases - am I wrong?

jlevison:
re: re: 3, see https://circleci.com/gh/orbs-network/orbs-network-go/2456
log is below but if the test itself timed out, we should have gotten all the goroutines and not just the paniced one, probably the problem with that failure is a system deadlock and we cannot debug it now
Moreover, we cannot reproduce it as its a flakiness issue :/

```
Attaching to project_orbs-acceptance_1
orbs-acceptance_1  | + source ./test.common.sh
[...]

PR #382, opened 2018-10-28
Debug deadlocks on acceptance timeouts

talkol:
When tests time out with panics that usually result from deadlocks, print goroutine stacks to help find what caused the deadlock


talkol:
Handles point 3 in https://github.com/orbs-network/orbs-network-go/issues/352

PR #388, opened 2018-10-29
Feature/basic supervision

### LINKED ISSUE #318: Handle long-running goroutines
Currently if a long-running goroutine panics and crashes, no one will know about it, leaving the node half-dead.

We need to, at the very least, add monitoring when each goroutine dies unexpectedly (via shared infrastructure), or preferably implement some sort of supervision scheme.

This is a blocker for mainnet IMO.

@erankirsh @talkol thoughts?

### LINKED ISSUE #377: test name does not show when a panic happens in a goroutine

electricmonk:
added basic supervision:
* all one-off goroutines are now logging panics as errors to the log
* transaction pool uses long-running supervision which reruns the function when it fails unless context is Done()

fixes #318 
fixes #377


synchronization/supervized/supervisor.go
```diff
@@ -0,0 +1,72 @@
...
+		f()
+	}()
+}
+
+func LongLiving(ctx context.Context, logger Errorer, f func()) {
+	defer recoverPanics(logger)
+	go func() {
+		for {
+			f()
+			select {
```
talkol: Since `f()` most probably has its own `select` inside, I'm not sure how I feel about the external `select`. Obviously the external `select` isn't allowed to block - that's why the `default` has to be there.. Consider changing to an `if ctx.Err() != nil` which is the non-blocking way to check a context is still not done, it sends a better message of not being allowed to block here.

synchronization/supervized/supervisor.go
```diff
@@ -0,0 +1,72 @@
...
+			default:
+				// repeat
+				//TODO count restarts, fail if too many restarts, etc
+			}
+		}
+	}()
+}
+
+func recoverPanics(logger Errorer) {
+	if err := recover(); err != nil {
```
talkol: Recommend to rename `err` to `p` since it isn't necessarily an `Error` type

synchronization/supervized/supervisor.go
```diff
@@ -0,0 +1,72 @@
...
+		e := errors.Errorf("goroutine panicked at [%s]: %s", identifyPanic(), err)
```
talkol: Maybe the second `%s` should be changed to `%v` since we're printing an unknown type

services/transactionpool/service.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/orbs-network/orbs-network-go/instrumentation/log"
 	"github.com/orbs-network/orbs-network-go/instrumentation/metric"
 	"github.com/orbs-network/orbs-network-go/synchronization"
+	"github.com/orbs-network/orbs-network-go/synchronization/supervized"
```
talkol: Maybe rename `supervized` to `goroutine`? `supervized.OneOff` doesn't sound like anything having to do with a thread / goroutine / coroutine. `goroutine.OneOff` is a little more clear that this is a threading mechanism.

Another option is to go with `supervized.LongLived` and `supervized.ShortLived` - it also sounds more like threads..

synchronization/supervized/supervisor.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/pkg/errors"
+	"runtime"
+	"strings"
+)
+
+type Errorer interface {
+	Error(message string, fields ...*log.Field)
+}
+
+func OneOff(logger Errorer, f func()) {
```
talkol: Maybe rename `OneOff` to `ShortLived`?

test/harness/services/gossip/adapter/tampering_transport_test.go
```diff
@@ -21,7 +22,7 @@ func newTamperingHarness() *tamperingHarness {
 	senderKey := "sender"
 	listenerKey := "listener"
 	listener := &mockListener{}
-	transport := NewTamperingTransport()
+	transport := NewTamperingTransport(log.GetLogger(log.Service("transport")))
```
talkol: `transport` isn't really a service.. The convention I used so far in the system was `log.String("adapter", "gossip")`. Search the project for `log.String("adapter",` - we have a few of those already (in all the adapters that have logs).

Maybe we should add `log.Adapter("gossip")` which is a parallel to `log.Service("gossip")`

synchronization/supervized/supervisor.go
```diff
@@ -0,0 +1,72 @@
...
+				return
+			default:
+				// repeat
+				//TODO count restarts, fail if too many restarts, etc
+			}
+		}
+	}()
+}
+
+func recoverPanics(logger Errorer) {
```
talkol: `recoverPanics` will effectively kill all stack traces on panics in the IDE when running tests.. It's an important mechanism for debugging..

Some alternatives to mitigate this:

1. Disable `recoverPanics` in test mode and only do this in production

2. When recovering from a panic, print a stack trace manually to log. Also, in testing mode, make sure the IDE identifies this stack trace and handles it like the default stack traces

talkol: This is the main issue I want to think about together, all the others are cosmetic

synchronization/supervized/supervisor.go
electricmonk: good point!

synchronization/supervized/supervisor.go
electricmonk: that's not the semantic; rather - the OneOff is just a single run, while the LongLived should not be allowed to return until context is closed.

test/harness/services/gossip/adapter/tampering_transport_test.go
electricmonk: yes

synchronization/supervized/supervisor.go
electricmonk: on the other hand, it allows us to understand where the panic comes from, no? that's the whole point of this function.

I'll add the stack trace to the log

electricmonk: added

electricmonk:
also fixes #318

synchronization/supervized/supervisor.go
talkol: `ShortLived` can have the same semantics. `LongLived` means needs to destroy the context to end, `ShortLived` doesn't need a context at all because it terminates quickly

services/transactionpool/service.go
talkol: Just find me a better name than `supervized.OneOff` - it's not clear what it does

PR #404, opened 2018-10-31
Fail acceptance tests on unexpected errors

electricmonk:
Acceptance Suite will no longer allow errors not specifically marked as allowed. Any such error logged will fail the suite.

@jlevison I added some weird errors in block sync and storage as expected since they occur, but I can't explain them properly

### DIFF
--- a/instrumentation/log/error_recording_logger.go
+++ b/instrumentation/log/error_recording_logger.go
@@ -1,6 +1,7 @@
 package log
 
 import (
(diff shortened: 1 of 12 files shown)


electricmonk:
@talkol now there's also one in the stress test:
```golang
   "consensus round tick failed", // (aborting shared state update due to inconsistency) //TODO investigate and explain, or fix and remove expected error
```

PR #405, opened 2018-10-31
Fix/in process network mess

electricmonk:


instrumentation/log/error_recording_logger.go
```diff
@@ -0,0 +1,116 @@
+package log
+
+import (
+	"fmt"
+	"sync"
+)
+
+type ErrorRecordingLogger struct {
```
netoneko: Are you sure we need another logger? What's the reasoning behind this?

electricmonk: not another logger - a decorator

netoneko:
@electricmonk please resolve merge conflicts and merge.

PR #407, opened 2018-10-31
Refactor/move crypto to client

talkol:
Some of the code we had in /crypto was written for the client
Since the client is in a different repo (orbs-client-sdk-go), I moved this code there


netoneko:
@talkol this pr has conflicts

talkol:
@netoneko because it became stale with how long it's waiting. I'll fix now, please merge

PR #409, opened 2018-10-31
Reduce e2e testnet target tps from 250 to 200

netoneko:


electricmonk:
@netoneko this is the kind of change I think is perfectly fine to do directly via Github by editing the file

Issue #305, opened 2018-10-07
block storage - GetTransactionReceipt 


noambergIL:
When the timestamp is not translated to blocks this function returns nil object + error
i suggest it returns the createEmptyTransactionReceiptResult so that can be used to return the current block + time of the block storage ... 

please note : the publicapi.GetTransactionStatus will need to be fixed as well.

PR #414, opened 2018-10-31
fix ide issue for running all tests

noambergIL:
fix issue #305


noambergIL:
@jlevison you may look at the service.go file when you want.

PR #423, opened 2018-11-01
Refactor file config tests

### LINKED ISSUE #265: Refactor file config tests to use specific config examples and not one big example
by @electricmonk 

> You have several tests all sharing the same config stub. What might happen is that all tests fail on parse error if any of the parse method is broken. I'd prefer creating an inline stub for the minimal config required by each individual test

netoneko:
Fixes #265


netoneko:
@electricmonk done

PR #424, opened 2018-11-01
refactor block sync channels

jlevison:


services/blockstorage/sync/block_sync.go
```diff
@@ -74,25 +88,27 @@ func (bs *BlockSync) syncLoop(ctx context.Context) {
 		// TODO add metrics
 		bs.currentState = bs.currentState.processState(ctx)
 	}
-
 }
 
-func (bs *BlockSync) HandleBlockCommitted() {
+func (bs *BlockSync) HandleBlockCommitted(ctx context.Context) {
+	ctx, _ = context.WithTimeout(ctx, bs.config.BlockSyncNoCommitInterval()/2)
```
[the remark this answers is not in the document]
jlevison: because as you suggested, this should not even be part of the code, the context should have a timeout as its coming from gossip, and we should add a timeout there on every call
until that happens, i wanted to add logic that will have it wait for less than the relevant action timer, so i /2 it

services/blockstorage/sync/block_sync.go
```diff
@@ -74,25 +88,27 @@ func (bs *BlockSync) syncLoop(ctx context.Context) {
...
-		bs.currentState.blockCommitted()
+		bs.currentState.blockCommitted(ctx)
 	}
 }
 
 func (bs *BlockSync) HandleBlockAvailabilityResponse(ctx context.Context, input *gossiptopics.BlockAvailabilityResponseInput) (*gossiptopics.EmptyOutput, error) {
+	ctx, _ = context.WithTimeout(ctx, bs.config.BlockSyncCollectResponseTimeout()/2)
 	if bs.currentState != nil {
-		bs.currentState.gotAvailabilityResponse(input.Message)
+		bs.currentState.gotAvailabilityResponse(ctx, input.Message)
```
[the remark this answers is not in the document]
jlevison: yeah, so was i, not sure why i eventually did it that way when looking at it now.. i'll catch you about that later

jlevison:
@electricmonk , @netoneko  can you look again at how i am instrumenting the metrics? decided to add it here as well

services/blockstorage/sync/block_sync.go
```diff
@@ -37,22 +38,51 @@ type BlockSyncStorage interface {
...
 	config       blockSyncConfig
 	currentState syncState
+	c            *blockSyncConduit
+
+	metrics *stateMachineMetrics
 }
 
-func NewBlockSync(ctx context.Context, config blockSyncConfig, gossip gossiptopics.BlockSync, storage BlockSyncStorage, logger log.BasicLogger) *BlockSync {
+type stateMachineMetrics struct {
+	statesTransitioned *metric.Gauge
```
electricmonk: I think a good question would be "what are you trying to measure?"

services/blockstorage/sync/factory.go
```diff
@@ -2,24 +2,79 @@ package sync
...
 	gossip  gossiptopics.BlockSync
 	storage BlockSyncStorage
+	c       *blockSyncConduit
 	logger  log.BasicLogger
+	metrics *stateMetrics
 }
 
-func NewStateFactory(config blockSyncConfig, gossip gossiptopics.BlockSync, storage BlockSyncStorage, logger log.BasicLogger) *stateFactory {
+type stateMetrics struct {
+	carStateLatency           *metric.Histogram
```
electricmonk: what's the difference between this and the next metric?

services/blockstorage/sync/factory.go
```diff
@@ -2,24 +2,79 @@ package sync
...
+
+type idleStateMetrics struct {
+	stateLatency *metric.Histogram
+	timesReset   *metric.Gauge
+	timesExpired *metric.Gauge
+}
+
+type processingStateMetrics struct {
+	stateLatency           *metric.Histogram
+	throughputRate         *metric.Rate
```
electricmonk: Rate in what? bytes? blocks?

services/blockstorage/sync/block_sync.go
jlevison: "is it even working"
also
"comparing two nodes with similar uptime should have similar values"

given that 99.999% are idle state changes (on each block committed by consensus, we shift from idle to idle)

services/blockstorage/sync/factory.go
jlevison: its a work in progress, apologies for not mentioning it, right now the collecting (car) finished and  waiting are only measting the state transition latency, need to add more metrics there but did not want to invest more efforts in case i am doing something wrong

services/blockstorage/sync/factory.go
jlevison: blocks, so originally it was "block-per-second" but that seems wrong as time is probably controlled by the system viewing the data, so i changed to another not-so-good-name, i guess i should call it blocksRate

services/blockstorage/sync/block_sync.go
```diff
@@ -71,28 +101,31 @@ func NewBlockSync(ctx context.Context, config blockSyncConfig, gossip gossiptopi
...
+func (bs *BlockSync) HandleBlockCommitted(ctx context.Context) {
+	ctx, _ = context.WithTimeout(ctx, bs.config.BlockSyncNoCommitInterval()/2)
 	if bs.currentState != nil {
-		bs.currentState.blockCommitted()
+		bs.currentState.blockCommitted(ctx)
 	}
 }
 
 func (bs *BlockSync) HandleBlockAvailabilityResponse(ctx context.Context, input *gossiptopics.BlockAvailabilityResponseInput) (*gossiptopics.EmptyOutput, error) {
+	ctx, _ = context.WithTimeout(ctx, bs.config.BlockSyncCollectResponseTimeout()/2)
```
golangcibot: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak

services/blockstorage/sync/block_sync.go
```diff
@@ -71,28 +101,31 @@ func NewBlockSync(ctx context.Context, config blockSyncConfig, gossip gossiptopi
...
 	if bs.currentState != nil {
-		bs.currentState.gotAvailabilityResponse(input.Message)
+		bs.currentState.gotAvailabilityResponse(ctx, input.Message)
 	}
 	return nil, nil
 }
 
 func (bs *BlockSync) HandleBlockSyncResponse(ctx context.Context, input *gossiptopics.BlockSyncResponseInput) (*gossiptopics.EmptyOutput, error) {
+	ctx, _ = context.WithTimeout(ctx, bs.config.BlockSyncCollectChunksTimeout()/2)
```
golangcibot: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak

services/blockstorage/block_sync_server_test.go
```diff
@@ -0,0 +1,44 @@
...
+		gossip: gossip,
+	}
+
+	s.lastBlockLock = &sync.RWMutex{}
+
+	lastBlock := builders.BlockPair().WithHeight(10).Build()
+	s.lastCommittedBlock = lastBlock
+	input := builders.BlockAvailabilityRequestInput().WithFirstBlockHeight(5).Build()
+	noPanicFunction := func() {
+		s.sourceHandleBlockAvailabilityRequest(ctx, input.Message)
```
golangcibot: Error return value of `s.sourceHandleBlockAvailabilityRequest` is not checked

jlevison: this is a test douchebag

electricmonk: you realize you're dissing a machine, right?

jlevison: its a bad robot

Issue #402, opened 2018-10-31
arrange dockerfile in Orbs repo. 


cmcm2222:
Orbs github repo looks a bit messy. 
Can I arrange package in Orbs repo?

talkol:
Also take a look at https://github.com/orbs-network/orbs-network-go/pull/427 which started addressing some of the mess

netoneko:
Addressed in #427

Issue #313, opened 2018-10-09
Improve directory structure of general file folders (non-code)


talkol:
In the root of the project we have:

* `/logs` which contain the runtime node logs when saved to file

* `/e2e-config` which contains JSON configs for e2e

Can we organize these folders a bit better?

* Maybe rename `logs` to `_logs` so it's clear it's not part of the build (right now it looks like it's a Go package named `logs`)

* Maybe move `/e2e-config` to be under the `/test/e2e` folder? Something like `/test/e2e/_config` again with the underline convention to say that it isn't part of the build

In general, the `go build` and `go test` tools ignore folders that start with underscore, so it's a clear way to mark folders that aren't part of the build

netoneko:
Fixed in #427

PR #427, opened 2018-11-01
Reorganize build folders

### LINKED ISSUE #384: Rename "acceptance" CI phase to "flakiness" and add some component tests there
1. We want to run more tests under the flakiness count 100 test - for example component tests of block storage. These component tests are flaky too and currently we're missing a structured way to discover the flakiness in them.

2. This means it's not really an "acceptance" phase, it's a "flakiness" phase.

    Rename the following:

    name of the CI phase from `acceptance` to `flakiness`
    `test.acceptance.sh` to `test.flakiness.sh`
    `Dockerfile.acceptance` to `Dockerfile.flakiness`
    `docker-test.acceptance.sh` to `docker-test.flakiness.sh`
    `docker-compose.acceptance.yml` to `docker-compose.flakiness.yml`
    and so forth...

### LINKED ISSUE #458: CircleCI stops running after 10m but tests have timeout of 20m
We should either increase CircleCI's timeout from 10m to >20m, or reduce the tests timeout to <10m.

netoneko:
Addresses #402

Also fixes #384 

Also fixes #458


talkol:
I'll review this tomorrow, please don't merge yet

talkol:
Let's address https://github.com/orbs-network/orbs-network-go/issues/313 this as well in this PR

.circleci/config.yml
```diff
@@ -5,24 +5,24 @@ jobs:
     steps:
       - checkout
       - run: ./git-submodule-checkout.sh
-      - run: ./docker-build.sh
-      - run: ./docker-test.sh
+      - run: ./docker/build/build.sh
+      - run: ./docker/test/test.sh
```
talkol: The `./docker` folder is not really just for docker files, it also has `test.sh` which we use regularly outside docker to test the project

I recommend we rename `./docker` to `./_build` - it's all files related to the project build (docker + test)

The underscore in the beginning is a good convention to indicate this directory doesn't contain go code (since go build ignores directories starting with underscore)

A second fix would be do rename the internal folder `./docker/build` to `./build/docker` because `./build/build` is poop :)

docker/test/docker-compose.yml
```diff
@@ -10,8 +10,8 @@ services:
     ports:
       - 8080:8080
     volumes:
-      - ./logs/:/opt/orbs/logs
-      - ./e2e-config/:/opt/orbs/config
+      - ../../logs/:/opt/orbs/logs
```
talkol: let's find a better place for `./logs` in the project root, maybe `./_logs` or `./_output/logs`

The idea is to indicate that this folder does not contain go source files (it looks like the go package logs)

Also, if we have more types of output files - it's better they're all in the same directory and not scattered in the root level of the project

docker/test/docker-compose.yml
```diff
@@ -10,8 +10,8 @@ services:
...
+      - ../../e2e-config/:/opt/orbs/config
```
talkol: let's find a better place for ./e2e-config` in the project root, maybe `./test/e2e/_config`

Again, the idea of the underscore convention is to indicate that it's not go source files

talkol:
Re: @electricmonk 1. flakiness test script should run all fast tests, not just the acceptance suite

I agree, my only issue that we added the `cpunoise` tag and it works best if packages are not run in parallel (since each will strain the CPU on its own)

therefore I recommend to run everything 100 times without `cpunoise` tag and then run separately the tests that have `cpunoise` support sequentially. Currently it's only `acceptance` and `blockstorage` since they're historically the most flaky

talkol:
Re: @electricmonk 2. can we create a ./bin directory and make sure all compiled binaries end up there?

Agree, but I would name it `./_bin` to indicate it doesn't contain go source files under the bin package.. since the rest of the source code is in top level

test/e2e/harness.go
```diff
@@ -99,7 +99,7 @@ func newHarness() *harness {
 			gossipPeers[publicKey.KeyForMap()] = config.NewHardCodedGossipPeer(uint16(firstRandomPort+i), "127.0.0.1")
 		}
 
-		os.MkdirAll(config.GetProjectSourceRootPath()+"/logs", 0755)
+		os.MkdirAll(config.GetProjectSourceRootPath()+"/_logs", 0755)
```
golangcibot: Error return value of `os.MkdirAll` is not checked

docker/test/docker-compose.yml
netoneko: Moved to `docker/test/e2e-config`.

docker/test/docker-compose.yml
netoneko: Moved to `_logs`.

netoneko:
@talkol @electricmonk I have addressed your comments, please take a look

electricmonk:
what about this comment from a few days ago:
> we have a lot of .sh files - should they all be there? maybe most of them should move under ./test? maybe trace-transaction.sh should move under devtools?

.circleci/config.yml
talkol: what about `./_docker` then?

talkol:
@netoneko I'm fine with the changes, good work!

itamararjuan:
I would recommend the following as well:

while `docker` is a good folder for Docker related bash scripts, adding one folder which consolidates all
DevOps related resources into one folder would make much more sense.

I recommend adding an `ops` directory and inside of it subfolder. one of them of course can be `docker` and within it one for the `Dockerfile`, one for the `scripts`, etc..

Also in addition, please add a `bin` folder that all compilation results will go into to make the root folder even more tidy.

@netoneko @electricmonk Thoughts?

electricmonk:
I'm not sure about aggregating all ops-related stuff under `./ops`. Can you point to other projects using this convention? what _is_ the best practice in other projects? 

I think `bin` was handled, but it's `_bin`. @netoneko?

itamararjuan:
> I'm not sure about aggregating all ops-related stuff under `./ops`. Can you point to other projects using this convention? what _is_ the best practice in other projects?
> 
> I think `bin` was handled, but it's `_bin`. @netoneko?

Why `_bin` ? that is definitely not a common practice. It's usually `bin`

As for your comments about an `ops` folder - it is definitely something I saw and utilized myself. Sometimes even an ops repository. Although in our case a separate repo would be an overkill.

itamararjuan:
`ops` is simply a TLD in this case. Just like the `services` (hexagon ones) hosts all the various `services` in the system

electricmonk:
I understand, but is this something someone new to the project would expect? please try to find some references as to how other projects (big open source projects, other blockchains) are organized

netoneko:
@electricmonk I think we have culled most of the scripts already and there is no apparent need to move test scripts to `./test`. The scripts that are test-related or build-related are used quite frequently, so I think it's worth keeping them in the root for now.

@itamararjuan @electricmonk I don't think we need an `ops` folder because all the operations-related things are handled in different repos anyway.

`_bin` was used because @talkol suggested that temporary files/directories should start with `_`, and `_bin` is a temporary directory produced after building the binaries.

@electricmonk regarding `trace-transaction.sh` - I think we should wait for your request id and rework it or remove it altogether.

netoneko:
Since Tal is fine with current changes, I'm going to merge this pr tomorrow morning and we can pick up any additional refactoring/tidying up in the next pull request.

PR #431, opened 2018-11-01
Added PublicApi field to return immediately in SendTransaction

talkol:


electricmonk:
I really don’t like this
Boolean flags are considered harmful
Why not add another endpoint, DepositTransaction or a bette name, indicating the different semantics

Shai

PR #432, opened 2018-11-01
return immediately for public api send transaction

noambergIL:


services/publicapi/test/send_transaction_test.go
```diff
@@ -112,3 +112,27 @@ func TestSendTransaction_TimesOut(t *testing.T) {
...
 	})
 }
+
+func TestSendTransaction_ReturnImmediately(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		timeout := 100 * time.Second // won't actually wait please don't change
+		harness := newPublicApiHarness(ctx, timeout)
+
+		txb := builders.Transaction().Builder()
+		harness.onAddNewTransaction(func() {})
```
ronnno: suggest cleaning up this line for readability, passing an empty func(){} for no reason ins't nice

services/publicapi/test/send_transaction_test.go
```diff
@@ -112,3 +112,27 @@ func TestSendTransaction_TimesOut(t *testing.T) {
...
+
+		start := time.Now()
+		result, _ := harness.papi.SendTransaction(ctx, &services.SendTransactionInput{
+			ClientRequest: (&client.SendTransactionRequestBuilder{
+				SignedTransaction: txb,
+			}).Build(),
+			ReturnImmediately: 1,
+		})
+
+		harness.verifyMocks(t) // contract test
```
ronnno: this is not the focus of this test, we don't need this here too..

electricmonk:
please revert this
boolean flags are considered harmful

Issue #400, opened 2018-10-30
flakiness in TestGammaCliDeployWithUserDefinedContract


ronnno:
see
https://circleci.com/gh/orbs-network/orbs-network-go/2616

talkol:
```
--- FAIL: TestGammaCliDeployWithUserDefinedContract (8.68s)
	require.go:201: 
			Error Trace:	gamma_server_and_cli_test.go:220
			            				gamma_server_and_cli_test.go:282
			Error:      	Not equal: 
			            	expected: 0x674
			            	actual  : 0x0
			Test:       	TestGammaCliDeployWithUserDefinedContract
			Messages:   	counter value did not match expected one
```

talkol:
it's similar to this open issue: https://github.com/orbs-network/orbs-network-go/issues/370

electricmonk:
it's a duplicate of #370 actually

electricmonk:
We think that the call to `getCounterValue` reaches the system before `StateStorage` reflects the transaction. (intra-node sync hasn't finished for the block containing the transaction).

Evidence:
transaction for contract with name Counter, method add has tx hash `72582ca45c77eaedabdc804d308b2b36d216932f39f6ec5bbb481fa2f5f290f6`:

```
info 2018-10-29T12:21:17.679423Z adding new transaction to the pool node=dfc06c service=transaction-pool flow=checkpoint transaction={Transaction:{ProtocolVersion:1,VirtualChainId:2a,Timestamp:156213b1fc30b5ab,Signer:{Scheme:(Eddsa){NetworkType:NETWORK_TYPE_TEST_NET,SignerPublicKey:dfc06c5be24a67adee80b35ab4f147bb1a35c55ff85eda69f40ef827bddec173,},},ContractName:Counter,MethodName:add,InputArgumentArray:1400000006000000616d6f756e7401007406000000000000,},Signature:d990551a7b0ab79834b66d634f9fe14355583dfdb313a03979df62780c291c7aabbebfc2faf81793da406b0a86734d989c0389bf26999f6b5f5520bbf97b8800,} txHash=72582ca45c77eaedabdc804d308b2b36d216932f39f6ec5bbb481fa2f5f290f6 function=transactionpool.(*service).AddNewTransaction source=services/transactionpool/add_new_transaction.go:15 
```

transaction with hash `72582ca45c77eaedabdc804d308b2b36d216932f39f6ec5bbb481fa2f5f290f6` entered `block-height=c`

[...]

Issue #370, opened 2018-10-27
Flakiness in gamma-cli test


electricmonk:
https://circleci.com/gh/orbs-network/orbs-network-go/2453

PR #435, opened 2018-11-01
Fixing flakiness in Gamma test

electricmonk:
Gamma test is susceptible to a race condition in the system where state storage can lag behind block storage and VM is updating its block height from state storage. The long term solution is to update VM's block height from block storage but for now the test will re-query the system if it returned a response with the wrong block height.


electricmonk:
fixes #400 
fixes #370

PR #440, opened 2018-11-04
BlockTracker no longer takes timeout

electricmonk:
Changed block tracker's API to rely on context cancellation for timing out
This transfers responsibility for setting timeout for the caller, but it makes it more flexible and ensures that timeouts can be global for a specific use case or test.

It also enables sending timeouts from the client .

### DIFF
--- a/services/blockstorage/test/commit_block_test.go
+++ b/services/blockstorage/test/commit_block_test.go
@@ -131,7 +131,7 @@ func TestCommitBlockWithSameTransactionTwice(t *testing.T) {
 		_, err = harness.commitBlock(ctx, block1)
 		require.NoError(t, err)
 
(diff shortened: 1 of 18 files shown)


ronnno:
This looks great.
I have just two suggestions
1. when WaitForBlock returns an error, consider using context.Err() (either wrapped or not wrapped) so that client code can examine the reason for failure. It's just a smell that we return an error that says "timed out" when it could be a shutdown.

1. I would consider creating a second c'tor: `NewBlockTrackerWithFixedTimeout` to reduce verbosity in client code. the implementation would be similar to the old implementation and it would allow both flavors of usage.

electricmonk:
re: 1 - done
re: 2 - YAGNI

PR #443, opened 2018-11-04
Refactored config and created dedicated e2e config to make e2e faster

talkol:


services/consensusalgo/benchmarkconsensus/common.go
```diff
@@ -41,7 +41,7 @@ func (s *service) setLastCommittedBlock(blockPair *protocol.BlockPairContainer,
 }
 
 func (s *service) requiredQuorumSize() int {
-	return int(math.Ceil(float64(s.config.NetworkSize(0)) * float64(s.config.ConsensusContextPercentageOfNodesRequiredForConsensus()/100)))
+	return int(math.Ceil(float64(s.config.NetworkSize(0)) * float64(s.config.ConsensusRequiredQuorumPercentage()/100)))
```
electricmonk: why not make the config a float?

talkol: Because Kirill doesn't support float config values and didn't want to add them just for a single field

PR #445, opened 2018-11-04
Delete e2e build artifacts only once when e2e starts instead of after every test

talkol:


test/e2e/main_test.go
```diff
@@ -0,0 +1,16 @@
+package e2e
+
+import (
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
+	_, dirToCleanup := getProcessorArtifactPath()
+	os.RemoveAll(dirToCleanup)
```
electricmonk: why clean both before and after?
can you extract this logic to a named function, `cleanNativeProcessorCache`?

PR #450, opened 2018-11-05
Added requestId field to logs and colorized according to it

electricmonk:


instrumentation/trace/context.go
```diff
@@ -0,0 +1,50 @@
...
+	name      string
+	requestId string
+}
+
+func NewContext(parent context.Context, name string) context.Context {
+	now := time.Now()
+	ep := &Context{
+		name:      name,
+		created:   now,
+		requestId: fmt.Sprintf("%s-%d", name, now.UnixNano()),
```
netoneko: I suggest we start using uuid v4 right away.

electricmonk: why?

netoneko: because it's unique and does not rely on timestamps

electricmonk: don't you think it's safe to assume that nano timestamps are also unique?

electricmonk: | They might be reasonably unique, but we are going to propagate them between nodes.

It's a good point, but I'm not sure that we'll propagate request id between nodes. We'll discuss this on Thursday.

But - really, what are the odds of two transactions arriving in the same nanosecond anywhere on the node? also, we can add node id to the request id which *will* make it more unique.

The problem with `uuid`s, btw, is that it requires an external dependency and we're trying to limit them.

noambergIL: i suggest we using something short for reuqest id.

services/blockstorage/service.go
```diff
@@ -82,8 +85,10 @@ func (s *service) GetLastCommittedBlockHeight(ctx context.Context, input *servic
 }
 
 func (s *service) CommitBlock(ctx context.Context, input *services.CommitBlockInput) (*services.CommitBlockOutput, error) {
+	logger := s.logger.WithTags(trace.LogFieldFrom(ctx))
+
 	txBlockHeader := input.BlockPair.TransactionsBlock.Header
```
noambergIL: for me this can be a blocker. it feels unnatural to remember you need to create a logger with tags each time you start ...

services/consensusalgo/benchmarkconsensus/common.go
```diff
@@ -45,10 +46,12 @@ func (s *service) requiredQuorumSize() int {
 }
 
 func (s *service) saveToBlockStorage(ctx context.Context, blockPair *protocol.BlockPairContainer) error {
+	logger := s.logger.WithTags(trace.LogFieldFrom(ctx))
```
noambergIL: do we have to create a new trace every func ?

services/publicapi/run_transaction.go
```diff
@@ -4,28 +4,32 @@ import (
...
 
-func (s *service) CallMethod(ctx context.Context, input *services.CallMethodInput) (*services.CallMethodOutput, error) {
+func (s *service) CallMethod(parentCtx context.Context, input *services.CallMethodInput) (*services.CallMethodOutput, error) {
 	if input.ClientRequest == nil {
 		err := errors.Errorf("error: missing input (client request is nil)")
 		s.logger.Info("call method received via public api failed", log.Error(err))
 		return nil, err
 	}
 
+	ctx := trace.NewContext(parentCtx, "PublicApi.SendTransaction")
```
noambergIL: this is call method why is the context here send transaction ?

electricmonk: copy-paste :(

services/consensusalgo/benchmarkconsensus/common.go
electricmonk: no, but we have to pass the trace to the logger somehow

services/blockstorage/service.go
electricmonk: you could also send the context to the log when logging:
```golang
s.logger.info("foobar", trace.LogFieldFrom(ctx))
```

instrumentation/log/formatters.go
```diff
@@ -64,15 +69,15 @@ func printParam(builder *strings.Builder, param *Field) {
...
+		value = param.StringVal
 	case ServiceType:
-		value = param.String
+		value = param.StringVal
 	case FunctionType:
-		value = param.String
+		value = param.StringVal
 	case SourceType:
-		value = param.String
+		value = param.StringVal
```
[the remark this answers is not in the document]
electricmonk: I don't think there's fallthrough in go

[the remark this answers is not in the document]
noambergIL: v := 42
switch v {
case 100:
	fmt.Println(100)
	fallthrough
case 42:
	fmt.Println(42)
	fallthrough
case 1:
	fmt.Println(1)
	fallthrough
default:
	fmt.Println("default")
}

PR #456, opened 2018-11-06
Fix helix network communication test

netoneko:


electricmonk:
already fixed in master

PR #459, opened 2018-11-06
CONTRIBUTING.md - added section on downloading logs from CircleCI

IdoZilberberg:


electricmonk:
@IdoZilberberg I think this is the wrong fix. @netoneko - why aren't the logs being tarballed in some cases?

PR #462, opened 2018-11-06
Use linter and fix ~100 issues.

noambergIL:
1) I tweeked Ido's yml a bit - the ignore part doesn't seem to work
2) i did a quick review of tests but in the commit its "excluded" - only errchk and unused  
3) Two errors i'm not sure how to fix at the moment : 
bootstrap/httpserver/server.go:78:29: Error return value of `server.httpServer.Serve` is not checked (errcheck)
		go server.httpServer.Serve(tcpKeepAliveListener{listener.(*net.TCPListener)})
		                          ^
crypto/hash/ripmd160.go:16:9: Error return value of `r.Write` is not checked (errcheck)
	r.Write(hash[:])
4) three unused, they seem to be false positives ... they only existed when i was about commit.
5) opened issues for 3 excluded dirs to their owners.


bootstrap/httpserver/server.go
```diff
@@ -47,8 +47,14 @@ func (ln tcpKeepAliveListener) Accept() (net.Conn, error) {
 	if err != nil {
 		return nil, err
 	}
-	tc.SetKeepAlive(true)
-	tc.SetKeepAlivePeriod(3 * time.Minute)
+	err = tc.SetKeepAlive(true)
+	if err != nil {
```
electricmonk: all of these patterns can be better like this:
```suggestion
	if err := tc.SetKeepAlive(true); err != nil {
```

bootstrap/httpserver/server.go
```diff
@@ -217,20 +226,26 @@ func translateStatusToHttpCode(responseCode protocol.RequestStatus) int {
 	return http.StatusNotImplemented
 }
 
-func writeMembuffResponse(w http.ResponseWriter, message membuffers.Message, httpCode int, orbsText string) {
+func (s *server) writeMembuffResponse(w http.ResponseWriter, message membuffers.Message, httpCode int, orbsText string) {
```
electricmonk: why did you make this change?

config/file_config.go
```diff
@@ -50,7 +50,8 @@ func parseNodesAndPeers(value interface{}) (nodes map[string]FederationNode, pee
 		for _, item := range nodeList {
 			kv := item.(map[string]interface{})
 
-			if publicKey, err := hex.DecodeString(kv["Key"].(string)); err == nil {
+			var publicKey []byte
+			if publicKey, err = hex.DecodeString(kv["Key"].(string)); err == nil {
```
electricmonk: I actually prefer the older way

config/file_config.go
```diff
@@ -75,19 +76,18 @@ func parseNodesAndPeers(value interface{}) (nodes map[string]FederationNode, pee
...
+		var err error
 
 		switch value.(type) {
 		case float64:
 			numericValue, err = parseUint32(value.(float64))
 		case string:
-			if parsedDuration, err := time.ParseDuration(value.(string)); err == nil {
-				duration = parsedDuration
-			}
+			duration, err = time.ParseDuration(value.(string))
```
electricmonk: didn't you change behavior here? what if the parse failed?

devtools/gammacli/commands/run.go
```diff
@@ -100,7 +104,7 @@ func (r *CommandRunner) HandleRunCommand(args []string) (string, error) {
 			return "", err
 		}
 
-		if err := json.Unmarshal(jsonBytes, tx); err != nil {
+		if err = json.Unmarshal(jsonBytes, tx); err != nil {
```
electricmonk: I don't understand this change

services/publicapi/waiter_test.go
```diff
@@ -190,7 +190,8 @@ func TestPublicApiWaiter_WaitGracefulShutdown(t *testing.T) {
 		wc2 := waiter.add(key)
 
 		var waitTillCancelled = func(wc *waiterChannel) {
-			ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
+			var cancel func()
+			ctx, cancel = context.WithTimeout(ctx, 1*time.Second)
```
electricmonk: why?

test/e2e/deploy_native_test.go
```diff
@@ -50,10 +50,10 @@ func TestDeploymentOfNativeContract(t *testing.T) {
 			WithMethod(primitives.ContractName(fmt.Sprintf("CounterFrom%d", counterStart)), "get")
 
 		printTestTime(t, "call method - start", &lt)
-		response, err := h.callMethod(getCounter.Builder())
+		response, err2 := h.callMethod(getCounter.Builder())
 		printTestTime(t, "call method - end", &lt)
 
-		if err == nil && response.CallMethodResult() == protocol.EXECUTION_RESULT_SUCCESS {
+		if err2 == nil && response.CallMethodResult() == protocol.EXECUTION_RESULT_SUCCESS {
```
electricmonk: err2?

test/e2e/harness.go
```diff
@@ -54,10 +54,10 @@ func getConfig() E2EConfig {
 	stressTestFailureRate := int64(2)
 	stressTestTargetTPS := float64(700)
 
-	if !bootstrap {
+	if !isBootstrap {
```
electricmonk: I disagree with this rename. If `bootstrap` is problematic, use `shouldBootstrap`

bootstrap/httpserver/server.go
noambergIL: because i needed the logger to log the error and didn't want to pass it in argument ... (i actually changed a function that was writeandlog that got the log as argument instead of receiver)

config/file_config.go
noambergIL: the linter sais since the err is in the declaration of the func then its overshadows ...

config/file_config.go
noambergIL: i looked and if parse fails the duration is 0.

devtools/gammacli/commands/run.go
noambergIL: i think again it was about overshadow as there is multiple use of err and its very unclear. i'd fix up the whole func if the owner would do it.

services/publicapi/waiter_test.go
noambergIL: overshadow  ... do you prefer the internal ctx will have a diff name ?

test/e2e/deploy_native_test.go
noambergIL: 1) test
2) overshadow

services/publicapi/waiter_test.go
noambergIL: changed name

test/e2e/harness.go
noambergIL: changed.

config/file_config.go
electricmonk: yeah, because that `err` is not scope to its `if` check. Fix that issue, this issue will disappear

config/file_config.go
electricmonk: ok, but in general does it make sense to have a zero duration as default?

devtools/gammacli/commands/run.go
electricmonk: sit with @itamararjuan and fix it?

services/publicapi/waiter_test.go
electricmonk: yes

test/e2e/deploy_native_test.go
electricmonk: again, overshadowing is a smell that variables are reassigned. The previous `err` needs to be scoped.

PR #478, opened 2018-11-08
Fix required quorum size

netoneko:
The test is terrible, don't merge yet.


netoneko:
@electricmonk please advise how to get rid of this ugly test.

PR #481, opened 2018-11-11
Changed protos and contract sdk for ethereum connector

talkol:


services/processor/native/sdk_ethereum.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+	if len(output.OutputArguments) != 1 || !output.OutputArguments[0].IsTypeBytesValue() {
+		return errors.Errorf("callMethod Sdk.Ethereum returned corrupt output value")
+	}
+
+	return ethereumUnpackOutput(out, output.OutputArguments[0].BytesValue())
+}
+
+func ethereumPackInputArguments(jsonAbi string, method string, args ...interface{}) ([]byte, error) {
```
talkol: Recommend to change args to `args []interface` and then change the call to 

`ethereumPackInputArguments(jsonAbi, methodName, args)`

jlevison:
approved the branch so @talkol  can review some minor changes i made and we can merge - connector work is not complete yet

bootstrap/node_logic.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/orbs-network/orbs-network-go/services/consensusalgo/leanhelixconsensus"
 	"github.com/orbs-network/orbs-network-go/services/consensuscontext"
 	"github.com/orbs-network/orbs-network-go/services/crosschainconnector/ethereum"
+	ethereumConnectorAdapter "github.com/orbs-network/orbs-network-go/services/crosschainconnector/ethereum/adapter"
```
talkol: Let's rename to `ethereumAdapter`

We have the crosschainConnector (the service) which has an adapter
Calling the adapter a connector too is confusing

bootstrap/node_logic.go
```diff
@@ -43,13 +44,14 @@ func NewNodeLogic(
 	logger log.BasicLogger,
 	metricRegistry metric.Registry,
 	nodeConfig config.NodeConfig,
+	ethereumConnector ethereumConnectorAdapter.EthereumConnection,
```
talkol: Let's rename to `ethereumAdapter`

services/crosschainconnector/ethereum/adapter/node_access.go
```diff
@@ -0,0 +1,28 @@
+package adapter
+
+import (
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+type EthereumNodeConnector struct {
```
talkol: Let's rename and remove the word `Connector`

Maybe `EthereumAdapter`.. the word `connector` is overused on the service level

services/crosschainconnector/ethereum/service.go
```diff
@@ -2,16 +2,71 @@ package ethereum
...
+	"github.com/orbs-network/orbs-network-go/services/crosschainconnector/ethereum/adapter"
 	"github.com/orbs-network/orbs-spec/types/go/services"
 )
 
+type ethereumConnectorConfig interface {
+	EthereumEndpoint() string
+}
+
 type service struct {
+	connector adapter.EthereumConnection
```
talkol: rename `connector` to `connection` or `adapter`

services/crosschainconnector/ethereum/adapter/simulator.go
```diff
@@ -0,0 +1,45 @@
...
+		return err
+	}
+	es.auth = bind.NewKeyedTransactor(key)
+
+	genesisAllocation := map[common.Address]core.GenesisAccount{
+		es.auth.From: {Balance: big.NewInt(10000000000)},
+	}
+
+	sim := backends.NewSimulatedBackend(genesisAllocation, 900000000000)
+	es.simClient = sim
```
talkol: This function is running potentially from multiple goroutines, it doesn't look very thread-safe

services/crosschainconnector/ethereum/service.go
```diff
@@ -2,16 +2,71 @@ package ethereum
...
+	logger    log.BasicLogger
+	config    ethereumConnectorConfig
 }
 
-func NewEthereumCrosschainConnector() services.CrosschainConnector {
-	return &service{}
+func NewEthereumCrosschainConnector(ctx context.Context, // TODO: why don't we use context here?
+	config ethereumConnectorConfig,
+	connector adapter.EthereumConnection,
```
talkol: Rename `connector` to `connection` or `adapter`

services/crosschainconnector/ethereum/service.go
```diff
@@ -2,16 +2,71 @@ package ethereum
...
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/orbs-network/orbs-network-go/instrumentation/log"
+	"github.com/orbs-network/orbs-network-go/services/crosschainconnector/ethereum/adapter"
 	"github.com/orbs-network/orbs-spec/types/go/services"
 )
 
+type ethereumConnectorConfig interface {
+	EthereumEndpoint() string
```
talkol: It seems to me that the endpoint is a configuration of the adapter, not the service

I think the service has no config

services/crosschainconnector/ethereum/service.go
```diff
@@ -2,16 +2,71 @@ package ethereum
...
+		connector: connector,
+		logger:    logger,
+		config:    config,
+	}
+
+	return s
+}
+
+func (s *service) setupClient() error {
+	if s.connector.GetClient() == nil {
```
talkol: IMO this entire function should move into the adapter

The service shouldn't bother with restarting clients or connections

The adapter should take in its constructor the endpoint config (it's a config of the adapter, not of the service). Then the adapter will only expose `GetClient()` which will check internally if the client is initialized and initialize it if it isn't.

services/crosschainconnector/ethereum/service.go
```diff
@@ -2,16 +2,71 @@ package ethereum
...
+		if err := s.connector.Dial(s.config.EthereumEndpoint()); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (s *service) EthereumCallContract(ctx context.Context, input *services.EthereumCallContractInput) (*services.EthereumCallContractOutput, error) {
-	panic("Not implemented")
+	if err := s.setupClient(); err != nil { // lazy setup - if the eth node is down it will retry setup on next call
```
talkol: IMO this entire step should be removed and moved into the adapter

netoneko:
@jlevison please resolve the conflicts and merge

services/crosschainconnector/ethereum/adapter/simulator.go
jlevison: you are right, but it is supposed to be test only code, i'll protect it because it makes no difference, this does makes me think i should go over the entire connection flow because i think i will mess up the real node client creation because of a rece condition, will fix that as well

services/crosschainconnector/ethereum/service.go
jlevison: yeah, and i fixed the naming, but we cannot just configure a adapter, the configution injection happens at service level, or am i missing something?

services/crosschainconnector/ethereum/service.go
jlevison: so when we getClient from someplace and the client is nil, the adapter will 'dial' transparently, right?

services/crosschainconnector/ethereum/service.go
jlevison: nvm, its done

PR #484, opened 2018-11-12
Give build tag to disable supervised panic recovery

talkol:


IdoZilberberg:
@electricmonk Please review over this PR. I merged it by mistake even though Tal wanted you to go over it first.

IdoZilberberg:
@talkol This will also work if adding the `norecover` build tag directly to a specific test:
`Edit Configurations... | TextXXX | Go tool arguments` then add `-tags norecover`.
This way you can set the flag for a specific test (and of course in that case don't set it in the global Goland build flags)

electricmonk:
looks good.
what do we want to do in CI?

PR #485, opened 2018-11-12
Bugfix/benchmark consensus zero quorum bug

talkol:


electricmonk:
Can you elaborate about the bug? reference an issue?

netoneko:
@electricmonk this pr continues work started in #478, when I have discovered that the quorum size was set to 0 due to rounding error and the tests didn't caught it because required quorum was 100% and not 66% like in production.

PR #487, opened 2018-11-13
Logger http output

netoneko:
screen shot 2018-11-13 at 13 43 44


instrumentation/log/http_output.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+type row struct {
+	level     string
+	timestamp time.Time
+	message   string
+	fields    []*Field
+}
+
+type httpOutput struct {
```
electricmonk: is this really an `httpOutput?` I think it's a `bufferingOutput` with an `httpWriter`. Or maybe even `logzioHttpWriter` - unless you think it's safe to assume that other log services will use the same http API

instrumentation/log/http_output.go
```diff
@@ -0,0 +1,95 @@
...
+	row := &row{level, time.Now(), message, fields}
+
+	out.lock.Lock()
+	out.logs = append(out.logs, row)
+	out.lock.Unlock()
+
+	out.flush()
+}
+
+func (out *httpOutput) flush() {
```
electricmonk: rename `flushIfNeeded` or `maybeFlush`

instrumentation/log/http_output.go
```diff
@@ -0,0 +1,95 @@
...
+		lock:      &sync.Mutex{},
+	}
+}
+
+type httpWriter struct {
+	url string
+}
+
+func (w *httpWriter) Write(p []byte) (n int, err error) {
+	reader := bytes.NewReader(p)
```
electricmonk: now, let's find a way to write to http via streaming as opposed to allocating a big byte array

instrumentation/log/http_output_test.go
```diff
@@ -0,0 +1,121 @@
...
+type httpOutputHarness struct {
+	port     uint16
+	listener net.Listener
+	router   *http.ServeMux
+	server   *http.Server
+}
+
+func newHttpHarness(handler http.Handler) *httpOutputHarness {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	port := uint16(20000 + r.Intn(40000))
```
electricmonk: please use port 0 to allocate a random port; this pattern is very bad and will eventually result in flakiness.

services/gossip/adapter/direct_transport.go
```diff
@@ -335,8 +335,6 @@ func calcPaddingSize(size uint32) uint32 {
 }
 
 func (t *directTransport) sendKeepAlive(ctx context.Context, conn net.Conn) error {
-	t.logger.Info("sending keepalive", log.String("peer", conn.RemoteAddr().String()))
```
electricmonk: why?

netoneko: Because it's polluting the logs and has no real value, so @talkol agreed to kill it. I kept the log that reports if we failed to send the keep alive request.

instrumentation/log/http_output.go
netoneko: It's already memory-efficient since we allocate a buffer once and then just write to it. There is no double allocation here.

electricmonk: what am I missing? I see that you create a new(bytes.Buffer) each time you flush the logs

netoneko: Yes, it happens on every flush, but before that there was double allocation with `strings.Join`, and now it's gone.

instrumentation/log/bulk_output.go
```diff
@@ -0,0 +1,95 @@
...
+	message   string
+	fields    []*Field
+}
+
+type bulkOutput struct {
+	formatter LogFormatter
+	writer    io.Writer
+
+	bulkSize int
+	delay    time.Duration
```
golangcibot: `delay` is unused

instrumentation/log/http_output_test.go
netoneko: Done.

instrumentation/log/http_output.go
netoneko: Done.

netoneko:
@electricmonk I have implemented what you asked (using a pipe), and I believe that it's a wrong turn for us.

My reasoning:
1. pipes are just fancy channels, so when I'm streaming a bulk of log rows, it will be blocking *on write*. 
2. if I wrap `flushIfNeeded` in a goroutine, I will have a goroutine leak if the streaming request gets stuck.
3. I could write this goroutine in a context, but is it really worth all this trouble if I can simply allocate some memory and send the bulk in a single request?

I have talked to @jlevison and he thinks that the overhead on this solutions is too high and streaming of logs is not worth it.

Here is the snipped of my implementation (I don't want to go through with it):

```golang
func (out *bulkOutput) flushIfNeeded() {
	var logs []*row

	out.lock.Lock()
	if len(out.logs) >= out.bulkSize {
		logs = out.logs
		out.logs = nil
	}
	defer out.lock.Unlock()

	if logs != nil {
		pipeReader, pipeWriter := io.Pipe()
		pipeWriter.Close()

		go func() {
			if err := out.writer.Send(pipeReader); err != nil {
				fmt.Println(fmt.Sprintf("%s failed to send logs via http: %s", time.Now().String(), err))
			}
		}()

		for _, row := range logs {
			pipeWriter.Write([]byte(out.formatter.FormatRow(row.timestamp, row.level, row.message, row.fields...)))
			pipeWriter.Write([]byte("\n"))
		}
	}
}

func (w *httpWriter) Send(body io.Reader) (err error) {
	resp, err := http.Post(w.url, "application/json", body)

	if err != nil {
		return errors.Errorf("Failed to send logs: %s", err)
	}

	defer resp.Body.Close()
	if resp.StatusCode != 200 {
		return errors.Errorf("Failed to send logs: %d, %s", resp.StatusCode, err)
	}

	return err
}
```

electricmonk:
as we discussed over the phone, for now just open an issue describing the possible vulnerability and put a hard limit on `bulkSize`, panicking if it's over 1000 and linking to the issue

electricmonk:
@netoneko approved, please merge and merge :)

PR #491, opened 2018-11-14
Acceptance tests corrupting gossip messages 

electricmonk:
This is an acceptance test we kinda forgot about - it tampers with the message binary data and currently fails. We need to investigate why it fails and make sure that it passes, as it kind of represents a case for a byzantine attacker


talkol:
Didn't forget about it, I'm waiting until @danokoch is finished with the gossip refactor which will add `isValid` tests on all messages before calling the handlers and then we can return to investigate this test. Until gossip does a better job of blocking corrupt messages it's too early to investigate

PR #493, opened 2018-11-14
e2e shared state

### LINKED ISSUE #482: Update e2e to always use random accounts
Right now the `e2e_testnet` step on CI is broken because we rely on the fact that e2e always starts with a clean system: https://circleci.com/gh/orbs-network/orbs-network-go/3633

Now that I have added a step that verifies that the leader was actually synced before proceeding with deployment of other nodes, e2e understandably fails.

electricmonk:
fixes #482


test/e2e/main_test.go
```diff
@@ -8,7 +8,12 @@ import (
 func TestMain(m *testing.M) {
 	cleanNativeProcessorCache()
 
-	exitCode := m.Run()
+	n := newInProcessE2ENetwork()
+
+	m.Run()
+	exitCode := m.Run() // run twice so that any test assuming a clean slate will fail; e2es shouldn't assume anything about system state
```
talkol: It's cool we run twice, although it's the slowest test and now we run it twice.. are we sure we always want to do this?

services/processor/native/repository/BenchmarkToken/contract.go
```diff
@@ -24,7 +24,7 @@ type contract struct{ *sdk.BaseContract }
 
 ///////////////////////////////////////////////////////////////////////////
 
-const TOTAL_SUPPLY = 10000
+const TOTAL_SUPPLY = 1000000
```
talkol: This doesn't really solve the problem for the forever live test net
Eventually it will run out of tokens
It runs twice successfully but won't survive infinite runs

talkol: Opened an issue https://github.com/orbs-network/orbs-network-go/issues/500

PR #499, opened 2018-11-14
Acceptance test asserting quorum size is taken into consideration

### LINKED ISSUE #477: Test startup time of first consensus round
I noticed that in testnet the first consensus round doesn't start until all the nodes are online, although the quorum size is 66%. I would be happy to be proven wrong, but we have no tests for it.

As @talkol suggested, the course of action should be the following:

1. [x] change quorum size from 100% to 66% in acceptance and verify that it runs smoothly
1. [ ] add an acceptance test that starts 2/3 of the nodes and verifies that the blocks are being created

electricmonk:
fixes #477 

Starts a network with 4 out of 6 nodes up and configures for 66% required quorum size. Asserts a block can be closed.

### DIFF
--- a/bootstrap/inmemory/network.go
+++ b/bootstrap/inmemory/network.go
@@ -97,6 +97,9 @@ func (n *Node) WaitForTransactionInState(ctx context.Context, txhash primitives.
 		panic(fmt.Sprintf("statePersistence.WaitUntilCommittedBlockOfHeight failed: %s", err.Error()))
 	}
 }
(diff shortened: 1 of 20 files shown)


talkol:
There are several configuration changes I don't understand, let's go over them f2f

Overall it seems that some configuration disappeared (like gossip peers) and the factory functions are more cluttered with boilerplate which increases the chance for mistakes (people adding a new config value and forgetting to set it in all the factories)

Issue #516, opened 2018-11-21
flakiness in inter node block sync in service sync acceptance test


ronnno:
triggered by TestInternalBlockSync_StateStorage
This test starts two networks and waits until state is synced first by inter-node sync and then by service sync. the break appears to be in the inter-node sync phase
```
--- FAIL: TestInternalBlockSync_StateStorage (0.39s)
	testify_extensions.go:41: Encountered unexpected errors:
			 failed to validate block received via sync (passed Error object: block height is 10, expected 11)
```
see:
https://circleci.com/gh/orbs-network/orbs-network-go/4121?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

ronnno:
and another one. this one triggered by `TestCreateGazillionTransactionsWhileTransportIsCorruptingRandomMessages`:

```
info 2018-11-21T16:55:25.388385Z transport message received node=a899b3 service=gossip entry-point=BlockSync request-id=BlockSync-1542819325387393758 header={ProtocolVersion:0,VirtualChainId:0,RecipientPublicKeys:[],RecipientMode:RECIPIENT_LIST_MODE_BROADCAST,Topic:(BlockSync)BLOCK_SYNC_AVAILABILITY_REQUEST,} function=gossip.(*service).OnTransportMessageReceived source=services/gossip/service.go:51 _test=acceptance _branch= _commit= _test-id=acceptance-TestCreateGazillionTransactionsWhileTransportIsCorruptingRandomMessages-1542819325-CONSENSUS_ALGO_TYPE_BENCHMARK_CONSENSUS
info 2018-11-21T16:55:25.388449Z transport message received node=a899b3 service=gossip entry-point=BlockSync request-id=BlockSync-1542819325387742041 header={ProtocolVersion:0,VirtualChainId:0,RecipientPublicKeys:[],RecipientMode:RECIPIENT_LIST_MODE_BROADCAST,Topic:(BlockSync)BLOCK_SYNC_AVAILABILITY_REQUEST,} function=gossip.(*service).OnTransportMessageReceived source=services/gossip/service.go:51 _test=acceptance _branch= _commit= _test-id=acceptance-TestCreateGazillionTransactionsWhileTransportIsCorruptingRandomMessages-1542819325-CONSENSUS_ALGO_TYPE_BENCHMARK_CONSENSUS
[...]

PR #517, opened 2018-11-21
Feature/intra node sync

ronnno:
Two go routines sync transaction pool and state storage services asynchronously consistently.

renamed block sync to internodesync and named intra node sync package servicesync


bootstrap/inmemory/network.go
```diff
@@ -120,12 +120,14 @@ func (n *Network) Size() int {
...
 	ch := make(chan *client.SendTransactionResponse)
 	supervised.GoOnce(n.Logger, func() {
+		defer close(ch)
 		publicApi := n.Nodes[nodeIndex].GetPublicApi()
 		output, err := publicApi.SendTransaction(ctx, &services.SendTransactionInput{
 			ClientRequest: (&client.SendTransactionRequestBuilder{SignedTransaction: tx}).Build(),
 		})
 		if err != nil {
 			panic(fmt.Sprintf("error sending transaction: %v", err)) // TODO: improve
+			return
```
golangcibot: unreachable code

bootstrap/inmemory/network.go
```diff
@@ -152,12 +154,14 @@ func (n *Network) CallMethod(ctx context.Context, tx *protocol.TransactionBuilde
...
 	ch := make(chan uint64)
 	supervised.GoOnce(n.Logger, func() {
+		defer close(ch)
 		publicApi := n.Nodes[nodeIndex].GetPublicApi()
 		output, err := publicApi.CallMethod(ctx, &services.CallMethodInput{
 			ClientRequest: (&client.CallMethodRequestBuilder{Transaction: tx}).Build(),
 		})
 		if err != nil {
 			panic(fmt.Sprintf("error calling method: %v", err)) // TODO: improve
+			return
```
golangcibot: unreachable code

ronnno:
CI flakiness tests fail on https://github.com/orbs-network/orbs-network-go/issues/516
This is extremely rare on dev machine but happens every time in CI

test/acceptance/block_sync_internal_test.go
```diff
@@ -0,0 +1,134 @@
...
+		// Resend an already committed transaction to Non-Leader
+		nonLeaderTxResponse, ok := <-network.SendTransaction(ctx, txBuilders[0].Builder(), 1)
+		require.True(t, ok)
+		require.Equal(t, protocol.TRANSACTION_STATUS_DUPLICATE_TRANSACTION_ALREADY_COMMITTED, nonLeaderTxResponse.TransactionStatus())
+	})
+}
+
+func waitForTransactionStatusCommitted(network harness.TestNetworkDriver, ctx context.Context, lastTx *protocol.Transaction, nodeIndex int) {
+	var txStatusOut *services.GetTransactionStatusOutput
+	for txStatusOut == nil || txStatusOut.ClientResponse.TransactionStatus() != protocol.TRANSACTION_STATUS_COMMITTED {
```
electricmonk: isn't this going to overwork the cpu?

test/acceptance/block_sync_internal_test.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func TestInternalBlockSync_StateStorage(t *testing.T) {
+
+	const transferAmount = 10
+	const transfers = 10
+	const totalAmount = transfers * transferAmount
+
+	var blockPairContainers []*protocol.BlockPairContainer
+	harness.Network(t).
```
electricmonk: I don't understand why you're starting two networks

test/acceptance/duplicate_tx_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"context"
+	"github.com/orbs-network/orbs-network-go/test/builders"
+	"github.com/orbs-network/orbs-network-go/test/harness"
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestSendSameTransactionMoreThanOnce1(t *testing.T) {
```
electricmonk: can you find better names than `1` and `2`?

test/acceptance/duplicate_tx_test.go
```diff
@@ -0,0 +1,85 @@
...
+		"error adding forwarded transaction to pending pool",
+		"error sending transaction",
+	).Start(func(ctx context.Context, network harness.TestNetworkDriver) {
+		ts := time.Now()
+
+		contract := network.GetBenchmarkTokenContract()
+		contract.DeployBenchmarkToken(ctx, 1)
+
+		// send three identical transactions to two nodes
+		network.SendTransactionInBackground(ctx, builders.TransferTransaction().WithTimestamp(ts).Builder(), 0)
```
electricmonk: why not assign the built tx to a variable and then send the variable? it'll better signify the fact that it's _the same_ transaction

test/acceptance/duplicate_tx_test.go
```diff
@@ -0,0 +1,85 @@
...
+		response0 := <- network.SendTransaction(ctx, builders.TransferTransaction().WithTimestamp(ts).Builder(), 1)
+		response1 := <- network.SendTransaction(ctx, builders.TransferTransaction().WithTimestamp(ts).Builder(), 1)
+
+		// second transaction will be accepted into the pool and committed by the internode sync
```
electricmonk: instead of comments, add messages to the assertion so that if they fail, the message appears in the log: "second transaction should be accepted into the pool and committed by the internode sync"

bootstrap/inmemory/network.go
```diff
@@ -124,13 +124,18 @@ func (n *Network) Size() int {
...
 		publicApi := n.Nodes[nodeIndex].GetPublicApi()
 		output, err := publicApi.SendTransaction(ctx, &services.SendTransactionInput{
 			ClientRequest: (&client.SendTransactionRequestBuilder{SignedTransaction: tx}).Build(),
 		})
 		if err != nil {
 			panic(fmt.Sprintf("error sending transaction: %v", err)) // TODO: improve
 		}
+
+		// wait for transaction to reach state storage
+		n.WaitForTransactionInState(ctx, output.ClientResponse.TransactionReceipt().Txhash())
```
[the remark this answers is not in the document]
ronnno: 1. there was no need for a channel before as well. the only case where calling code does not call this method or any of it's wrappers without a direct `<-` is in `requireSuccess` which uses a timer in a non idiomatic way and should be replaced with `context.WithTimeout`. So you are right - we should not be returning a channel
1. this line is problematic in another way - it is too aggressive. it makes tests run slower, and all thats really needed to make the semantics similar to before this fix is to use `WaitForTransactionInNodeState(... nodeIndex ...)` I added this here in my fight against flakiness and unfortunately I see it is not enough.

I will change it to be `WaitForTransactionInNodeState`. and if you would like I can remove the channel from this call but this is orthogonal to this PR

services/blockstorage/servicesync/service_sync_test.go
```diff
@@ -0,0 +1,149 @@
...
+			if committedBlockPair.TransactionsBlock.Header.BlockHeight() == currentHeight+1 {
+				currentHeight++
+			}
+			return currentHeight + 1, nil
+		}).Times(5)
+
+		// run sync loop
+		reportedHeight, err := syncOnce(ctx, sourceMock, committerMock, log.GetLogger())
+		require.NoError(t, err)
+		require.True(t, currentHeight == reportedHeight)
```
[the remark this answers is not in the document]
ronnno: done

services/transactionpool/get_transactions_for_ordering.go
```diff
@@ -10,10 +10,13 @@ import (
 
 func (s *service) GetTransactionsForOrdering(ctx context.Context, input *services.GetTransactionsForOrderingInput) (*services.GetTransactionsForOrderingOutput, error) {
 
+	//TODO fail if requested block height is in the past
+	s.logger.Info("GetTransactionsForOrdering called for block height", log.BlockHeight(input.BlockHeight))
```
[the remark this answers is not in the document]
ronnno: `s.logger.Info("GetTransactionsForOrdering called for block height", trace.LogFieldFrom(ctx), log.BlockHeight(input.BlockHeight))`

test/acceptance/block_sync_internal_test.go
ronnno: I can either use test.Eventually or add a  sleep(1*Millisecond) to the loop. Do you have a better idea how to wait for a transaction to reach TxPool?

test/acceptance/block_sync_internal_test.go
ronnno: To simulate a consistent set of blocks representing a blockchain with a history of contract deployment and some transactions.
Because we want to test the state by running CallMethod to check balances, the contract must be deployed on the target node. Deployment is represented in state so this test actually tests both contract deployment and state propagation through service sync.

test/acceptance/duplicate_tx_test.go
ronnno: done

test/acceptance/duplicate_tx_test.go
ronnno: because the interface accepts a builder and each builder can be used just once.

test/acceptance/duplicate_tx_test.go
ronnno: done

test/acceptance/block_sync_internal_test.go
ronnno: Used test.Eventually

PR #521, opened 2018-11-25
Gossip topics - refactoring and tests

danokoch:
- Moved the encoding/decoding logic to a separate folder (codec) 
- Added tests for the codec related actions on gossip topics


services/gossip/codec/testutil.go
```diff
@@ -0,0 +1,39 @@
+package codec
+
+import (
+	"reflect"
+)
+
+func emptyPayloads(num int) [][]byte {
```
electricmonk: this should be in a `builders.go` file

services/gossip/codec/testutil.go
```diff
@@ -0,0 +1,39 @@
...
+	res := [][]byte{}
+	for i := 0; i < num; i++ {
+		res = append(res, []byte{})
+	}
+	return res
+}
+
+func containsNil(obj interface{}) bool {
```
electricmonk: this should go to `test/testify_extensions.go` as `RequireDoesNotContainNil` or a better name

PR #580, opened 2018-12-04
Unit test: service-sync block committer triggered immediately

ronnno:
This new test (requested by @talkol) makes sure that no one breaks the service sync mechanism so that it runs slowly

However, there are some limitations to this test and we should think if it's desired:

- Tried to use `runtime.Gosched()` and measured how many `runtime.Gosched()`s are needed - it was between 1 and 2000. so we can reliably assume that a small amount of `runtime.Gosched()` will suffice
- Used a timer of 5 Milliseconds as the test requirement to trigger a commit. This may lead to flakiness. I've seen it fail with 1 Millisecond. And mostly it passes at around 500-700Microseconds (on my dev machine)
- Sync speed may be throttled by network once deployed across machines - making this test of little significance


netoneko:
Closing because it's stale.

PR #918, opened 2019-02-18
[DISCUSS F2F] Stress Tests in acceptance suite are now more lenient

electricmonk:
These tests will now require only 2f+1 nodes to have committed the tx set while tampering is taking place


netoneko:
Closing because it's stale.

PR #1003, opened 2019-03-03
[Experiment] Run GC during startup to lower memory consumption

netoneko:
This is an experimental branch, please do not merge.

Increases startup time but lowers the memory consumption (from 3Gb to 2Gb on 50k+ blocks with 1.7M keys in state storage).


netoneko:
Closing this pr, we can always come back to it if we hit any problems.

PR #1014, opened 2019-03-04
Updated go-ethereum to latest stable (1.9.6)

jlevison:
also includes some minor fixes around that upgrade and hashing


electricmonk:
@erankirsh any reason not to merge this?

jlevison:
no no, its unstable:
1. there is a newer release already, 1.8.27
2. there is a problem with a cache mechanism they added about 6 months ago, and we need to either fork their code, or stop using the simulator - it only affects the simulator
3. we should probably extract all our ethereum related code to an external repository to remove the dependency on go-ethereum
4. we should definitly cherry pick this commit on this PR: 8850d3324baeb572e7f2815b16295c284479e7b2

electricmonk:
> we should probably extract all our ethereum related code to an external repository to remove the dependency on go-ethereum
That is an unmaintainable fork, IMO

jlevison:
we should probably take this offline and think of the best course of action, but:
- configure tests to *not* run anything related to the ethereum simulator in flakiness tests (when -count >1)
- remove all usage of the ethereum simulator, move everything to ganache-cli

electricmonk:
Ethereum simulator is no more. @jlevison let's try to revive this after the holidays? we now only use `go-ethereum` as client code.

jlevison:
excellent!! i will take care of it on the week of the 27th

jlevison:
rewrote history to clean it up

at latest version, including the gomodule (both go-ethereum and leveldb needed the bump)
fixed all changes go-eth made that affected our code so it passes all test

crypto/hash/keccak256.go
```diff
@@ -7,14 +7,18 @@
...
 const (
 	KECCAK256_HASH_SIZE_BYTES = 32
 )
 
-func CalcKeccak256(data []byte) primitives.Keccak256 {
-	return crypto.Keccak256(data)
+func CalcKeccak256(data ...[]byte) primitives.Keccak256 {
+	d := sha3.NewLegacyKeccak256()
+	for _, b := range data {
+		d.Write(b)
```
golangcibot: Error return value of `d.Write` is not checked (from `errcheck`)

PR #1094, opened 2019-04-01
Refactor code that deploys into testnet

itamararjuan:
This PR adds code that deals with deploying into the CI testnet in a manner that follows our approach in recent months of having a centralized S3 bucket that holds information about which virtual chains Boyar should deploy.

The code downloads and modifies the CI's testnet S3 bucket central config and updates the version to deploy for all the chains available there and re-uploads that file into S3.

In addition, the code aims to check that the version was indeed deployed on all provided nodes and that the block height is progressing in all chains. then and only then will the step of deployment to testnet completes successfully and e2e tests can run safely against said chains.

The code also fixes a problem with parsing the nightly junit xml stats as well.


electricmonk:
@netoneko please review

.circleci/check-testnet-deployment.js
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env node
```
netoneko: We don't need this code. It can be solved by importing nebula and using metrics-related methods from there.

.circleci/deploy-testnet.sh
```diff
@@ -1,16 +1,27 @@
...
-
 export NVM_DIR="/opt/circleci/.nvm" && . $NVM_DIR/nvm.sh && nvm install v10.14.1 && nvm use v10.14.1
+export COMMIT_HASH=$(./docker/hash.sh)
 
-echo $TESTNET_SSH_PUBLIC_KEY > ~/.ssh/id_rsa.pub
+curl -O https://s3.eu-central-1.amazonaws.com/boyar-ci/boyar/config.json
+node .circleci/testnet-deploy-tag.js $COMMIT_HASH
 
-export COMMIT_HASH=$(./docker/hash.sh)
+aws s3 cp --acl public-read config.json s3://boyar-ci/boyar/config.json
```
netoneko: Please use nebula instead. We have all the API that is needed for that.

.circleci/testnet-deploy-tag.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
```
netoneko: Same about nebula.

test.flakiness.sh
```diff
@@ -18,6 +18,10 @@ if [[ "${LAST_COMMIT_MESSAGE}" == *"#extraflaky"* ]]; then
 fi
 
 if [[ $1 == "NIGHTLY" ]]; then
+    curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
```
netoneko: Please move to a separate pr. This is not related to the testnet.

.circleci/check-testnet-deployment.js
itamararjuan: After going through your metrics related code in Nebula  - I am not quite sure it's the same.
The method `waitForSync()` waits for 1 endpoint to sync. My code here is not testing for this.
Syncing can happen also with mismatched versions and that's not what I'm aiming for.
I'm not moving forward until I see the network is completely updated to the new version as a deployment should be.

I agree that we can replace my code for `eventuallyClosingBlocks()` and use your `waitForSync()`
What do you think ?

.circleci/deploy-testnet.sh
itamararjuan: The code in Nebula can generate a complete configuration and requires the complete set of network  JSON files. 
If I do that here - I need to copy or add a git submodule of testnet configuration in here (which is a private repo)

I was aiming to try and only change the version deployed to the chains in the file instead of changing the entire file. I can do that - let me know how do you want to solve sharing the JSONs required to run these APIs in nebula.

The `aws s3 cp` calls are still required as far as I can remember.

.circleci/testnet-deploy-tag.js
itamararjuan: This will be decided once you answer my comment regarding how to share the testnet configuration.
Perhaps we shall move the configuration of the CI testnet over to this repo and save us from the complexity of sharing a piece of a private repo.

test.flakiness.sh
itamararjuan: Moved already and also merged to master. Will be deleted from here.

itamararjuan:
I've addressed the concerns that @netoneko had with this PR.
When you are ready to review let me know so we can proceed with merging this PR to master

Preferrably before Thursday if possible :-)

.circleci/check-testnet-deployment.js
```diff
@@ -0,0 +1,99 @@
...
+
+const { waitUntilSync, getBlockHeight, getCommit } = require('@orbs-network/orbs-nebula/lib/metrics');
```
netoneko: We decided to unpublish `@orbs-network/orbs-nebula` a month ago. Why is it used here?

.circleci/check-testnet-deployment.js
```diff
@@ -0,0 +1,99 @@
...
+    process.exit(1);
+}
+
+async function getTopology() {
+    const result = await fetch(TOPOLOGY);
+    const topology = await result.json();
+    return topology;
+}
+
+async function eventuallyDeployed({ chainId, nodes }) {
```
netoneko: `waitUnitlCommit` from `metrics` does that already: https://github.com/orbs-network/nebula/blob/master/lib/metrics.js#L112

.circleci/check-testnet-deployment.js
```diff
@@ -0,0 +1,99 @@
...
+}
+
+async function eventuallyClosingBlocks({ chainId, nodes }) {
+    const firstEndpoint = `${nodes[0].ip}/vchains/${chainId}`;
+
+    // First let's get the current blockheight and wait for it to close another 50 blocks.
+    const currentBlockheight = await getBlockHeight(firstEndpoint);
+
+    try {
+        await waitUntilSync(firstEndpoint, currentBlockheight + 50);
```
netoneko: 50 is too long, blocks are closed at a rate of 9 seconds on an idle system. It would be 7 minutes after we waited for the network to start producing blocks, which itself can take 10-15 minutes.

.circleci/check-testnet-deployment.js
```diff
@@ -0,0 +1,99 @@
...
+(async () => {
+    const topology = await getTopology();
+    const nodes = topology.network;
+    const chains = topology.chains.map(chain => chain.Id);
+
+    const results = await Promise.all(chains.map((chainId) => eventuallyDeployed({ chainId, nodes })));
+    if (results.filter(r => r.ok === true).length === chains.length) {
+        console.log('New version deployed successfully on all chains in the testnet');
+    } else {
+        console.error('New version was not deployed on all nodes within the defined 5 minutes window, quiting..');
```
netoneko: 5 minutes is too low. It will never work that fast.

.circleci/check-testnet-deployment.js
```diff
@@ -0,0 +1,99 @@
...
+const TOPOLOGY = 'https://s3.us-west-2.amazonaws.com/boyar-testnet-bootstrap/boyar/config.json';
+const TARGET_HASH = process.argv[2];
+
+if (!TARGET_HASH) {
+    console.log('No target hash to check');
+    process.exit(1);
+}
+
+async function getTopology() {
+    const result = await fetch(TOPOLOGY);
```
netoneko: You have already downloaded it with curl, why do it here again? I would prefer modifying the local file.

.circleci/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "orbs-network-go-circleci",
+  "version": "1.0.0",
+  "description": "",
+  "main": "check-testnet-deployment.js",
+  "scripts": {},
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@orbs-network/orbs-nebula": "^1.1.2",
```
netoneko: Should be just `orbs-nebula`.

.circleci/check-testnet-deployment.js
itamararjuan: In your previous review you wanted me to use your code from Nebula
This is the official package. We talked about unpublish (and we did) the other package (`orbs-nebula`)

This scoped package is the official one which we left

.circleci/check-testnet-deployment.js
itamararjuan: Agreed - changing it.

.circleci/check-testnet-deployment.js
itamararjuan: I do not agree. I want to see all nodes have updated to the new version. This is not the same.
Your method is single endpoint centric

.circleci/check-testnet-deployment.js
itamararjuan: I'm updating it to `10`

.circleci/check-testnet-deployment.js
itamararjuan: I've updated it to 15 minutes

.circleci/check-testnet-deployment.js
netoneko: Official docs still use `npm install -g orbs-nebula`

Also maybe it would be easier to just install from our github master? We rarely release new nebula version.

.circleci/check-testnet-deployment.js
netoneko: You can run it multiple times waiting for the same commit. Same effect, less code.

.circleci/deploy-testnet.sh
netoneko: Ok, let's keep it that way for now.

.circleci/check-testnet-deployment.js
```diff
@@ -0,0 +1,91 @@
...
+        console.log('Sleeping for 30 seconds..');
+        await new Promise((r) => { setTimeout(r, 30 * 1000); });
+    } while (versionDeployed === false);
+
+    return {
+        ok: versionDeployed
+    };
+}
+
+async function eventuallyClosingBlocks({ chainId, nodes }) {
```
netoneko: Same can be done with nebula call for status, waiting for all of them to become green.

.circleci/check-testnet-deployment.js
netoneko: Cool.

PR #1099, opened 2019-04-02
Disable histograms to try and get rid of mem leaks in nightly

IdoZilberberg:
This is not a good thing - obviously we need histograms, but must check if this is the cause of mem leaks on nightly.
Histograms are the greatest memory hogger between tests, after logs (logs have been fixed in prev PR which was already merged).

### DIFF
--- a/instrumentation/metric/histogram.go
+++ b/instrumentation/metric/histogram.go
@@ -14,10 +14,39 @@ import (
 	"time"
 )
 
-type Histogram struct {
+const ENABLE_HISTOGRAMS = false
+
+type Histogram interface {
+	metric
+	RecordSince(t time.Time)
+	Record(measurement int64)
+}
+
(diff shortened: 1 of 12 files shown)


electricmonk:
I don't like the interface; why do you need polymorphism here? just keep everything internal to the `Histogram` struct and check the flag in the constructor AND record functions. This has a smaller footprint on the system.

netoneko:
How did you discover that the histograms are responsible for the leak? They are supposed to rotate all the time. They don't store all the measurements in memory.

IdoZilberberg:
@netoneko Pprof on local test runs:
image

IdoZilberberg:
Closing this because will try to resolve mem leak with distributing the nightly's high test count of 500 into 10 parallel processes of count=50 each.

PR #1105, opened 2019-04-04
Issue/tcp timeout

### LINKED ISSUE #973: TestDirectOutgoing_AdapterSendsUnicast - is flaky
https://circleci.com/gh/orbs-network/orbs-network-go/11616
```
Error Trace:	direct_outgoing_test.go:81
            				context.go:11
            				direct_outgoing_test.go:66
Error:      	Not equal: 
            	expected: []byte{0x2, 0x0, 0x0, 0x0, 0x1, 0x0, 0x0, 0x0, 0x11, 0x0, 0x0, 0x0, 0x2, 0x0, 0x0, 0x0, 0x22, 0x33, 0x0, 0x0}
            	actual  : []byte{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}
            	
            	Diff:
            	--- Expected
            	+++ Actual
            	@@ -1,4 +1,4 @@
            	 ([]uint8) (len=20) {
            	- 00000000  02 00 00 00 01 00 00 00  11 00 00 00 02 00 00 00  |................|
            	- 00000010  22 33 00 00                                       |"3..|
            	+ 00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
            	+ 00000010  00 00 00 00                                       |....|
            	 }
Test:       	TestDirectOutgoing_AdapterSendsUnicast
```

ronnno:
resolves #973

- fixed several minor issues causing flakiness in DirectTransport tests. see #973 

- added `GOSSIP_RECONNECT_INTERVAL` config value to guarantee that we never sleep for very long durations between reconnect attempts in tests which override `GOSSIP_CONNECTION_KEEP_ALIVE_INTERVAL` to very high values - since the second value was used to sleep between reconnect attempts.

- in `TestDirectOutgoing_ErrorDuringSendCausesReconnect` avoid possible race condition (theoretical - not observed) in case `Send()` queues a message just before `Close()` is called (while the queue is enabled) but the message remains in queue - then the connection is closed and immediately reopened, and only after the queue is re-enabled the message is sent by the clientLoop - the test would fail because we expect an empty keep alive but would receive a valid message. fixed by avoiding `Send` altogether - in any case we don't know if it happens before after or during the `Close`

### DIFF
--- a/config/system_factory_presets.go
+++ b/config/system_factory_presets.go
@@ -77,6 +77,7 @@ func defaultProductionConfig() mutableNodeConfig {
 
 	cfg.SetDuration(GOSSIP_CONNECTION_KEEP_ALIVE_INTERVAL, 1*time.Second)
 	cfg.SetDuration(GOSSIP_NETWORK_TIMEOUT, 30*time.Second)
+	cfg.SetDuration(GOSSIP_RECONNECT_INTERVAL, 5*time.Second)
 
 	// 10 minutes + 60 blocks is about 25 minutes
 	cfg.SetDuration(ETHEREUM_FINALITY_TIME_COMPONENT, 10*time.Minute)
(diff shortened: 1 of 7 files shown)


config/system_factory_presets.go
```diff
@@ -77,6 +77,7 @@ func defaultProductionConfig() mutableNodeConfig {
 
 	cfg.SetDuration(GOSSIP_CONNECTION_KEEP_ALIVE_INTERVAL, 1*time.Second)
 	cfg.SetDuration(GOSSIP_NETWORK_TIMEOUT, 30*time.Second)
+	cfg.SetDuration(GOSSIP_RECONNECT_INTERVAL, 5*time.Second)
```
netoneko: Why so long? I think 0.5 second should be fine here too.

ronnno: reducing to 1 second... don't want it too low in case several nodes are permanently offline or unreachable. then, reconnect attempt rate is multiplied by the number of offline nodes and it might become noisy.

PR #1114, opened 2019-04-14
Add more packages to sanitizer

netoneko:
The most important question is if we need any of this at all or we should simply proxy certain functions (essentially building our own library for strings, encoding, etc).

If we choose to go down that path, this pr should be closed.


electricmonk:
I know that @talkol has major objections for whitelisting _any_ golang library for fear of creating non-deterministic contracts.
I'm not sure that I agree with this, as we've clearly seen that we can create non-deterministic contracts without any library, but there's a decision that needs to be made here.

@erankirsh ?

netoneko:
This pr also bans goroutines, declaration of channels and writing to a channel in the contracts.

services/processor/native/sanitizer/test/config.go
```diff
@@ -13,6 +13,7 @@ func SanitizerConfigForTests() *sanitizer.SanitizerConfig {
 		ImportWhitelist: map[string]bool{
 			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1"`:       true,
 			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/state"`: true,
+			`"time"`: true,
```
electricmonk: why? do we want to allow contracts to sleep?

services/processor/native/sanitizer/test/declaration_test.go
```diff
@@ -0,0 +1,22 @@
...
+package test
+
+import (
+	"github.com/orbs-network/orbs-network-go/services/processor/native/sanitizer"
+	"github.com/orbs-network/orbs-network-go/services/processor/native/sanitizer/test/usecases"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestCodeWithChannelDeclaration(t *testing.T) {
```
electricmonk: Why is this test in this file and other tests are in `statements_test.go`?

services/processor/native/sanitizer/test/usecases/create_channel.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1"
+	"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/state"
+)
+
+var PUBLIC = sdk.Export(add)
+var SYSTEM = sdk.Export(_init)
+
+var COUNTER_KEY = []byte("count")
```
electricmonk: please remove any non-pertinent code from these files, otherwise people will not understand why it's there

services/processor/native/sanitizer/test/config.go
netoneko: No, but they might want to do date conversion and date-based checks.

electricmonk:
Can you add a reject for Sleeps?

Shai

services/processor/native/sanitizer/test/declaration_test.go
netoneko: Moved to statements.

services/processor/native/sanitizer/test/usecases/create_channel.go
netoneko: Done.

netoneko:
I have also introduced a blacklist for functions and banned named imports. Since we curate the list of libraries the names are not supposed to clash anyways.

services/processor/native/sanitizer/declarations.go
```diff
@@ -0,0 +1,53 @@
...
+				expr := node.(*ast.UnaryExpr)
+				if expr.Op == token.ARROW {
+					err = errors.New("sending to channels not allowed")
+					return false
+				}
+			case *ast.CallExpr:
+				expr := node.(*ast.CallExpr)
+				if expr, ok := expr.Fun.(*ast.SelectorExpr); ok {
+					pkg := fmt.Sprintf("%s", expr.X)
+					f := fmt.Sprintf("%s", expr.Sel)
```
golangcibot: S1025: should use String() instead of fmt.Sprintf (from `gosimple`)

netoneko: There is no String() method, so that's irrelevant anyway.

services/processor/native/sanitize_code.go
```diff
@@ -27,7 +34,40 @@ func (s *service) createSanitizer() *sanitizer.Sanitizer {
 			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/safemath"`: true,
 			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/service"`:  true,
 			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/state"`:    true,
+
+			// Text
+			`"fmt"`:           true,
```
electricmonk: wait wait wait
we can't whitelist anything that might result in non-deterministic transactions
specifically, @talkol has objected to whitelisting `fmt` for these reasons

Can you please explain next to each import why it's whitelisted? preferably, create a struct:
```
type whitelist struct {
   packageName string
   useCase string
```
then populate it with justifications for whitelisting stuff

services/processor/native/sanitizer/test/imports_test.go
```diff
@@ -20,3 +21,18 @@ func TestCodeWithInvalidImport(t *testing.T) {
...
+	require.Error(t, err)
+	require.Equal(t, `native code verification error: custom import names not allowed`, err.Error())
+	require.Empty(t, output)
+}
+
+func TestCodeWithJsonAndHex(t *testing.T) {
+	source := usecases.Serialization
+	output, err := sanitizer.NewSanitizer(native.SanitizerConfigForProduction()).Process(source)
+	require.NoError(t, err)
+	require.Equal(t, source, output, "valid file content should be altered by sanitizer")
```
electricmonk: should ==> shouldn't

services/processor/native/sanitize_code.go
erankirsh: @netoneko At this stage, since we do not support non deterministic execution, I suggest to remove fmt

netoneko: Done.

netoneko: Instead of struct I replaced `bool` with a string.

services/processor/native/sanitize_code.go
```diff
@@ -24,39 +24,40 @@ func (s *service) createSanitizer() *sanitizer.Sanitizer {
...
 			// SDK
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1"`:          true,
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/address"`:  true,
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/env"`:      true,
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/ethereum"`: true,
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/events"`:   true,
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/safemath"`: true,
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/service"`:  true,
-			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/state"`:    true,
+			`"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1"`:          "SDK",
```
electricmonk: this code runs upon deployment, right? so SLA isn't SUPER critical?
if so, I'd consider not using a map, but rather creating a collection of whitelisted libs, with support for glob patterns, so that we can do `github.com/orbs-network/orbs-contract-sdk/go/sdk/*`

it's ok to leave it like this now and open an issue, but keep in mind that right now, on every package added to the sdk we'll need to add it here.

services/processor/native/sanitize_code.go
```diff
@@ -24,39 +24,40 @@ func (s *service) createSanitizer() *sanitizer.Sanitizer {
...
-			`"strings"`:       true,
-			`"strconv"`:       true,
-			`"text/template"`: true,
+			`"strings"`:       "Text manipulation",
+			`"strconv"`:       "Text manipulation",
+			`"text/template"`: "Text manipulation",
 
 			// Time
-			`"time"`: true,
+			`"time"`: "Time manipulation",
```
electricmonk: use case please :)

electricmonk:
@netoneko approved but please see my comment re: maintainability of the whitelist

PR #1116, opened 2019-04-14
adding a new workflow that is about tagged artifacts

itamararjuan:
This PR is an attempt to try and fix the problem with CircleCI not buildling our GitHub tags
and introduces a new workflow which is using `filters` for all the related jobs which are required
to build and release a `gamma-server` Docker image based on a release tag.


netoneko:
Closed in favor of #1124

PR #1127, opened 2019-04-21
remove logging of metrics to logs on interval

ronnno:
mute logging of system metrics to system logs:

monitoring (dashboard) now extracts the metrics directly via http polling, and not by extracting the metrics from the logs.

To make the logs more readable - muting automatic metrics logging.


netoneko:
I think it also makes sense to remove the filter from `main.go` as well: https://github.com/orbs-network/orbs-network-go/blob/master/main.go#L57

ronnno:
right - removing all log filters

ronnno:
monitoring (dashboard) now extracts the metrics directly via http polling, and not by extracting the metrics from the logs.

PR #1128, opened 2019-04-21
Add default handler for http api

netoneko:


electricmonk:
@netoneko what is the purpose of this PR? why do we need a default handler?

netoneko:
@electricmonk because it looks weird that we don't have one. We use this endpoint in may places like SDK, and when people open it, they are going to be confused when it returns 404.

PR #1131, opened 2019-04-23
stop fake consensus goroutine when context done

### LINKED ISSUE #1121: still flaky - TestSyncPetitioner_Stress_SingleThreadedConsensusAlgoDoesNotDeadlock
https://circleci.com/gh/orbs-network/orbs-network-go/13999#tests/containers/4

ronnno:
potentially resolves #1121 
did not find the reason directly but found that the fake consensus go-routine keeps running and keeps the harness and mocks in memory after the test finishes.  it also causes cpu to be consumed.

While this does not explain the bug, since it's rarely reproduced, closing with this indirect fix.


ronnno:
did not find the reason directly but found that the fake consensus go-routine keeps running and keeps the harness and mocks in memory after the test finishes. it also causes cpu to be consumed.

While this does not explain the bug, since it's rarely reproduced, closing with this indirect fix.

PR #1136, opened 2019-04-24
Bugfix/multiple blocks closing on incoming tx

electricmonk:


test/acceptance/empty_block_time_test.go
```diff
@@ -0,0 +1,26 @@
...
+			contract := callcontract.NewContractClient(network)
+			time.Sleep(100 * time.Millisecond)
+			heightBeforeTx, _ := network.BlockPersistence(0).GetLastBlockHeight()
+
+			_, txHash := contract.Transfer(ctx, 0, 43, 5, 6)
+			network.WaitForTransactionInState(ctx, txHash)
+
+			heightAfterTx, _ := network.BlockPersistence(0).GetLastBlockHeight()
+
+			require.InDelta(t, uint64(heightAfterTx), uint64(heightBeforeTx), 1.0, "incoming transaction triggered closure of more than one block")
```
netoneko: I would not use `InDelta` here because it would pass if `heightAfterTx` was the same as `heightBeforeTx`.

I think that something like this would be a bit more clear:

`require.Equal(t, uint64(heightBeforeTx)+1, uint64(heightAfterTx), "incoming transaction triggered closure of more than one block")`

PR #1138, opened 2019-04-24
Feature/v106

noambergIL:
don't merge ... for oded rewivew only


ronnno:
Created against the wrong branch. this is a hotfix for v1.0.x

PR #1141, opened 2019-04-28
Removed random port from tcp transport tests

electricmonk:
- [x] Remove random port from `direct_harness.go`


electricmonk:
@talkol FYI - specifically regarding a change I made where the outgoing queues are now being created along with the client goroutine as opposed to before; you had code there that specifically created them before, but you didn't specify why

PR #1147, opened 2019-05-05
Fixing some V1 TODOs

electricmonk:
* Added benchmarks for ECDSA signing
* Unified `WaitUntilShutdown()` mechanism in Orbs Node main and Gamma main
* Added a mutex in TCP Transport construction to prevent concurrent map access issue (rare)
* Removed wrong usage of acceptance test config factory in non-acceptance tests

### DIFF
--- a/config/service_factory_presets.go
+++ b/config/service_factory_presets.go
@@ -107,6 +107,21 @@ func ForLeanHelixConsensusTests(keyPair *testKeys.TestEcdsaSecp256K1KeyPair) Lea
 	return cfg
 }
 
+func ForBenchmarkConsensusTests(keyPair *testKeys.TestEcdsaSecp256K1KeyPair, leaderKeyPair *testKeys.TestEcdsaSecp256K1KeyPair, validators map[string]ValidatorNode) NodeConfig {
+	cfg := emptyConfig()
+	cfg.SetGenesisValidatorNodes(validators)
+	cfg.SetBenchmarkConsensusConstantLeader(leaderKeyPair.NodeAddress())
+	cfg.SetActiveConsensusAlgo(consensus.CONSENSUS_ALGO_TYPE_BENCHMARK_CONSENSUS)
+	cfg.SetUint32(CONSENSUS_CONTEXT_MAXIMUM_TRANSACTIONS_IN_BLOCK, 1)
+	cfg.SetUint32(VIRTUAL_CHAIN_ID, 42)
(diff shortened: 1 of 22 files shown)


services/consensusalgo/benchmarkconsensus/test/harness.go
```diff
@@ -89,7 +75,7 @@ func newHarness(tb testing.TB, isLeader bool) *harness {
 		blockStorage:     blockStorage,
 		consensusContext: consensusContext,
 		reporting:        log,
-		config:           cfg.OverrideNodeSpecificValues(":8080", 0, nodeKeyPair.NodeAddress(), nodeKeyPair.PrivateKey(), ""),
```
ronnno: this leaves only one other instance where `OverrideNodeSpecificValues()` is called. 
It follows the only invocation of `config.ForE2E()` (in `e2e/network.go:85`).

Consider joining `OverrideNodeSpecificValues()` and  `ForE2E()`  into a single function (`ForE2E()`) and removing `OverrideNodeSpecificValues()` from the interface `OverridableConfig`

netoneko:
It feels to me that we're going a bit overboard with abstractions here. I suggest we consider unifying `DevelopmentNetwork` and creating a node, when running a production node will be the same as creating a network with a single node (and with a different set of adapters).

I feel that it might make our code a little cleaner. In the same time, we can get rid of e2e code that constructs network of "production" nodes too.

We already have 2 places where we create networks and one we create a single node. Maybe it would be just easier if the process was one and the same.

config/service_factory_presets.go
```diff
@@ -107,6 +107,21 @@ func ForLeanHelixConsensusTests(keyPair *testKeys.TestEcdsaSecp256K1KeyPair) Lea
 	return cfg
 }
 
+func ForBenchmarkConsensusTests(keyPair *testKeys.TestEcdsaSecp256K1KeyPair, leaderKeyPair *testKeys.TestEcdsaSecp256K1KeyPair, validators map[string]ValidatorNode) NodeConfig {
```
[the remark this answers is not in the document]
electricmonk: I think we moved them here because we didn't want to expose `mutableConfig` to the system itself

services/consensusalgo/benchmarkconsensus/test/harness.go
electricmonk: subsequent PR

electricmonk:
@netoneko I'm not sure I follow; the logic in `NewDevelopmentNetwork` is _significantly_ different than the one in `NewNode`. Or am I missing your point? 

Please open an issue with your suggested changes and we can discuss it separately from this PR.

PR #1152, opened 2019-05-06
flakiness fix - extend sleep duration for advisory lock test

ronnno:
Flakiness


ronnno:
The test runs two processes for x seconds expecting them both to attempt locking the blocks file simultaneously. 
A possible flakiness is that one process ends its sleep before the second one starts - in this case, both processes will succeed and the lock conflict will not be triggered.

since this is an integration test and not a fast test, the simplest solution is extending the sleep. this means the test will execute for at least 3 seconds.

PR #1159, opened 2019-05-12
enforce network type field in blocks file header

ronnno:
## Description

Validate `NETWORK_TYPE` configuration value matches the value found in blocks file header, on filesystem block storage adapter init. Fail loading if values do not match.

## Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)

## Checklist

## Further comments

Currently network type is not in use. production configurations use the reserved value 0.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -137,6 +137,7 @@ type FilesystemBlockPersistenceConfig interface {
 	BlockStorageFileSystemDataDir() string
 	BlockStorageFileSystemMaxBlockSizeInBytes() uint32
 	VirtualChainId() primitives.VirtualChainId
+	NetworkType() protocol.SignerNetworkType
 }
 
 type GossipTransportConfig interface {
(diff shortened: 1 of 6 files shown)


electricmonk:
WAIT
I’m under the assumption we decided to deprecate network type as a concept.

Shai

ronnno:
This is intended for cleanliness and consistency - and avoids the question of what will happen with this field in the future. as it currently exists, and since the header structure indicates it should have the configured value, and since the spec currently indicates that a vchain is identified by both network type and vchain id, it's a bug to ignore the value. 

if not to enforce it, the opposite alternative is to remove any reference to the config value in the file header: to rename the field to "reserved" and to override it with 0.

I prefer enforcing it because it will ensure that in the future if we replace this concept with something else we will address the questions of representation in the blocks file and of validation as part of a future refactor. 

@OdedWx which option do you prefer?

electricmonk:
I don’t think we actually set this value anywhrre

Shai

ronnno:
the value itself is left 0 "reserved" in all deployments, some tests use "testnet" (particularly - consensus algorithm tests i believe) and the test I've added in this PR increments the value to 1 to check we detect mismatches.

OdedWx:
I think that it makes more sense to compare it to teh configuration then read it and not do anything with.
The address.network_type should be deprecated. We may consider to add it to the header/transactions in future protocol revisions as a mean of convenience.

Issue #1161, opened 2019-05-12
Logging of block timestamp validation failure is confusing


IdoZilberberg:
**Describe the bug**

The log is difficult to understand. Instead of human-readable ISO8601 timestamps, it prints nanos since Epoch, and even those are printed in hex notation!

Example from logz.io:

=LH= H=454927 V=0 ID=718d43 LHMSG RECEIVED PREPREPARE IGNORE - err=ValidateBlockProposal failed ValidateTransactionsBlock, newBlockHeight=454927: currentTimestamp **159df14d1ef527d1** prevTimestamp **159df14bde5aed90** now 2019-05-12 13:05:23.111749211 +0000 UTC m=+5009.974342532 allowed jitter 30s, err=currentBlockTimestamp is later upperJitterLimit: currentBlockTimestamp=**1557666359674218449** upperJitterLimit=**1557666353111749211**: ErrInvalidBlockTimestamp

**Steps To Reproduce**

Write a test that intentionally fails validateTxTransactionsBlockTimestamp() and runs t.Log(err) on the resulting error.

**Expected behavior**

Print both unix nanos since epoch in decimal format, and ISO8601 format for human readability.

**Logs**

Above.

**Environment:**

Production vchain 1100000 (but every env will have this)

itamararjuan:
Me and @electricmonk took a look at this issue and found out that the generated primitive types from `protocol.mb.go` contain the wrong kind of variable type in their `String()` method

```
type TimestampNano uint64

func (x TimestampNano) String() string {
	return fmt.Sprintf("%x", uint64(x))
}
```

The `%x` should have been a `%s` to make sure that Nano Timestamps are displayed as ISO8601 standard time strings so they are more human readable and not `%x` which doesn't help when we try to debug timestamp problems (in block validations error messages for example)

Fixing this the right way involves re-generating the `protocol.mb.go` file using `membufc` after we make adjustments to the TimestampNano field in the membuffers project.

This seems to be the correct solution according to @electricmonk  but might have impact on other people/parts of the system.  This requires a second look from @erankirsh @OdedWx @talkol  maybe?

PR #1162, opened 2019-05-12
Feature/improve block timestamp validation logs tests

IdoZilberberg:
Fixes https://github.com/orbs-network/orbs-network-go/issues/1161

@bolshchikov do we find this acceptable to refer to an issue instead of adding details here? o/w it means duplicating all the info here.


bolshchikov:
@IdoZilberberg , yes

electricmonk:
@IdoZilberberg why isn't this merged? maybe review with @ronnno as part of getting him into LH context

PR #1164, opened 2019-05-13
Tech Debt - Block Storage tests

ronnno:
## Description

Addressing several `//TODO`s in block storage:
1. Structural block validations were moved to `ConsensusContext` service so no need for tests for these validations in BlockStorage
1. Removed redundant sanity checks in `GetBlockByTx`
1. Added contract tests for `GetTransactionReceipt`
1. Expanded component test for `GetTransactionReceipt`
1. Removed the need for confusing `.Reset()` calls on Block Storage test harness mock objects in several tests
1. Reinstated `TestBlockPersistenceContract_WriteOutOfOrderPast_FailsWhenBlockDifferent` as `TestBlockPersistenceContract_WriteOutOfOrder_PastBlocksIgnored`
1. Removed stale `//TODO`s

### DIFF
--- a/services/blockstorage/adapter/memory/memory_persistence.go
+++ b/services/blockstorage/adapter/memory/memory_persistence.go
@@ -119,7 +119,7 @@ func (bp *InMemoryBlockPersistence) GetBlockByTx(txHash primitives.Sha256, minBl
 	var candidateBlocks []*protocol.BlockPairContainer
 	for _, blockPair := range allBlocks {
 		bts := blockPair.TransactionsBlock.Header.Timestamp()
-		if maxBlockTs > bts && minBlockTs < bts {
+		if maxBlockTs >= bts && minBlockTs <= bts {
 			candidateBlocks = append(candidateBlocks, blockPair)
 		}
 	}
--- a/services/blockstorage/test/block_sync_server_test.go
+++ b/services/blockstorage/test/block_sync_server_test.go
@@ -25,7 +25,7 @@ func TestSourceRespondToAvailabilityRequests(t *testing.T) {
 		harness := newBlockStorageHarness(t).
 			withNodeAddress(sourceAddress).
 			withSyncBroadcast(1).
-			withValidateConsensusAlgosAtLeast(0).
+			expectValidateConsensusAlgos().
 			start(ctx)
 
 		harness.commitSomeBlocks(ctx, 3)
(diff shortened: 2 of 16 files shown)


ronnno:
Please note while reviewing that `blockstorage/test/harness.go` now exposes `*configForBlockStorageTests` rather than just the interface. this makes the code slightly simpler but slightly violates encapsulation. open for comments on that as I'm not sure this is optimal

services/blockstorage/adapter/memory/memory_persistence.go
```diff
@@ -119,7 +119,7 @@ func (bp *InMemoryBlockPersistence) GetBlockByTx(txHash primitives.Sha256, minBl
 	var candidateBlocks []*protocol.BlockPairContainer
 	for _, blockPair := range allBlocks {
 		bts := blockPair.TransactionsBlock.Header.Timestamp()
-		if maxBlockTs > bts && minBlockTs < bts {
+		if maxBlockTs >= bts && minBlockTs <= bts {
```
electricmonk: this changes behavior - why?

services/blockstorage/test/block_sync_server_test.go
```diff
@@ -22,7 +22,12 @@ import (
 func TestSourceRespondToAvailabilityRequests(t *testing.T) {
 	test.WithContext(func(ctx context.Context) {
 		sourceAddress := keys.EcdsaSecp256K1KeyPairForTests(4).NodeAddress()
-		harness := newBlockStorageHarness(t).withNodeAddress(sourceAddress).withSyncBroadcast(1).start(ctx)
+		harness := newBlockStorageHarness(t).
+			withNodeAddress(sourceAddress).
+			withSyncBroadcast(1).
+			withValidateConsensusAlgosAtLeast(0).
```
electricmonk: maybe I'm missing something but are you ever passing a non-zero values?
anyway, I'd create a special modifier for this use case, something like `ignoringHandleBlockConsensus()`

services/blockstorage/adapter/memory/memory_persistence.go
ronnno: not really. these timestamps include grace periods. which mean they are of excessive range on purpose. the next paragraph goes over all the candidate blocks selected here and looks inside them exhaustively so if we add two more blocks to the candidate list it will not change the result, assuming the transaction is not found exactly at the outer edge of the grace period.

that is why the condition never broke any test in the past. but to be more strict in interpreting the spec i think these conditions should be inclusive. this is a minor fix.

services/blockstorage/test/block_sync_server_test.go
ronnno: renamed to `expectValidateConsensusAlgos()`

PR #1165, opened 2019-05-14
Use our fork of ntp library

netoneko:
## Description

Uses our fork of NTP library; continues #1163

## Types of changes

## Checklist

## Further comments

None

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -52,9 +52,9 @@
 [submodule "vendor/github.com/orbs-network/scribe"]
 	path = vendor/github.com/orbs-network/scribe
 	url = https://github.com/orbs-network/scribe
-[submodule "vendor/github.com/beevik/ntp"]
-	path = vendor/github.com/beevik/ntp
-	url = https://github.com/beevik/ntp
 [submodule "vendor/golang.org/x/net"]
 	path = vendor/golang.org/x/net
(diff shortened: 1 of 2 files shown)


electricmonk:
Why did we fork this library?

netoneko:
We want to control all the source code for security reasons.

electricmonk:
Did we fork all other dependencies? I don’t think we actually went forward with this. Is there anything specific in this lib that scares you?

Shai

netoneko:
@electricmonk the only dependency is `golang.org/x/net`, and I believe we can trust it.

electricmonk:
Go-ethereum?

Shai

IdoZilberberg:
@netoneko The original ntp lib has BSD-2 license, not MIT. I think has no restrictions but before going forward with using a forked version of it you **must** consult legal to:
a) make sure using BSD-2 code does not prevent us from using MIT license, and in general does not create any constraints or obligations on us.
b) ask what to do about the license header we put in all the files - there is already the header of the original author.
@electricmonk

netoneko:
@IdoZilberberg forking the library does not mean changing its license, only hosting it in our github account. Let's consult legal if we can link our code with bsd2 libs. My understanding is that we could do that.

PR #1166, opened 2019-05-15
resolve issue 524

### LINKED ISSUE #524: BlockStorage validation logic duplicated in Adapter
`CommitBlock()` calls `validateBlockDoesNotExist` not under lock. This leads to a race condition which is currently resolved by repeating same checks inside the adapter logic under a mutex lock. 

This leaks service logic into the adapter. 

Two possible solutions:
1. pessimistic locking 1 - use a mutex lock around block validation and writing to adapter. 
2. pessimistic locking 2 - use a go routine and buffered channel to synchronize CommitBlock()
3. optimistic locking - perform block height validation not under mutex but carry the validated block height into the adapter call. adapter would then ignore calls if the validated block height is already stale.

while we should prefer optimistic locking for performance, in this case, it may lead us to more edge cases where detecting a FORK would complicate the code while the costs of validating block height under a lock should be very small.. @talkol what do you think?

ronnno:
## Description

1. resolves #524
2. more consistent fork detection behavior. previously were open to race conditions which could prevent fork detection, now resolved.
3. comply with spec and check for forks for any attempt to write a block which already exists in storage - and not just for forks in the top persisted block

## Further comments

`BlockPersistence.WriteNextBlock()` now also returns the top persisted block height, and does not return an error if a future block is attempted. it merely returns false on the `added` param. the calling service later examines the result of the execution and executes the spec:
1. If the block was not `added` to persistent storage and the proposed block height is higher than the expected block height, the service now returns the error according to the spec. the adapter remains free of service logic. it merely adds the 
2. An error returned from `BlockPersistence.WriteNextBlock()` is always an IO error - never a block ordering issues. these concerns are in the service only

### DIFF
--- a/services/blockstorage/adapter/memory/memory_persistence.go
+++ b/services/blockstorage/adapter/memory/memory_persistence.go
@@ -77,9 +77,9 @@ func (bp *InMemoryBlockPersistence) GetLastBlockHeight() (primitives.BlockHeight
 
 func (bp *InMemoryBlockPersistence) WriteNextBlock(blockPair *protocol.BlockPairContainer) (bool, primitives.BlockHeight, error) {
 
-	added, pHeight, err := bp.validateAndAddNextBlock(blockPair)
-	if err != nil || !added {
-		return added, pHeight, err
+	added, pHeight := bp.validateAndAddNextBlock(blockPair)
+	if !added {
+		return added, pHeight, nil
 	}
 
 	bp.metrics.size.Add(sizeOfBlock(blockPair))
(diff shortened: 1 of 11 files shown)


services/blockstorage/adapter/memory/memory_persistence.go
```diff
@@ -76,38 +75,31 @@ func (bp *InMemoryBlockPersistence) GetLastBlockHeight() (primitives.BlockHeight
...
 	}
 
 	bp.metrics.size.Add(sizeOfBlock(blockPair))
 
-	return added, nil
+	return added, pHeight, nil
 }
 
-func (bp *InMemoryBlockPersistence) validateAndAddNextBlock(blockPair *protocol.BlockPairContainer) (bool, error) {
+func (bp *InMemoryBlockPersistence) validateAndAddNextBlock(blockPair *protocol.BlockPairContainer) (bool, primitives.BlockHeight, error) {
```
electricmonk: this can never return an error, please change signature

services/blockstorage/commit_block.go
```diff
@@ -65,7 +60,43 @@ func (s *service) commitBlock(ctx context.Context, input *services.CommitBlockIn
 		})
 	}
 
-	logger.Info("committed a block", logfields.BlockHeight(txBlockHeader.BlockHeight()), log.Int("num-transactions", len(input.BlockPair.TransactionsBlock.SignedTransactions)))
+	logger.Info("committed a block", logfields.BlockHeight(proposedBlockHeight), log.Int("num-transactions", len(input.BlockPair.TransactionsBlock.SignedTransactions)))
 
 	return nil, nil
 }
+
+func (s *service) detectForks(proposedBlock *protocol.BlockPairContainer, logger log.Logger) error {
```
electricmonk: Consider adding a unit test for this logic as it's self contained

ronnno: it's already covered by component tests. but I've added unit tests now.

services/blockstorage/adapter/memory/memory_persistence.go
ronnno: fixed

PR #1168, opened 2019-05-19
Add TPS benchmark to acceptance test suite

netoneko:
## Description

Adds two benchmarks (one baseline and one with ability to override current config values) for processing 1000 simple transactions.

Takes `3s` to complete with LH and `1s` to complete with benchmark.

## Types of changes

## Checklist

- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

If anyone wants to dig around looking for a low hanging fruit in regards to performance it's probably the best place to look.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -106,6 +106,7 @@ type NodeConfig interface {
 type OverridableConfig interface {
 	NodeConfig
 	ForNode(nodeAddress primitives.NodeAddress, privateKey primitives.EcdsaSecp256K1PrivateKey) NodeConfig
+	MergeWithFileConfig(source string) (mutableNodeConfig, error)
 }
 
 type mutableNodeConfig interface {
@@ -121,7 +122,6 @@ type mutableNodeConfig interface {
 	SetNodePrivateKey(key primitives.EcdsaSecp256K1PrivateKey) mutableNodeConfig
 	SetBenchmarkConsensusConstantLeader(key primitives.NodeAddress) mutableNodeConfig
 	SetActiveConsensusAlgo(algoType consensus.ConsensusAlgoType) mutableNodeConfig
-	MergeWithFileConfig(source string) (mutableNodeConfig, error)
 	Clone() mutableNodeConfig
 }
 
--- a/test/acceptance/stress_benchmark_test.go
+++ b/test/acceptance/stress_benchmark_test.go
@@ -1,0 +1,47 @@
+package acceptance
+
+import (
+	"context"
+	"fmt"
+	"github.com/orbs-network/orbs-network-go/config"
+	"github.com/orbs-network/orbs-network-go/test/rand"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func BenchmarkHappyFlow1k(b *testing.B) {
+	numTransactions := 1000
+
+	rnd := rand.NewControlledRand(b)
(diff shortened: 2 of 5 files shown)


netoneko:
The benchmark now looks looks like this (first is block height, last is TPS rate in block storage):

```
finished sending  1000 transactions in 118.079256ms
finished waiting for 1000 transactions in 1.801969411s
stats {54 {738.197504 805.306367 872.415231 872.415231 872.415231 792.3639442285714 70} 319.77419354838713}
```

test/acceptance/stress_test.go
```diff
@@ -131,15 +133,23 @@ func sendTransfersAndAssertTotalBalance(ctx context.Context, network *NetworkHar
...
+
 	for i := 0; i < network.Size(); i++ {
 		actualSum := contract.GetBalance(ctx, i, toAddress)
 		require.EqualValuesf(t, expectedSum, actualSum, "recipient balance did not equal expected balance in node %d", i)
 
 		actualRemainder := contract.GetBalance(ctx, i, fromAddress)
 		require.EqualValuesf(t, benchmarktoken.TOTAL_SUPPLY-expectedSum, actualRemainder, "sender balance did not equal expected balance in node %d", i)
 	}
+
+	return
```
IdoZilberberg: I personally find this style less readable than the explicit form `return transferDuration, waitDuration` - had to look for the usage of `waitDuration` before I realized it was implicitly returned.

test/acceptance/stress_benchmark_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return e
+}
+
+func BenchmarkHappyFlow1kWithOverrides(b *testing.B) {
+	numTransactions := 1000
+
+	rnd := rand.NewControlledRand(b)
+	newHarness().WithConfigOverride(func(cfg config.OverridableConfig) config.OverridableConfig {
+		c, err := cfg.MergeWithFileConfig(`{
```
IdoZilberberg: Do we need a new config override mechanism here? is it not possible to use existing code?
@electricmonk please comment on this.

electricmonk:
I'm not sure that I like the coupling of the acceptance suite to the file config format; on the other hand, I don't have a better idea. @ronnno @erankirsh your opinion will be appreciated

netoneko:
@electricmonk can we move forward with this since it's been 2 days?

test/acceptance/stress_test.go
noambergIL: i also don't like the style of a test function returning time values since tomorrow we might want to add more like this ...

config/config.go
```diff
@@ -110,6 +110,7 @@ type NodeConfig interface {
 type OverridableConfig interface {
 	NodeConfig
 	ForNode(nodeAddress primitives.NodeAddress, privateKey primitives.EcdsaSecp256K1PrivateKey) NodeConfig
+	MergeWithFileConfig(source string) (mutableNodeConfig, error)
 }
```
noambergIL: there is something not so good having overridable returning a mutable, when mutable inherits from override ... looks bad ...

ronnno: I agree. its a smell. the problem here imho is that we have two interfaces with synonymous names: `Overridable` and `mutable`. whats the difference? only the public access...

but i think you're right... it should be consistent here..

PR #1173, opened 2019-05-21
Execute young blocks received on block sync

electricmonk:
## Description

The need arises to run an "audit node"; specifically in our use case, the Election contract prints debug logs to stdout and we'd like to access these logs, which is only possible by running a node (Orbs Audit node) and hooking into the Docker logs.

To implement this, I added a configurable threshold `INTER_NODE_SYNC_AUDIT_BLOCKS_YOUNGER_THAN` (of type `time.Duration`) - blocks younger than this value (system time, not last block time) will be executed on incoming sync so that their transaction executions can be logged to stdout.

## Types of changes

## Checklist

## Further comments

This is temporary; in the long term we hope to have a proper audit node design


electricmonk:
@gadcl if you'll recall, we discussed this -
if the log was below `validateBlockConsensus`, the test would need to fabricate a valid block with proof, which is a nearly impossible task without actually running LeanHelix. 

Re: `BlockProvider` - this is an interface that we implement for LeanHelix's API; for the sake of the algo service itself, from the Orbs side, I don't think it really matters where it resides

PR #1178, opened 2019-06-03
V1.0.7

noambergIL:
## Description

fix double delegation when first transfer was recorded then delegate (no new ones created)
fix to ignore double delegation that already exist
fix to remove too many metric logs


electricmonk:
@noambergIL does this need to be merged?

noambergIL:
yes, at the moment we have a "version" that is in production, and (unfortunately) we need to patch it, so its good to have a standard. there are two options have a major 1.0.x branch that any new patch starts from (and ends merged into) ... or have a chain so that the next patch (fix the rewards) will start from 1.0.7 ... currently we started the former approach ...

PR #1180, opened 2019-06-04
Shut down gamma with debug call

netoneko:
## Description

This PR introduces an API call that shuts down Gamma.

## Types of changes

## Checklist

- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

This is a requirement from @gilamran so he can continue his work on Playground. Essentially, in case of shutdown a hard reset of individual gamma server will happen, but it will be automatically restarted by docker.

### DIFF
--- a/bootstrap/gamma/server_handlers.go
+++ b/bootstrap/gamma/server_handlers.go
@@ -1,0 +1,21 @@
+package gamma
+
+import (
+	"net/http"
+	"time"
+)
+
+func (s *GammaServer) addGammaHandlers(router *http.ServeMux) {
+	router.HandleFunc("/debug/gamma/shutdown", s.shutdownHandler)
+}
+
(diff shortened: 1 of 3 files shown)


bootstrap/gamma/server_handlers.go
```diff
@@ -0,0 +1,21 @@
...
+	router.HandleFunc("/debug/gamma/shutdown", s.shutdownHandler)
+}
+
+func (s *GammaServer) shutdownHandler(writer http.ResponseWriter, request *http.Request) {
+	if request.Method != http.MethodPost {
+		writer.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	writer.Write([]byte(`{"status":"shutting down"}`))
```
golangcibot: Error return value of `writer.Write` is not checked (from `errcheck`)

PR #1183, opened 2019-06-06
Redial Ethereum connection on every call, fixes #1160

### LINKED ISSUE #1160: ethereum connector recovery from a connection failure
**Describe the bug**

The connection object (the client) is 'dialed if needed' only if the object is nil - there is no logic to refresh it if it fails during runtime or if the node changes (dns change) for example.

it is possible that a flow exists where it does recover, but it should be refactored (the dial logic) to be more robust.

**Steps To Reproduce**

1. run the system
2. kill the ethereum node
3. it will not try to reconnect properly

**Expected behavior**

an ability to recover / restart the connection, either manually or periodically.

**Logs**
n/a
**Screenshots**
n/a
**Environment:**
n/a (all)
**Additional context**
n/a

netoneko:
## Description

Fixes #1160 

## Types of changes

## Checklist

### DIFF
--- a/services/crosschainconnector/ethereum/adapter/rpc_connection.go
+++ b/services/crosschainconnector/ethereum/adapter/rpc_connection.go
@@ -13,18 +13,12 @@ import (
 	"github.com/orbs-network/scribe/log"
 	"github.com/pkg/errors"
 	"math/big"
-	"sync"
 )
 
 type EthereumRpcConnection struct {
 	connectorCommon
 
 	config ethereumAdapterConfig
-
-	mu struct {
-		sync.Mutex
-		client *ethclient.Client
-	}
 }
 
 func NewEthereumRpcConnection(config ethereumAdapterConfig, logger log.Logger) *EthereumRpcConnection {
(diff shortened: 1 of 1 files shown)


services/crosschainconnector/ethereum/adapter/rpc_connection.go
```diff
@@ -33,36 +27,17 @@ func NewEthereumRpcConnection(config ethereumAdapterConfig, logger log.Logger) *
 	}
 	rpc.logger = logger.WithTags(log.String("adapter", "ethereum"))
 	rpc.getContractCaller = func() (caller EthereumCaller, e error) {
-		return rpc.dialIfNeededAndReturnClient()
+		return rpc.dial()
```
electricmonk: Yeah, I was wondering why we were reusing the same connection. Good catch.

electricmonk:
@netoneko @jlevison I'm approving this - but:
Please make sure to add a benchmark (gobench) that measures call latency with cached vs recreated connections, to make sure that the redial operation doesn't add significant latency to the call. Ethereum calls might occur at a fast pace, so if there's significant latency incurred, we might wanna opt for a stabler cached connection solution.

netoneko:
@electricmonk this concern will only arise if someone will actually start using the system to communicate with Ethereum, which we haven't tried yet even in tests

electricmonk:
@netoneko we already do that from the Elections contract. Please add the benchmark now. Otherwise, no one will ever remember to even check the Ethereum connector later.

PR #1184, opened 2019-06-06
Upon PR creation: deploy new chain and E2E + stress against it

itamararjuan:
This PR is intended for a feature development in our CI pipeline for this project
Don't approve for now


.circleci/chain_on_pr.sh
```diff
@@ -0,0 +1,89 @@
...
+    node .circleci/check-testnet-deployment.js $PR_CHAIN_ID
+
+    echo "Running the E2E suite against the newly deployed isolated chain for this PR.."
+    export API_ENDPOINT=http://35.161.123.97/vchains/$PR_CHAIN_ID/ \
+        STRESS_TEST_NUMBER_OF_TRANSACTIONS=100 \
+        STRESS_TEST_FAILURE_RATE=20 \
+        STRESS_TEST_TARGET_TPS=200 \
+        STRESS_TEST=true
+
+    ./git-submodule-checkout.sh
```
netoneko: I would move it to a separate step in `config.yml` as usual. It is confusing to have it here.

.circleci/chain_on_pr.sh
```diff
@@ -0,0 +1,89 @@
...
+
+    echo "Finished vendoring.."
+
+    echo "Starting E2E tests against network ($PR_CHAIN_ID)"
+    export VCHAIN=$PR_CHAIN_ID
+    go test ./test/e2e/... -v
+
+    echo "Disabling the $PR_CHAIN_ID network..."
+
+    node .circleci/testnet-disable-chain.js $PR_CHAIN_ID
```
netoneko: I think any manipulation that is done to the config should be done to a new version that was just downloaded. Otherwise we will inevitably run into complications because of the race condition that we have here.

.circleci/check-testnet-deployment.js
```diff
@@ -51,8 +67,13 @@ async function eventuallyClosingBlocks({ chainId, nodes }) {
 }
 
 (async () => {
-    const nodes = topology.network;
-    const chains = topology.chains.map(chain => chain.Id);
+    const nodes = topology.network.filter(({ ip }) => ip !== '54.149.67.22');
```
netoneko: Why is there a special case for this specific IP? It should be configurable.

.circleci/config.yml
```diff
@@ -152,6 +160,10 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+
+      - deploy_new_chain_pr_only:
+          requires:
+            - tests
```
netoneko: I believe it requires e2e because only e2e publishes the docker image.

.circleci/deploy-testnet.sh
```diff
@@ -17,14 +17,16 @@ curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh |
 export NVM_DIR="/opt/circleci/.nvm" && . $NVM_DIR/nvm.sh && nvm install v10.14.1 && nvm use v10.14.1
 export COMMIT_HASH=$(./docker/hash.sh)
 
-curl -O https://s3.eu-central-1.amazonaws.com/boyar-ci/boyar/config.json
+cd .circleci && npm install @orbs-network/orbs-nebula && cd ..
+
+curl -O https://boyar-testnet-bootstrap.s3-us-west-2.amazonaws.com/boyar/config.json
```
netoneko: This should be configurable in all shell scripts and taken from the environment.

.circleci/deploy-testnet.sh
```diff
@@ -17,14 +17,16 @@ curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh |
...
 node .circleci/testnet-deploy-tag.js $COMMIT_HASH
 
-aws s3 cp --acl public-read config.json s3://boyar-ci/boyar/config.json
+aws s3 cp --acl public-read config.json s3://boyar-testnet-bootstrap/boyar/config.json
```
netoneko: Same here.

.circleci/testnet-deploy-new-chain-for-pr.js
```diff
@@ -0,0 +1,61 @@
...
+This script updates a local Boyar config.json file to provision a new chain
+for E2E purposes on an isolated blockchain network
+
+Usage: 
+
+$ testnet-deploy-new-chain-for-pr.js <GitHub PR Link>
+
+Examples:
+
+$ testnet-deploy-new-chain-for-pr.js https://github.com/orbs-network/orbs-network-go/pull/1184
```
netoneko: Please use `$CIRCLE_PR_NUMBER` instead.

.circleci/testnet-deploy-new-chain-for-pr.js
```diff
@@ -0,0 +1,61 @@
...
+    newChain.HttpPort = newChain.HttpPort + 2;
+    newChain.GossipPort = newChain.GossipPort + 2;
+
+    configuration.chains.push(newChain);
+}
+
+fs.writeFileSync(configFilePath, JSON.stringify(configuration, 2, 2));
+
+console.log(prNumber);
+process.exit(0);
```
netoneko: Why?

.circleci/testnet-remove-chain.js
```diff
@@ -0,0 +1,41 @@
...
+const configFilePath = path.join(process.cwd(), 'config.json');
+
+if (!targetChainId) {
+    console.log('No chainId given!');
+    process.exit(1);
+}
+
+const configuration = require(configFilePath);
+
+const chainIndex = configuration.chains.findIndex(chain => chain.Id === parseInt(targetChainId));
```
netoneko: You do this manipulation quite a lot, maybe it should be extracted to a separate file.

.circleci/chain_on_pr.sh
itamararjuan: The race condition is due to the fact we manage the chains in the shape of a file on S3. So you recommend downloading a new version before every modification from S3?

.circleci/chain_on_pr.sh
itamararjuan: Are you talking about running the vendoring script or running the tests ? or both?
My reason for not having separate steps in the `config.yml` is due to the fact the terminal state gets discarded between `-run:` lines. So the fact it's all in 1 script makes it easier for me to maintain the state of having NVM installed and the correct Node.js version used , same goes for AWS CLI in case it sets any environment variables.

In order to move these parts of running the tests to a separate line in the configuration I would need to break this script apart to what happens before the tests are run and what happens after it. I don't find it that important.

Can you elaborate why you think it's confusing?

.circleci/config.yml
itamararjuan: oh, that is strange. The tests are passing while I specify the commit based docker tag.
Could it be that Boyar deployed something else in it's absence after failing silently to find the correct image?

Can you double check that ?

itamararjuan: Usually the `build` or `tests` steps which produces the binary should finish with exporting a working and tested image. Why the E2E from all things?

.circleci/deploy-testnet.sh
itamararjuan: Agreed. Where would you put this URL ?

.circleci/testnet-deploy-new-chain-for-pr.js
itamararjuan: From my checks it's not available anymore. When I started developing I printed the `env` and looked at whatever is available to work with. the only thing was `$CI_PULL_REQUESTS` - no other variable was available in the environment to work with. You can check for yourself.

.circleci/testnet-deploy-new-chain-for-pr.js
itamararjuan: I provide the new chain ID  which was just added and use that in the main bash script.
What is your `why` aimed at?

.circleci/testnet-remove-chain.js
itamararjuan: I agree, I'll refactor the common logic to one file and use it where needed.

.circleci/chain_on_pr.sh
netoneko: I would like you to move this specific line: `./git-submodule-checkout.sh` (I am not sure we actually need to do this at all). Nothing seems to depend on these modules.

.circleci/chain_on_pr.sh
netoneko: Yes, exactly. Download the file before each modification.

.circleci/config.yml
netoneko: Because we need run e2e inside docker and we need to build the image anyway. After the image passes e2e, we can export it.

Tests run separately for quick feedback.

The explanation why your code still works is simple: boyar can't pull the image before it is built, so it retries and retries until it succeeds (meaning e2e finished and uploaded the image).

.circleci/deploy-testnet.sh
netoneko: You can pick an environment variable and inject it into circle ci configuration.

.circleci/testnet-deploy-new-chain-for-pr.js
netoneko: Why do you need to call `process.exit(0)` explicitly?

.circleci/testnet-deploy-new-chain-for-pr.js
netoneko: You are correct.

.circleci/chain_on_pr.sh
itamararjuan: Moved @netoneko

.circleci/check-testnet-deployment.js
itamararjuan: I preferably want to remove this filtering all together.
For how long is this node going to be out of commission?

netoneko: We can redeploy it tomorrow, doesn't really matter to me anymore.

PR #1196, opened 2019-06-23
added a skip for the block storage test on remote networks

itamararjuan:
Adding a skip on a test regarding block storage which is not suitable for running against remote networks where tampering of the block storage file isn't trivial

### DIFF
--- a/test/e2e/block_storage_init_test.go
+++ b/test/e2e/block_storage_init_test.go
@@ -9,6 +9,7 @@ package e2e
 import (
 	"github.com/orbs-network/orbs-network-go/test"
 	"github.com/stretchr/testify/require"
+	"os"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


test/e2e/block_storage_init_test.go
```diff
@@ -19,6 +20,12 @@ func TestInitialBlockHeight(t *testing.T) {
 		t.Skip("Skipping E2E tests in short mode")
 	}
 
+	// This test is useless against remote networks since we cannot tamper with their storage
+	// So for the time being we skip this test.
+	if runningAgainstRemoteEnv := len(os.Getenv("API_ENDPOINT")) > 0; runningAgainstRemoteEnv == true {
```
electricmonk: I think @netoneko already have a facade representing whether we're running against a local or remote environment; look for something like "should bootstrap"

itamararjuan: I saw that @electricmonk but the harness doesn't expose it anywhere from what I saw ?
I don't see anyone calling `getConfig` and it's also not capitalized so not meant to be public anyways? or am I missing something? 
The only thing I see the harness exposing is a client and not the config?

electricmonk: maybe now we have a use case for exposing it?

itamararjuan: Agreed and changed - you can take another look at the code.

PR #1197, opened 2019-06-23
Add dependency on contract-external-libraries-go

netoneko:
Following conversation with @OdedWx and @erankirsh:
- [x] struct serialization extracted to a separate repo: https://github.com/orbs-network/contract-external-libraries-go
- [x] added this repo as a dependency for **experimental** builds

### DIFF
--- a/docker/build/Dockerfile.export
+++ b/docker/build/Dockerfile.export
@@ -3,7 +3,6 @@ FROM golang:1.11.4-alpine
 RUN apk add --no-cache gcc musl-dev
 
 ADD ./vendor/github.com/orbs-network/orbs-contract-sdk/go/ /go/src/github.com/orbs-network/orbs-network-go/vendor/github.com/orbs-network/orbs-contract-sdk/go/
-ADD ./vendor/github.com/orbs-network/contract-external-libraries-go/ /go/src/github.com/orbs-network/orbs-network-go/vendor/github.com/orbs-network/contract-external-libraries-go/
 
 ADD ./_bin/orbs-node /opt/orbs/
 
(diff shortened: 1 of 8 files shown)


docker/build/Dockerfile.export
```diff
@@ -3,6 +3,7 @@ FROM golang:1.11.4-alpine
 RUN apk add --no-cache gcc musl-dev
 
 ADD ./vendor/github.com/orbs-network/orbs-contract-sdk/go/ /go/src/github.com/orbs-network/orbs-network-go/vendor/github.com/orbs-network/orbs-contract-sdk/go/
+ADD ./vendor/github.com/orbs-network/contract-external-libraries-go/ /go/src/github.com/orbs-network/orbs-network-go/vendor/github.com/orbs-network/contract-external-libraries-go/
```
electricmonk: should this be here AND in the experimental file?

docker/build/Dockerfile.gamma
```diff
@@ -3,6 +3,7 @@ FROM golang:1.11.4-alpine
 RUN apk add --no-cache gcc musl-dev
 
 ADD ./vendor/github.com/orbs-network/orbs-contract-sdk/go/ /go/src/github.com/orbs-network/orbs-network-go/vendor/github.com/orbs-network/orbs-contract-sdk/go/
+ADD ./vendor/github.com/orbs-network/contract-external-libraries-go/ /go/src/github.com/orbs-network/orbs-network-go/vendor/github.com/orbs-network/contract-external-libraries-go/
```
electricmonk: ditto

docker/build/Dockerfile.export
netoneko: Fixed.

OdedWx:
Looks good, 
Note that ClearStruct is equivalent to WriteStruct with an empty struct as provided to ClearStruct. Worth mentioning in MD.

erankirsh:
I hereby consent to adding the external libraries.
Note that Orbs Production network will NOT include these external libraries and it should only be used in Gamma for development reference. 
Once a library has been deemed necessary - Orbs network will implement it in the SDK to integrate it into production network.

PR #1199, opened 2019-06-24
add option to specify whether an environment is remote for e2e config

itamararjuan:
The previous solution was to condition based on `shouldBootstrap` which doesn't exactly means remote environment.

It is conditioned based on the specifying an endpoint which is not good enough for our case since the e2e which runs against Docker containers is also using this very same environment variable but that still doesn't make the network remote.

I'm introducing a new variable to explicitly state whether the environment we run the e2e against is remote as in NOT a Docker containers (running on the same machine) kind of network or an in process network.


electricmonk:
1) E2E fails weirdly
2) Where do you set this env variable?

PR #1208, opened 2019-06-30
fixed incorrect storage metrics

jlevison:
## Description

fixed the issue with the block storage metrics
@ronnno please verify its good now

## Types of changes

### DIFF
--- a/services/blockstorage/adapter/filesystem/file_system_persistence.go
+++ b/services/blockstorage/adapter/filesystem/file_system_persistence.go
@@ -26,14 +26,16 @@ import (
 )
 
 type metrics struct {
-	size *metric.Gauge
+	bytesWritten *metric.Gauge
+	sizeOnDisk   *metric.Gauge
 }
 
 const blocksFilename = "blocks"
 
 func newMetrics(m metric.Factory) *metrics {
 	return &metrics{
-		size: m.NewGauge("BlockStorage.FileSystemSize.Bytes"),
+		bytesWritten: m.NewGauge("BlockStorage.Write.Bytes"),
+		sizeOnDisk:   m.NewGauge("BlockStorage.FileSystemSize.Bytes"),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ronnno:
@IdoZilberberg  did you notice the initialization in `NewBlockPersistence`? I think this way is better since it does not rely on `writeNextBlock` in order to provide an absolute file size measurement.

electricmonk:
@ronnno can we merge this one?

services/blockstorage/adapter/filesystem/file_system_persistence.go
```diff
@@ -84,6 +86,12 @@ func NewBlockPersistence(ctx context.Context, conf config.FilesystemBlockPersist
 		codec:        codec,
 	}
 
+	if fi, err := file.Stat(); err != nil {
+		adapter.logger.Info("unable to read file size for metrics", log.Error(err))
```
electricmonk: @ronnno @jlevison do we want to return an error here? is this a valid case?

services/blockstorage/adapter/filesystem/file_system_persistence.go
```diff
@@ -26,14 +26,16 @@ import (
...
+	bytesWritten *metric.Gauge
+	sizeOnDisk   *metric.Gauge
 }
 
 const blocksFilename = "blocks"
 
 func newMetrics(m metric.Factory) *metrics {
 	return &metrics{
-		size: m.NewGauge("BlockStorage.FileSystemSize.Bytes"),
+		bytesWritten: m.NewGauge("BlockStorage.Write.Bytes"),
```
electricmonk: why do we need two metrics? what value do we gain from measuring bytes written since startup?

jlevison: perhaps see if there is a spike in storage, maybe a better metric for that would be avg bytes per block, last 1000 blocks

services/blockstorage/adapter/filesystem/file_system_persistence.go
jlevison: don't think so - i cannot really point on when this line will ever run: if we cannot access the file, we have bigger problems (it will crash before) - the only scenario where this fails is if we cannot access the file.

electricmonk: just for the sake of being uniform, I think it's better to return an error

PR #1211, opened 2019-07-03
Changed goroutine model of gossip server-side

electricmonk:
Before this change, `DirectTransport` had a goroutine per connection, and handling the message occurred on that goroutine, effectively blocking further communication from that peer. 

This PR creates a goroutine per gossip topic, writing from the connection goroutines to the topic goroutines via a buffered channel. This essentially serializes all messages from _all peers_ to a single goroutine, but frees the connection goroutines and creates QoS per topic. 

In addition, we create a one-off goroutine per Block Sync request, so that scanning blocks or reading chunks from disk will not block the Block Sync topic goroutine.

It is expected that the Lean Helix topic will not be blocked, as it will have a goroutine that deals with reading from the topic.


services/gossip/service.go
```diff
@@ -71,14 +89,48 @@ func (s *service) OnTransportMessageReceived(ctx context.Context, payloads [][]b
...
+}
+
+type gossipMessageDispatcher map[gossipmessages.HeaderTopic]chan gossipMessage
+
+// These channels are buffered because we don't want to assume that the topic consumers behave nicely
+// In fact, Block Sync should create a new one-off goroutine per "server request", Consensus should read messages immediately and store them in its own queue,
+// and Transaction Relay shouldn't block for long anyway.
+func makeMessageDispatcher() (d gossipMessageDispatcher) {
+	d = make(gossipMessageDispatcher)
+	d[gossipmessages.HEADER_TOPIC_TRANSACTION_RELAY] = make(chan gossipMessage, 10)
```
netoneko: I suggest making the buffer size configurable. We don't know if 10 is the optimal configuration.

services/gossip/service.go
```diff
@@ -71,14 +89,48 @@ func (s *service) OnTransportMessageReceived(ctx context.Context, payloads [][]b
...
+}
+
+func (d gossipMessageDispatcher) dispatch(logger log.Logger, header *gossipmessages.Header, payloads [][]byte) {
+	ch := d[header.Topic()]
+	if ch == nil {
+		logger.Error("no message channel for topic", log.Int("topic", int(header.Topic())))
+		return
+	}
+
+	ch <- gossipMessage{header: header, payloads: payloads} //TODO should the channel have *gossipMessage as type?
```
netoneko: Is there a possibility of a deadlock here if too many messages were sent into the buffered channel?

services/gossip/service.go
electricmonk: yagni, I think

services/gossip/service.go
electricmonk: I don't think so - worst case scenario is that the transport goroutine is blocked on the channel

noambergIL:
As far as i see it looks good.

electricmonk:
@gadcl what do you mean, "monitoring"? are you requesting that we add something?

gadcl:
> @gadcl what do you mean, "monitoring"? are you requesting that we add something?

we discussed (suggested by Eran) the option of adding a len(ch) vs. cap(ch) metric to monitor buffer usage.

services/gossip/dispatcher.go
```diff
@@ -0,0 +1,120 @@
...
+	case gossipmessages.HEADER_TOPIC_TRANSACTION_RELAY:
+		return d.transactionRelay, nil
+	case gossipmessages.HEADER_TOPIC_BLOCK_SYNC:
+		return d.blockSync, nil
+	case gossipmessages.HEADER_TOPIC_LEAN_HELIX:
+		return d.leanHelix, nil
+	case gossipmessages.HEADER_TOPIC_BENCHMARK_CONSENSUS:
+		return d.benchmarkConsensus, nil
+	default:
+		return nil, errors.Errorf("no message channel for topic", log.Int("topic", int(topic)))
```
golangcibot: printf: Errorf call has arguments but no formatting directives (from `govet`)

electricmonk:
Added manual draining of topics on shutdown, this seems to resolve the memory issue. Not ideal and probably not needed, but this should act as a workaround until we decide whether or not to retain the memory leak test

cc @gadcl @noambergIL @IdoZilberberg

PR #1215, opened 2019-07-07
Feature/eth block time

noambergIL:
add support for block time in ethereum connector.

### DIFF
--- a/services/crosschainconnector/ethereum/service.go
+++ b/services/crosschainconnector/ethereum/service.go
@@ -48,7 +48,7 @@ func NewEthereumCrosschainConnector(connection adapter.EthereumConnection, confi
 	return s
 }
 
-// TODO NOAM  remove this
+// TODO	https://github.com/orbs-network/orbs-network-go/issues/1214 find way to remove simulator and remove this ctor
 func NewEthereumCrosschainConnectorWithFakeTimeGetter(connection adapter.EthereumConnection, config config.EthereumCrosschainConnectorConfig, parent log.Logger, metrics metric.Factory) services.CrosschainConnector {
 	logger := parent.WithTags(LogTag)
 	blockTimeGetter := timestampfinder.NewFakeBlockTimeGetter(logger)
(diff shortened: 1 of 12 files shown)


electricmonk:
@jlevison can you please co-review this? you have a lot of domain expertise here

jlevison:
of course, but its a big pr with very little commits, it will take time, how urgent is this?

services/crosschainconnector/ethereum/service.go
```diff
@@ -155,19 +171,78 @@ func (s *service) EthereumGetTransactionLogs(ctx context.Context, input *service
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &services.EthereumGetBlockNumberByTimeOutput{
+		EthereumBlockNumber: uint64(blockNumberAndTime.BlockNumber),
+	}, nil
+}
+
+func (s *service) EthereumGetBlockTime(ctx context.Context, input *services.EthereumGetBlockTimeInput) (*services.EthereumGetBlockTimeOutput, error) {
```
electricmonk: is this the latest block time? latest final? let's change the naming with @OdedWx so that it's crystal clear what each variant returns

services/crosschainconnector/ethereum/test/harness.go
```diff
@@ -109,3 +109,25 @@ func (h *harness) packInputArgumentsForSampleStorage(method string, args []inter
 		return ethereum.ABIPackFunctionInputArguments(parsedABI, method, args)
 	}
 }
+
+//type mockFinality struct {
```
electricmonk: what's this?

services/crosschainconnector/ethereum/service.go
noambergIL: its clear.
if EthereumGetBlockNumber is : the newest block under finality then 
this means it's the time of the finality block

the only 100% clear name would be EthereumGetLatestBlockNumberUnderFinalit
and ... Time .... these names are too much.

PR #1220, opened 2019-07-09
Remote/env fix on e2e harness + fixing PR chain deployments

itamararjuan:


.circleci/chain_on_pr.sh
```diff
@@ -51,6 +51,7 @@ then
 
     echo "Running the E2E suite against the newly deployed isolated chain for this PR.."
     export API_ENDPOINT=http://$TESTNET_NODE_IP/vchains/$PR_CHAIN_ID/ \
+        REMOTE_ENV="true" \
```
netoneko: You don't have to set it here, `REMOTE_ENV` is the opposite of `shouldBootstrap` from `test/e2e/harness.go`.

You don't have to read it from the environment.

itamararjuan: Actually `shouldBootstrap` isn't exactly the same as it is identifying the Docker e2e as a remote environment whereas this is probably the ONLY place you don't want to skip this test (that sets up the pre-historic blocks data)

That's why I introduced this new variable and new logic inside the e2e harness.

.circleci/config.yml
```diff
@@ -137,7 +137,7 @@ jobs:
       BOOTSTRAP_URL: "https://boyar-testnet-bootstrap.s3-us-west-2.amazonaws.com/boyar/config.json"
       BOOTSTRAP_S3_URI: "s3://boyar-testnet-bootstrap/boyar/config.json"
       TESTNET_NODE_IP: "35.161.123.97"
-      TESTNET_IGNORE_IPS: "54.149.67.22"
+      TESTNET_IGNORE_IPS: ""
```
electricmonk: what's this?

.circleci/testnet-deploy-new-chain-for-pr.js
```diff
@@ -33,11 +33,12 @@ if (!targetTag) {
...
 // The namespace 100000 is PR chains teritory
 const prLinkParts = githubPRLink.split('/');
-const prNumber = parseInt(prLinkParts[prLinkParts.length - 1]) + 100000;
+const prNumber = parseInt(prLinkParts[prLinkParts.length - 1]);
+const chainNumber = prNumber + 100000;
 
 const configuration = require(configFilePath);
 
-const chainIndex = configuration.chains.findIndex(chain => chain.Id === prNumber);
+const chainIndex = configuration.chains.findIndex(chain => chain.Id === chainNumber);
```
electricmonk: is this the Boyar config? if so, I'd like to see this abstracted in some way that signals that this is Boyar related stuff; if and when Boyar starts to read virtual chain config from a contract, this code will need to change as well and I'd like it to be in one place (here, not in Boyar - but still - not all over the place)

.circleci/config.yml
itamararjuan: This IP is not relevant  - It was there because @netoneko  was testing something special on that node and it was not available for me. So I made an option to exclude certain nodes from the network so that my health check post deployment will work

.circleci/testnet-deploy-new-chain-for-pr.js
itamararjuan: If we get to a point where we handle deploying new chains through triggering a smart contract this entire code will vanish or change dramtically. Yes this code alerts the S3 file and re-uploads it adding a new chain to it. How would you like to emphasize it ?

electricmonk: just extract this to function with appropriate names

PR #1222, opened 2019-07-10
Allow gamma to modify next block time

electricmonk:
Added a new handler to gamma, `/debug/gamma/inc-time?seconds-to-add=n`, that adds a delta to the time of subsequent blocks.

This, on conjunction with Ganache's inc-time function, should allow us to not sleep in finality-sensitive integrative tests

### DIFF
--- a/bootstrap/gamma/e2e/harness.go
+++ b/bootstrap/gamma/e2e/harness.go
@@ -44,8 +44,9 @@ func waitForBlock(endpoint string, targetBlockHeight primitives.BlockHeight) fun
 func runGammaOnRandomPort(t testing.TB, overrideConfig string) string {
 	port := test.RandomPort()
 	endpoint := fmt.Sprintf("http://127.0.0.1:%d", port)
+	t.Log(t.Name(), "running Gamma at", endpoint)
 	gamma.RunMain(t, port, overrideConfig)
-	require.True(t, test.Eventually(WAIT_FOR_BLOCK_TIMEOUT, waitForBlock(endpoint, 1)))
+	require.True(t, test.Eventually(WAIT_FOR_BLOCK_TIMEOUT, waitForBlock(endpoint, 1)), "Gamma did not start ")
 
 	return endpoint
 }
--- a/bootstrap/gamma/e2e/harness.go
+++ b/bootstrap/gamma/e2e/harness.go
@@ -1,0 +1,67 @@
+package e2e
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/orbs-network/orbs-client-sdk-go/codec"
+	orbsClient "github.com/orbs-network/orbs-client-sdk-go/orbs"
+	"github.com/orbs-network/orbs-spec/types/go/primitives"
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
+	"github.com/stretchr/testify/require"
+	"io/ioutil"
+	"net/http"
+	"testing"
+	"time"
+)
+
+const WAIT_FOR_BLOCK_TIMEOUT = 10 * time.Second
+
+type metrics map[string]map[string]interface{}
+
+func waitForBlock(endpoint string, targetBlockHeight primitives.BlockHeight) func() bool {
+	return func() bool {
+		res, err := http.Get(endpoint + "/metrics")
+		if err != nil {
+			fmt.Println(err)
+			return false
+		}
+
+		readBytes, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			fmt.Println(err)
+			return false
+		}
+		m := make(metrics)
+		json.Unmarshal(readBytes, &m)
+
+		blockHeight := m["BlockStorage.BlockHeight"]["Value"].(float64)
+		return primitives.BlockHeight(blockHeight) >= targetBlockHeight
+	}
+}
+
+func sendTransaction(t testing.TB, orbs *orbsClient.OrbsClient, sender *orbsClient.OrbsAccount, contractName string, method string, args ...interface{}) *codec.SendTransactionResponse {
+
+	tx, _, err := orbs.CreateTransaction(sender.PublicKey, sender.PrivateKey, contractName, method, args...)
+	require.NoError(t, err, "failed creating tx %s.%s", contractName, method)
+	res, err := orbs.SendTransaction(tx)
+	require.NoError(t, err, "failed sending tx %s.%s", contractName, method)
+	require.EqualValues(t, codec.TRANSACTION_STATUS_COMMITTED.String(), res.TransactionStatus.String(), "transaction to %s.%s not committed", contractName, method)
+	require.EqualValues(t, codec.EXECUTION_RESULT_SUCCESS.String(), res.ExecutionResult.String(), "transaction to %s.%s not successful", contractName, method)
+
+	return res
+}
+
+func deployContract(t *testing.T, orbs *orbsClient.OrbsClient, account *orbsClient.OrbsAccount, contractName string, code []byte) {
+	sendTransaction(t, orbs, account, "_Deployments", "deployService", "LogCalculator", uint32(protocol.PROCESSOR_TYPE_NATIVE), []byte(code))
+}
+
+func sendQuery(t testing.TB, orbs *orbsClient.OrbsClient, sender *orbsClient.OrbsAccount, contractName string, method string, args ...interface{}) *codec.RunQueryResponse {
+	q, err := orbs.CreateQuery(sender.PublicKey, contractName, method, args...)
+	require.NoError(t, err, "failed creating query %s.%s", contractName, method)
+	res, err := orbs.SendQuery(q)
+	require.NoError(t, err, "failed sending query %s.%s", contractName, method)
+	require.EqualValues(t, codec.REQUEST_STATUS_COMPLETED.String(), res.RequestStatus.String(), "failed calling %s.%s", contractName, method)
+	require.EqualValues(t, codec.EXECUTION_RESULT_SUCCESS.String(), res.ExecutionResult.String(), "failed calling %s.%s", contractName, method)
+
+	return res
+}
--- a/bootstrap/gamma/e2e/harness.go
+++ b/bootstrap/gamma/e2e/harness.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"github.com/orbs-network/orbs-client-sdk-go/codec"
 	orbsClient "github.com/orbs-network/orbs-client-sdk-go/orbs"
+	"github.com/orbs-network/orbs-network-go/bootstrap/gamma"
+	"github.com/orbs-network/orbs-network-go/test"
 	"github.com/orbs-network/orbs-spec/types/go/primitives"
 	"github.com/orbs-network/orbs-spec/types/go/protocol"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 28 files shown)


services/transactionpool/init.go
```diff
@@ -71,6 +74,14 @@ func NewTransactionPool(ctx context.Context,
 	return s
 }
 
+func createClockIfNeeded(maybeClock adapter.Clock) adapter.Clock {
+	if maybeClock == nil {
+		return adapter.NewSystemClock()
```
ronnno: wouldn't it be more readable to use `adapter.NewSystemClock()` instead of `nil` in all the places you pass nil to maybeClock? this would remove the use of the notation for "optional" which is called maybe. this "maybe" convention is not very clear to me, why not optional - or better yet - just make it mandatory.  it would also remove the need for the cryptic "if needed" expression here..

bootstrap/gamma/gamma_e2e_test.go
```diff
@@ -104,3 +104,28 @@ func TestGammaWithEmptyBlocks(t *testing.T) {
...
+func TestGammaSetBlockTime(t *testing.T) {
+	randomPort := test.RandomPort()
+	runMain(t, randomPort, "")
+	endpoint := fmt.Sprintf("http://0.0.0.0:%d", randomPort)
+
+	require.True(t, test.Eventually(WAIT_FOR_BLOCK_TIMEOUT, waitForBlock(endpoint, 1)))
+
+	res, err := http.Post(endpoint+"/debug/gamma/inc-time?seconds-to-add=10", "text/plain", nil)
+	require.NoError(t, err, "failed incrementing next block time")
+	require.EqualValues(t, 200, res.StatusCode, "http call to increment time failed")
```
ronnno: I may prefer `REST/API/RPC` over `http` here - since it's the "how" and not the "what"...

bootstrap/gamma/gamma_e2e_test.go
```diff
@@ -104,3 +104,28 @@ func TestGammaWithEmptyBlocks(t *testing.T) {
...
+
+	sender, _ := orbsClient.CreateAccount()
+	transferTo, _ := orbsClient.CreateAccount()
+	client := orbsClient.NewClient(endpoint, 42, codec.NETWORK_TYPE_TEST_NET)
+
+	desiredTime := time.Now().Add(10 * time.Second)
```
ronnno: why 10 seconds? a major potential complication with this feature is clashing with logic in consensus algorithm which may rely on the system clock. if this feature is meant to be robust with lean-helix I'd run this test once for each consensus algorithm, and with clock deltas which are far larger then any config constant. for example, 10 minutes, or maybe even 45 minutes

electricmonk:
> I'm not sure I understand the motivation for this feature completely. If we want to just avoid waiting for "Finality" to be obtained in integration tests involving Ethereum and Orbs - to reduce feedback time - it may be much more robust and efficient to just create an API which overrides the finality feature (for instance, `setFinalityBlockNumber()` - or even just `disableEthereumFinality()` since gamma works with a shared ethereum source - there is no real need for the finality mechanism in gamma. it exists there only to more accurately simulate the orbs contract behavior.

Disabling finality isn't good enough - the purpose is to e2e test integrative contracts (like Elections), where skipping finality is kinda like shooting ourselves in the leg

 
> If the motivation is more broad - for instance, we want to be able to write tests which emulate time passing - then we cannot just override the TransactionPool's clock. we may want to simulate the creation of empty blocks during the time skipped. we may want to change the system clock so that all processes observe the time passage uniformly and not just transaction pool..

The txpool clock is the source for generating the proposed block time, and as such, it propagates (as far as I understand - I might be missing something here) throughout the system.

> In the bare minimum I'd limit the applicability of this feature to something which does not break lean helix (no more than jitter time config values), and does not support negative values, or even reducing the delta once it's set. and I'd also test the feature using the largest allowed values on gamma running lean helix (in e2e tests).

We definitely need to limit the feature, but first - we need to see that it actually achieves what it sets to achieve :)

bootstrap/gamma/gamma_e2e_test.go
electricmonk: can you please elaborate?

services/transactionpool/init.go
electricmonk: Maybe; the tradeoff here is that I need `AdjustableClock` to be shared by all nodes in a Gamma process, while `systemClock` can (and most likely _must_) be independently instantiated for each node, so I would need to find a way to do both. This is kinda like `nodeDependencyProvider` in `inmemory.Network`, where Gamma passes it as empty, indicating that `inmemory.NewNetworkWithNumOfNodes` should create the dependencies on its own. There might be a refactoring hiding here somewhere, I'm not sure yet.

ronnno:
> The txpool clock is the source for generating the proposed block time, and as such, it propagates ... throughout the system.

For smart contract code running under consensus this is true. but lean helix uses system clock to establish consensus. Let's discuss with @IdoZilberberg ?

But first, maybe we should limit the feature no non-lean-helix consensus. that may be a good solution. again, after verifying that we never use system clock anywhere in the codebase...

bootstrap/gamma/gamma_e2e_test.go
ronnno: it's a minor comment - the language draws the attention to the transport. but the transport did not necessarily failed here. the PRC/API did not go through - thats the point. we _know_ it by looking at the http code. but what it means is that the server failed to acknowledge the request. 

So possibly: `"failed incrementing next block time, with http code %v", res.StatusCode`

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		m := make(metrics)
+		json.Unmarshal(readBytes, &m)
+
+		blockHeight := m["BlockStorage.BlockHeight"]["Value"].(float64)
+		return primitives.BlockHeight(blockHeight) >= targetBlockHeight
+	}
+}
+
+func runGammaOnRandomPort(t testing.TB, overrideConfig string) string {
```
golangcibot: `runGammaOnRandomPort` is unused (from `deadcode`)

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+	require.NoError(t, err, "failed creating tx %s.%s", contractName, method)
+	res, err := orbs.SendTransaction(tx)
+	require.NoError(t, err, "failed sending tx %s.%s", contractName, method)
+	require.EqualValues(t, codec.TRANSACTION_STATUS_COMMITTED.String(), res.TransactionStatus.String(), "transaction to %s.%s not committed", contractName, method)
+	require.EqualValues(t, codec.EXECUTION_RESULT_SUCCESS.String(), res.ExecutionResult.String(), "transaction to %s.%s not successful", contractName, method)
+
+	return res
+}
+
+func deployContract(t *testing.T, orbs *orbsClient.OrbsClient, account *orbsClient.OrbsAccount, contractName string, code []byte) {
```
golangcibot: `deployContract` is unused (from `deadcode`)

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+	sendTransaction(t, orbs, account, "_Deployments", "deployService", "LogCalculator", uint32(protocol.PROCESSOR_TYPE_NATIVE), []byte(code))
+}
+
+func sendQuery(t testing.TB, orbs *orbsClient.OrbsClient, sender *orbsClient.OrbsAccount, contractName string, method string, args ...interface{}) *codec.RunQueryResponse {
```
golangcibot: `sendQuery` is unused (from `deadcode`)

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+	require.NoError(t, err, "failed creating query %s.%s", contractName, method)
+	res, err := orbs.SendQuery(q)
+	require.NoError(t, err, "failed sending query %s.%s", contractName, method)
+	require.EqualValues(t, codec.REQUEST_STATUS_COMPLETED.String(), res.RequestStatus.String(), "failed calling %s.%s", contractName, method)
+	require.EqualValues(t, codec.EXECUTION_RESULT_SUCCESS.String(), res.ExecutionResult.String(), "failed calling %s.%s", contractName, method)
+
+	return res
+}
+
+func timeTravel(t *testing.T, endpoint string, delta time.Duration) {
```
golangcibot: `timeTravel` is unused (from `deadcode`)

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+			return false
+		}
+
+		readBytes, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			fmt.Println(err)
+			return false
+		}
+		m := make(metrics)
+		json.Unmarshal(readBytes, &m)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/stretchr/testify/require"
+	"io/ioutil"
+	"net/http"
+	"testing"
+	"time"
+)
+
+const WAIT_FOR_BLOCK_TIMEOUT = 10 * time.Second
+
+type metrics map[string]map[string]interface{}
```
golangcibot: U1000: type `metrics` is unused (from `unused`)

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+
+func waitForBlock(endpoint string, targetBlockHeight primitives.BlockHeight) func() bool {
```
golangcibot: U1000: func `waitForBlock` is unused (from `unused`)

bootstrap/gamma/e2e/harness.go
```diff
@@ -0,0 +1,84 @@
...
+func runGammaOnRandomPort(t testing.TB, overrideConfig string) string {
+	port := test.RandomPort()
+	endpoint := fmt.Sprintf("http://127.0.0.1:%d", port)
+	gamma.RunMain(t, port, overrideConfig)
+	require.True(t, test.Eventually(WAIT_FOR_BLOCK_TIMEOUT, waitForBlock(endpoint, 1)))
+
+	return endpoint
+}
+
+func sendTransaction(t testing.TB, orbs *orbsClient.OrbsClient, sender *orbsClient.OrbsAccount, contractName string, method string, args ...interface{}) *codec.SendTransactionResponse {
```
golangcibot: U1000: func `sendTransaction` is unused (from `unused`)

services/transactionpool/init.go
ronnno: I see your point, it's valid in principle but there is also an argument for allowing `systemClock` to be shared, assuming it's a shallow wrapper around `time.Now()` by definition.. 

A clockBuilder would solve this issue somewhat verbosely (granted it's ugly but if you make it a generator function maybe the overhead can be minimized).

electricmonk:
@ronnno LeanHelix must not use system clock; the only relevant time to the consensus algo is the (proposed) block time, and it is set by the transaction pool on `GetTrasnactionsForOrdering`.

For now let's treat this as an undocumented feature of Gamma aimed at the core maintainer team only, and decide later if we want to document it or make sure it's more robust.

PR #1225, opened 2019-07-15
Support multiple files in contracts

netoneko:
## Description

- [x] Introduces ellipsis syntax to contracts (arrays of variable length, `func a(b int, c ...string)`
- [x] Allows `_Deployments.deployService` to receive multiple source code files as parameters in backwards compatible manner

## Types of changes

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/services/processor/native/adapter/fake/fake_compiler.go
+++ b/services/processor/native/adapter/fake/fake_compiler.go
@@ -16,7 +16,7 @@ import (
 
 type FakeCompiler interface {
 	adapter.Compiler
-	ProvideFakeContract(fakeContractInfo *sdkContext.ContractInfo, code string)
+	ProvideFakeContract(fakeContractInfo *sdkContext.ContractInfo, code ...string)
 }
 
 type fakeCompiler struct {
@@ -38,11 +38,11 @@ func (c *fakeCompiler) ProvideFakeContract(fakeContractInfo *sdkContext.Contract
 	c.provided[code] = fakeContractInfo
 }
 
-func (c *fakeCompiler) Compile(ctx context.Context, code string) (*sdkContext.ContractInfo, error) {
+func (c *fakeCompiler) Compile(ctx context.Context, code ...string) (*sdkContext.ContractInfo, error) {
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
-	contractInfo, found := c.provided[code]
+	contractInfo, found := c.provided[code[0]]
 	if !found {
 		return nil, errors.New("fake contract for given code was not previously provided with ProvideFakeContract()")
 	}
--- a/services/processor/native/adapter/fake/fake_compiler.go
+++ b/services/processor/native/adapter/fake/fake_compiler.go
@@ -16,6 +16,7 @@ import (
 
 type FakeCompiler interface {
 	adapter.Compiler
+	// Does not support multi-file fakes!
 	ProvideFakeContract(fakeContractInfo *sdkContext.ContractInfo, code ...string)
 }
 
@@ -31,11 +32,11 @@ func NewCompiler() *fakeCompiler {
 	}
 }
 
-func (c *fakeCompiler) ProvideFakeContract(fakeContractInfo *sdkContext.ContractInfo, code string) {
+func (c *fakeCompiler) ProvideFakeContract(fakeContractInfo *sdkContext.ContractInfo, code ...string) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
-	c.provided[code] = fakeContractInfo
+	c.provided[code[0]] = fakeContractInfo
 }
 
 func (c *fakeCompiler) Compile(ctx context.Context, code ...string) (*sdkContext.ContractInfo, error) {
--- a/services/processor/native/adapter/fake/fake_compiler.go
+++ b/services/processor/native/adapter/fake/fake_compiler.go
@@ -36,6 +36,10 @@ func (c *fakeCompiler) ProvideFakeContract(fakeContractInfo *sdkContext.Contract
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
+	if len(code) > 1 {
+		panic("fake compiler does not support multiple files in contracts")
+	}
+
 	c.provided[code[0]] = fakeContractInfo
 }
 
--- a/services/processor/native/adapter/native_compiler.go
+++ b/services/processor/native/adapter/native_compiler.go
@@ -11,6 +11,7 @@ package adapter
 import (
 	"context"
 	"encoding/hex"
+	"fmt"
 	sdkContext "github.com/orbs-network/orbs-contract-sdk/go/context"
 	"github.com/orbs-network/orbs-network-go/crypto/hash"
 	"github.com/orbs-network/orbs-network-go/instrumentation/metric"
@@ -81,7 +82,7 @@ func (c *nativeCompiler) warmUpCompilationCache() {
 	}
 }
 
-func (c *nativeCompiler) Compile(ctx context.Context, code string) (*sdkContext.ContractInfo, error) {
+func (c *nativeCompiler) Compile(ctx context.Context, code ...string) (*sdkContext.ContractInfo, error) {
 	c.metrics.sourceSize.Record(int64(len(code)))
 	start := time.Now()
 	defer c.metrics.totalCompileTime.RecordSince(start)
@@ -90,15 +91,18 @@ func (c *nativeCompiler) Compile(ctx context.Context, code string) (*sdkContext.
 	hashOfCode := getHashOfCode(code)
 
 	writeTime := time.Now()
-	sourceCodeFilePath, err := writeSourceCodeToDisk(hashOfCode, code, artifactsPath)
+	sourceCodeFilePaths, err := writeSourceCodeToDisk(hashOfCode, code, artifactsPath)
 	c.metrics.writeToDiskTime.RecordSince(writeTime)
-	defer os.Remove(sourceCodeFilePath)
+	for _, sourceCodeFilePath := range sourceCodeFilePaths {
+		defer os.Remove(sourceCodeFilePath)
+	}
+
 	if err != nil {
 		return nil, errors.Wrap(err, "could not write source code to disk")
 	}
 
 	buildTime := time.Now()
-	soFilePath, err := buildSharedObject(ctx, hashOfCode, sourceCodeFilePath, artifactsPath)
+	soFilePath, err := buildSharedObject(ctx, hashOfCode, sourceCodeFilePaths, artifactsPath)
 	c.metrics.buildTime.RecordSince(buildTime)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not build a shared object")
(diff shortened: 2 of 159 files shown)


services/processor/native/get_code.go
```diff
@@ -52,21 +52,26 @@ func (s *service) retrieveContractInfo(ctx context.Context, executionContextId p
...
 		return nil, err
 	}
 
-	code, err := s.sanitizeDeployedSourceCode(string(codeBytes))
-	if err != nil {
-		return nil, errors.Wrapf(err, "source code for contract '%s' failed security sandbox audit", contractName)
+	var code []string
+
+	for _, cb := range codeBytes {
+		sanitizedCode, err := s.sanitizeDeployedSourceCode(string(cb))
```
golangcibot: shadow: declaration of "err" shadows declaration at line 55 (from `govet`)

netoneko:
@OdedWx please take look at this part specifically:

https://github.com/orbs-network/orbs-network-go/pull/1225/files#diff-ceed5a0f600f4262522a39b49fcb4cb8

services/processor/native/adapter/native_compiler.go
```diff
@@ -90,15 +91,18 @@ func (c *nativeCompiler) Compile(ctx context.Context, code string) (*sdkContext.
...
 	writeTime := time.Now()
-	sourceCodeFilePath, err := writeSourceCodeToDisk(hashOfCode, code, artifactsPath)
+	sourceCodeFilePaths, err := writeSourceCodeToDisk(hashOfCode, code, artifactsPath)
 	c.metrics.writeToDiskTime.RecordSince(writeTime)
-	defer os.Remove(sourceCodeFilePath)
+	for _, sourceCodeFilePath := range sourceCodeFilePaths {
+		defer os.Remove(sourceCodeFilePath)
+	}
+
 	if err != nil {
```
noambergIL: are you sure this is best practice ? it looks ugly but i understand it allows some of them to fail. 
if os.remove fails no one gets any log ...

services/processor/native/adapter/native_compiler.go
```diff
@@ -81,7 +82,7 @@ func (c *nativeCompiler) warmUpCompilationCache() {
 	}
 }
 
-func (c *nativeCompiler) Compile(ctx context.Context, code string) (*sdkContext.ContractInfo, error) {
+func (c *nativeCompiler) Compile(ctx context.Context, code ...string) (*sdkContext.ContractInfo, error) {
 	c.metrics.sourceSize.Record(int64(len(code)))
```
noambergIL: question ... since you rely on order of files and then name them later with .i ... are we 100% all nodes read the files the same order ?

services/processor/native/call.go
```diff
@@ -98,22 +105,38 @@ func (s *service) prepareMethodInputArgsForCall(methodInstance types.MethodInsta
...
+				res = append(res, reflect.ValueOf(arg.Uint32Value()))
+			case reflect.Uint64:
+				res = append(res, reflect.ValueOf(arg.Uint64Value()))
+			case reflect.Slice:
+				if !arg.IsTypeBytesValue() {
+					return nil, errors.Errorf("method '%s' expects arg %d to be bytes but it has %s", functionNameForErrors, i, arg.StringType())
+				}
+				res = append(res, reflect.ValueOf(arg.BytesValue()))
+			default:
+				return nil, errors.Errorf("method '%s' expect arg %d to be of different type", functionNameForErrors, i)
```
noambergIL: how does this affect gamma-cli ? there was a limit to number of args there ...

services/processor/native/get_code.go
```diff
@@ -52,21 +52,26 @@ func (s *service) retrieveContractInfo(ctx context.Context, executionContextId p
 func (s *service) retrieveDeployedContractInfoFromState(ctx context.Context, executionContextId primitives.ExecutionContextId, contractName string) (*sdkContext.ContractInfo, error) {
 	start := time.Now()
 
-	codeBytes, err := s.callGetCodeOfDeploymentSystemContract(ctx, executionContextId, contractName)
+	codeBytes, err := s.getFullCodeOfDeploymentSystemContract(ctx, executionContextId, contractName)
```
noambergIL: i'd change the codeBytes var name to reflect its multi files

services/processor/native/get_code.go
```diff
@@ -90,7 +95,30 @@ func (s *service) retrieveDeployedContractInfoFromState(ctx context.Context, exe
 	return newContractInfo, nil
 }
 
-func (s *service) callGetCodeOfDeploymentSystemContract(ctx context.Context, executionContextId primitives.ExecutionContextId, contractName string) ([]byte, error) {
+func (s *service) getFullCodeOfDeploymentSystemContract(ctx context.Context, executionContextId primitives.ExecutionContextId, contractName string) ([][]byte, error) {
+	codeBytes, err := s.callGetCodeOfDeploymentSystemContract(ctx, executionContextId, contractName, 0)
```
noambergIL: why this separation ?

services/processor/native/adapter/native_compiler.go
netoneko: Since they use the same compiler, I think yes. The source code is the same in the end.

services/processor/native/adapter/native_compiler.go
netoneko: Even if some pollution happens, it's a temporary storage that we don't really care about.

services/processor/native/call.go
netoneko: I does not affect gamma-cli per se, support for this should be added to gamma-cli.

The only way to make it work is to construct a special transaction and gamma-cli does not know how to do it yet.

services/processor/native/get_code.go
netoneko: For backwards compatibility. So we can check that it worked for a singe case and then if we have more we will also fetch them.

netoneko: My goal here was to do small steps and not break the existing state of the `_Deployments` contract.

netoneko:
1. My main goal was to do the multifile but ellipsis allowed me to have backwards compatible API for that, so I followed this path. In the future I will separate different features into separate PRs.
1. I was not aware that gamma-cli has a cap on args, and I we will definitely change that.
1. I believe it can be solved by a dedicated `deploy` function in client sdk. After this work is merged, we can start developing it.

electricmonk:
@netoneko please split off ellipsis syntax to a separate PR

netoneko:
@electricmonk #1230

services/processor/native/get_code.go
netoneko: Fixed

services/processor/native/get_code.go
```diff
@@ -52,21 +52,26 @@ func (s *service) retrieveContractInfo(ctx context.Context, executionContextId p
...
 		return nil, err
 	}
 
-	code, err := s.sanitizeDeployedSourceCode(string(codeBytes))
-	if err != nil {
-		return nil, errors.Wrapf(err, "source code for contract '%s' failed security sandbox audit", contractName)
+	var code []string
+
+	for _, rawCodeFile := range rawCodeFiles {
+		sanitizedCode, err := s.sanitizeDeployedSourceCode(rawCodeFile)
```
golangcibot: shadow: declaration of "err" shadows declaration at line 55 (from `govet`)

services/processor/native/adapter/fake/fake_compiler.go
```diff
@@ -31,18 +32,18 @@ func NewCompiler() *fakeCompiler {
...
+	c.provided[code[0]] = fakeContractInfo
 }
 
-func (c *fakeCompiler) Compile(ctx context.Context, code string) (*sdkContext.ContractInfo, error) {
+func (c *fakeCompiler) Compile(ctx context.Context, code ...string) (*sdkContext.ContractInfo, error) {
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
-	contractInfo, found := c.provided[code]
+	contractInfo, found := c.provided[code[0]]
```
electricmonk: this may cause weird issues in tests, if someone attempts to fake compilation, for some odd reason, of a contract with more than one file. Please either make the fake support multiple files properly, or panic if attempting to fake a contract with more than a single file

services/processor/native/adapter/native_compiler.go
electricmonk: I'd like to see something slightly safer and more robust:
- hash each file separately, so that `getHashOfCode` returns `[]string`
- pass an array of structs to `writeSourceCodeToDisk` (something like `{Name: hash[i], Code: code[i[}`)
- since hash is unique for given code, you don't need to name using `.i`

WDYT?

services/processor/native/adapter/native_compiler_test.go
```diff
@@ -37,9 +37,12 @@ func TestCompileCodeWithExistingArtifacts(t *testing.T) {
 
 	t.Log("Build fresh artifacts")
 
-	sourceFilePath, err := writeSourceCodeToDisk("testPrefix", code, tmpDir)
+	sourceFilePath, err := writeSourceCodeToDisk("testPrefix", []string{code}, tmpDir)
```
electricmonk: please add a test for compilation of a contract with multiple files

services/processor/native/get_code.go
```diff
@@ -90,7 +95,30 @@ func (s *service) retrieveDeployedContractInfoFromState(ctx context.Context, exe
 	return newContractInfo, nil
 }
 
-func (s *service) callGetCodeOfDeploymentSystemContract(ctx context.Context, executionContextId primitives.ExecutionContextId, contractName string) ([]byte, error) {
+func (s *service) getFullCodeOfDeploymentSystemContract(ctx context.Context, executionContextId primitives.ExecutionContextId, contractName string) ([]string, error) {
+	code, err := s.callGetCodeOfDeploymentSystemContract(ctx, executionContextId, contractName, 0)
```
electricmonk: why do you treat the first part separately from the others?

services/processor/native/repository/_Deployments/contract_test.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, d := range diffs {
+		fmt.Println(string(d.Key), "=", string(d.Value))
+	}
+}
+
+func TestGetMultipleFilesCode(t *testing.T) {
+	diffs, _, _ := InSystemScope(nil, nil, func(m Mockery) {
+		m.MockServiceCallMethod("hello", "_init", nil)
+
+		deployService2("hello", 2, []byte("contract"), []byte("more contract stuff"))
```
electricmonk: what's `deployService2`?

services/processor/native/repository/_Deployments/deploy.go
```diff
@@ -27,15 +28,16 @@ func getInfo(serviceName string) uint32 {
...
+	res := _readCode(serviceName, index)
+	return res
+}
+
+func getCodeParts(serviceName string) uint32 {
+	return _codeCounter(serviceName)
 }
 
-func deployService(serviceName string, processorType uint32, code []byte) {
+func deployService(serviceName string, processorType uint32, code ...[]byte) {
```
electricmonk: did you verify with @OdedWx that this doesn't break protocol? the question here is whether a deployment transaction that was written against the deployment old contract (with no ellipsis) can be re-executed successfully against the new contract.

services/processor/native/repository/_Deployments/deploy.go
```diff
@@ -27,15 +28,16 @@ func getInfo(serviceName string) uint32 {
 	return processorType
 }
 
-func getCode(serviceName string) []byte {
```
electricmonk: this method is public, so any transaction that attempted to call `getCode(serviceName)` will not be executed properly on the new contract

services/processor/native/adapter/native_compiler.go
netoneko: There is no need for that. Compilation already happens in a separate directory, `file.1.go`, `file.2.go` and so on are good enough. File names do not affect compilation in any way.

services/processor/native/repository/_Deployments/contract_test.go
netoneko: FIxed.

services/processor/native/get_code.go
netoneko: As I have explained to to Noam: backwards compatibility. I can refactor it if you want.

services/processor/native/repository/_Deployments/deploy.go
netoneko: It's a read-only method that is only used by prism.

services/processor/native/adapter/native_compiler_test.go
netoneko: It is right here:

https://github.com/orbs-network/orbs-network-go/pull/1225/files/42b4077a127425d030c224324a31cc9a3432a8a5#diff-dca888d44e4077264d0e31da69b13522R27

Unfortunately, it does not run automatically. But it is covered by e2e which compiles both single file and multi file contracts.

services/processor/native/repository/_Deployments/deploy.go
netoneko: It will be re-executed successfully.

services/processor/native/repository/_Deployments/deploy.go
```diff
@@ -76,10 +81,29 @@ func _writeProcessor(serviceName string, processorType uint32) {
...
+}
+
+func _codeKey(serviceName string, index uint32) []byte {
+	if index == 0 { // backwards compatibility
+		return []byte(serviceName + ".Code")
+	}
+	return []byte(serviceName + ".Code." + strconv.FormatInt(int64(index), 10))
+}
+
+func _codeCounter(serviceName string) uint32 {
```
OdedWx: Let's keep code_counter semantics correct with 0 value implies 1 code part.

services/processor/native/repository/_Deployments/deploy.go
```diff
@@ -76,10 +81,29 @@ func _writeProcessor(serviceName string, processorType uint32) {
 	state.WriteUint32([]byte(serviceName+".Processor"), processorType)
 }
 
-func _readCode(serviceName string) []byte {
```
OdedWx: In order to avoid a synched Prism and orbs-network-go upgrade, it's better to keep the API backward compatible even if it returns a wrong result (only part of the code).
Let's add a new function getCodePart(index uint32), where getCode is identical to getCodePart(0)

OdedWx: From my side can merge after the fixes, verifying Prism is functional.

netoneko: I have verified that Prims did not break.

PR #1231, opened 2019-07-16
fix bug. get block by time dont use getfinality 

noambergIL:
## Description
fix issue with getblocknumberbytime -> it should not used getfinality to find block number.
used the getbytimestamp
then just verifies that the found block is past finality.


electricmonk:
why no test? :(

Issue #1232, opened 2019-07-16
Flakiness in direct transport: concurrent map read and map write


ronnno:
**Describe the bug**

**Steps To Reproduce**
flaky

**Logs**

```
info 2019-07-16T11:36:35.258994Z copying blocks file source=/Users/ido/dev/orbs/go/src/github.com/orbs-network/orbs-network-go/test/e2e/_data/blocks target=/tmp/orbs/e2e/a328846cd5b4979d68a8c58a9bdfeee657b34de7/blocks function=e2e.deployBlockStorageFiles source=test/e2e/file_utils.go:68 _test=e2e _branch= _commit= 
info 2019-07-16T11:36:35.280098Z built index node=a328846cd5b4979d68a8c58a9bdfeee657b34de7 adapter=block-storage valid-block-bytes=349720 block-height=538 function=filesystem.buildIndex source=services/blockstorage/adapter/filesystem/file_system_persistence.go:212 _test=e2e _branch= _commit= 
fatal error: concurrent map read and map write
info 2019-07-16T11:36:35.280353Z attempting outgoing transport connection node=a328846cd5b4979d68a8c58a9bdfeee657b34de7 adapter=gossip peer=127.0.0.1:46164 entry-point=Gossip.Transport.TCP.Client.127.0.0.1:46164 request-id=Gossip.Transport.TCP.Client.127.0.0.1:46164-1563276995280327000 function=tcp.(*DirectTransport).clientMainLoop source=services/gossip/adapter/tcp/direct_outgoing.go:24 _test=e2e _branch= _commit= 
goroutine 36 [running]:
runtime.throw(0x490e9b5, 0x21)
    /usr/local/Cellar/go/1.12.6/libexec/src/runtime/panic.go:617 +0x72 fp=0xc00018d698 sp=0xc00018d668 pc=0x402f712
runtime.mapaccess1_faststr(0x47e6560, 0xc00038b380, 0xc0001fd760, 0x14, 0xc0000c65f0)
    /usr/local/Cellar/go/1.12.6/libexec/src/runtime/map_faststr.go:21 +0x469 fp=0xc00018d708 sp=0xc00018d698 pc=0x40148e9
github.com/orbs-network/orbs-network-go/services/gossip/adapter/tcp.(*DirectTransport).connect.func1()
[...]

PR #1236, opened 2019-07-18
Refactor/Rename DirectTransport mutexes

ronnno:
## Description

A previous PR (#1232) expanded the scope of DirectTransport mutex to protect against concurrent access to the outgoing message queue by peer map (`outgoingPeerQueues`).

The current PR adds a dedicated mutex for controlling access to outgoing queues map, and re-organizes the items under each mutex in a separate struct type. the new types are `lockableOutgoingQueues` and `lockableTransportServer`.

In addition to some renaming I also removed the locking of the server mutex in some cases where I coulf not find a good reason for the lock. for instance, around calls to `connect()` (not renamed to `connectForever()` to expose the fact that it launches a listener using `GoForever()`)   

Points to notice in review:
1. Cases where I removed locking of `mutex` without replacing it with any other lock
1. `connectForever()`'s anonymous client-main-loop function is now not accessing the outgoing queues map anymore. This is done to avoid the need to obtain a lock inside the anonymous function every time it is launched. which means that if the queue object is ever changed the client loop will never pick up on it.

### DIFF
--- a/services/gossip/adapter/tcp/direct_harness.go
+++ b/services/gossip/adapter/tcp/direct_harness.go
@@ -60,7 +60,7 @@ func newDirectHarnessWithConnectedPeersWithTimeouts(t *testing.T, ctx context.Co
 	transport := makeTransport(ctx, t, cfg)                                               // step 3: create the transport; it will attempt to establish connections with the peer servers repeatedly until they start accepting connections
 	// end of section where order matters
 
-	peerTalkerConnection := establishPeerClient(t, transport.serverPort)           // establish connection from test to server port ( test harness ==> SUT )
+	peerTalkerConnection := establishPeerClient(t, transport.GetServerPort())      // establish connection from test to server port ( test harness ==> SUT )
 	peersListenersConnections := establishPeerServerConnections(t, peersListeners) // establish connection from transport clients to peer servers ( SUT ==> test harness)
 
 	h := &directHarness{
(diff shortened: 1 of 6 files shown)


services/gossip/adapter/tcp/direct_incoming.go
```diff
@@ -29,24 +29,27 @@ func (t *DirectTransport) serverListenForIncomingConnections(ctx context.Context
...
 	go func() {
 		<-ctx.Done()
-		t.mutex.Lock()
-		defer t.mutex.Unlock()
-		t.serverListeningUnderMutex = false
-		listener.Close()
+		t.server.Lock()
+		defer t.server.Unlock()
+		t.server.listening = false
+		err := listener.Close()
```
golangcibot: shadow: declaration of "err" shadows declaration at line 24 (from `govet`)

services/gossip/adapter/tcp/direct_transport.go
```diff
@@ -154,14 +178,30 @@ func (t *DirectTransport) Send(ctx context.Context, data *adapter.TransportData)
...
+	defer t.server.Unlock()
+
+	t.server.port = v
 }
 
-func (t *DirectTransport) safeLenOfOutgoingPeerQueues() int {
-	t.mutex.RLock()
-	defer t.mutex.RUnlock()
-	return len(t.outgoingPeerQueues)
+func (t *DirectTransport) allOutgoingQueuesEnabled() bool {
```
golangcibot: U1000: func `(*DirectTransport).allOutgoingQueuesEnabled` is unused (from `unused`)

services/gossip/adapter/tcp/direct_harness.go
```diff
@@ -60,7 +60,7 @@ func newDirectHarnessWithConnectedPeersWithTimeouts(t *testing.T, ctx context.Co
 	transport := makeTransport(ctx, t, cfg)                                               // step 3: create the transport; it will attempt to establish connections with the peer servers repeatedly until they start accepting connections
 	// end of section where order matters
 
-	peerTalkerConnection := establishPeerClient(t, transport.serverPort)           // establish connection from test to server port ( test harness ==> SUT )
+	peerTalkerConnection := establishPeerClient(t, transport.GetServerPort())      // establish connection from test to server port ( test harness ==> SUT )
```
electricmonk: why introduce a getter?

PR #1241, opened 2019-07-23
Added new metrics per outgoing TCP connection

electricmonk:
Due to a bug in the `MetricRegistry` we missed the fact that only a single connection's outgoing metrics where being reported - because of a name collision.

This PR adds a `panic` when attempting to register a metric with a name that already exists, as well as introducing new metrics per client:
`Gossip.OutgoingConnection.Queue.Usage.[NodeAddress].Percent`
`Gossip.OutgoingConnection.SendError.[NodeAddress].Count`
`Gossip.OutgoingConnection.EnqueueErrors.[NodeAddress].Count`

where `NodeAddress` stands for a hex-encoded representation of the Node's Orbs address

### DIFF
--- a/services/gossip/adapter/tcp/direct_outgoing.go
+++ b/services/gossip/adapter/tcp/direct_outgoing.go
@@ -57,7 +57,8 @@ func (t *DirectTransport) clientHandleOutgoingConnection(ctx context.Context, co
 			// meaning do a regular send (we have data)
 			err := t.sendTransportData(ctx, conn, data)
 			if err != nil {
-				t.metrics.outgoingConnectionSendErrors.Inc()
+				t.metrics.outgoingConnectionSendErrors.Inc() //TODO remove, replaced by following metric
+				queue.sendErrors.Inc()
 				t.logger.Info("failed sending transport data, reconnecting", log.Error(err), log.String("peer", queue.networkAddress), trace.LogFieldFrom(ctx))
 				conn.Close()
 				return true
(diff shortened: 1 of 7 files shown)


electricmonk:
Re 1 - yes, I don't want to remove it in case anyone uses it
Re 2 - you mean from the test harness? the cost is relatively low here IMO - if someone extracts and reuses it, the test will panic on re-registered metric

services/gossip/adapter/tcp/direct_outgoing.go
```diff
@@ -93,7 +94,8 @@ func (t *DirectTransport) clientHandleOutgoingConnection(ctx context.Context, co
 func (t *DirectTransport) addDataToOutgoingPeerQueue(data *adapter.TransportData, outgoingQueue *transportQueue) {
 	err := outgoingQueue.Push(data)
 	if err != nil {
-		t.metrics.outgoingConnectionSendQueueErrors.Inc()
+		t.metrics.outgoingConnectionSendQueueErrors.Inc() //TODO remove, replaced by following metric
```
IdoZilberberg: Remove when? is there an issue?

services/gossip/adapter/tcp/queue.go
```diff
@@ -26,17 +27,21 @@ type transportQueue struct {
...
+func NewTransportQueue(maxSizeBytes int, maxSizeMessages int, metricFactory metric.Factory, peerNodeAddress string) *transportQueue {
 	q := &transportQueue{
 		channel:     make(chan *adapter.TransportData, maxSizeMessages),
 		maxBytes:    maxSizeBytes,
 		maxMessages: maxSizeMessages,
 	}
 	q.protected.bytesLeft = maxSizeBytes
 
-	q.usagePercentageMetric = metricFactory.NewGauge("Gossip.OutgoingConnection.Queue.Usage.Percent")
+	q.usagePercentageMetric = metricFactory.NewGauge(fmt.Sprintf("Gossip.OutgoingConnection.Queue.Usage.%s.Percent", peerNodeAddress))
```
IdoZilberberg: BTW how long is the "peerNodeAddress"? is this 40 chars long?
Let's discuss this again before it goes to production. I'll want to see this on Grafana before production.

electricmonk: yeah, I think it's gonna be 40 chars long

services/gossip/adapter/tcp/direct_outgoing.go
electricmonk: backwards compatibility - I can't assume no one is using this metric

PR #1243, opened 2019-07-23
TCP Transport - client-related code moved to own struct

electricmonk:
This PR moves all code that deals with a single client connection within `DirectTransport` to a new struct, `clientConnection`. This is in preparation for a future PR to allow updating peer topology in runtime.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -233,6 +233,7 @@ type ValidatorNode interface {
 type GossipPeer interface {
 	GossipPort() int
 	GossipEndpoint() string
+	HexAddress() string
 }
 
 type HttpServerConfig interface {
--- a/config/config.go
+++ b/config/config.go
@@ -233,7 +233,7 @@ type ValidatorNode interface {
 type GossipPeer interface {
 	GossipPort() int
 	GossipEndpoint() string
-	HexAddress() string
+	HexOrbsAddress() string
 }
 
 type HttpServerConfig interface {
--- a/services/gossip/adapter/tcp/direct_outgoing.go
+++ b/services/gossip/adapter/tcp/direct_outgoing.go
@@ -39,19 +39,19 @@ type clientConnection struct {
 	sendQueueErrors *metric.Gauge
 }
 
-func newClientConnection(peerNodeAddress string, peer config.GossipPeer, parentLogger log.Logger, metricFactory metric.Registry, sharedMetrics *metrics, transportConfig config.GossipTransportConfig) *clientConnection {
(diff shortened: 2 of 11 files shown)


services/gossip/adapter/tcp/direct_transport.go
```diff
@@ -112,31 +111,26 @@ func newDirectTransportServer() *lockableTransportServer {
...
 func (t *DirectTransport) connectForever(bgCtx context.Context, peerNodeAddress string, peer config.GossipPeer) {
-	t.outgoingQueues.Lock()
-	defer t.outgoingQueues.Unlock()
+	t.clientConnections.Lock()
+	defer t.clientConnections.Unlock()
 
 	if peerNodeAddress != t.config.NodeAddress().KeyForMap() {
 
-		newQueue := NewTransportQueue(SEND_QUEUE_MAX_BYTES, SEND_QUEUE_MAX_MESSAGES, t.metricRegistry, peerNodeAddress)
+		client := newClientConnection(peer.HexAddress()[:6], peer, t.logger, t.metricRegistry, t.metrics, t.config)
```
ronnno: why do you enforce the length of HexAddress here?

services/gossip/adapter/tcp/direct_outgoing.go
```diff
@@ -10,56 +10,107 @@ import (
...
 		}
 	}
 }
 
 // returns true if should attempt reconnect on error
-func (t *DirectTransport) clientHandleOutgoingConnection(ctx context.Context, conn net.Conn, queue *transportQueue) bool {
-	t.logger.Info("successful outgoing gossip transport connection", log.String("peer", queue.networkAddress), trace.LogFieldFrom(ctx))
-	t.metrics.activeOutgoingConnections.Inc()
-	defer t.metrics.activeOutgoingConnections.Dec()
+func (c *clientConnection) clientHandleOutgoingConnection(ctx context.Context, conn net.Conn, queue *transportQueue) bool {
```
ronnno: suggest to extract some of the branching code here into functions - replace comments with function names to make this easier to read

config/config.go
```diff
@@ -233,6 +233,7 @@ type ValidatorNode interface {
 type GossipPeer interface {
 	GossipPort() int
 	GossipEndpoint() string
+	HexAddress() string
```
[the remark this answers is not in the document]
electricmonk: Orbs address. Id for logs mostly.

services/gossip/adapter/tcp/direct_transport.go
electricmonk: I do not enforce the length - I use only the first 6 chars

electricmonk:
good review; handled most of it in another commit
re connection object - I think it's actually beautiful that we don't need to reference it; it's cleaner, IMO, maybe more functional, that we don't "pollute the state" of the client with anything mutable.

PR #1261, opened 2019-08-06
Adding new cleanup logic to be run within Chain on PR workflow

itamararjuan:
This PR helps making sure the testnet remains clutter-free and removes stale chains which belonged to PRs which are already closed. 

The clean logic was added as a part of the normal Chain on PR workflow to make sure it never runs without the testnet being in it's best shape possible.


.circleci/testnet-cleanup-mark.js
```diff
@@ -0,0 +1,41 @@
...
+(async function () {
+    console.log('Querying closed github PRs...');
+    const closedPRs = await getClosedPullRequests();
+
+    let removeCounter = 0;
+    let updatedConfiguration = configuration;
+
+    for (let key in closedPRs) {
+        let aClosedPR = closedPRs[key];
+        let aClosedPRChainId = 100000 + aClosedPR.number;
```
electricmonk: I'm assuming this isn't the only place you're calculating chain id from PR number - you need to encapsulate it

itamararjuan: Refactored

PR #1265, opened 2019-08-07
(self) trigger transactions

noambergIL:
self generate trigger-transaction and validate in consensus context.
+some cleaning ups.

added first draft of contract for trigger and a first draft empty function for elections.


services/consensuscontext/create_transactions_block.go
```diff
@@ -48,10 +53,25 @@ func (s *service) createTransactionsBlock(ctx context.Context, input *services.R
...
 		BlockProof:         (&protocol.TransactionsBlockProofBuilder{}).Build(),
 	}
 
 	s.metrics.transactionsRate.Measure(int64(len(txBlock.SignedTransactions)))
 	return txBlock, nil
 }
+
+func (s *service) updateTransactions(txs []*protocol.SignedTransaction, blockTime primitives.TimestampNano) []*protocol.SignedTransaction {
+	if s.config.ConsensusContextTriggersEnabled() {
+		txs = append(txs, (&protocol.SignedTransactionBuilder{
```
electricmonk: This would be much nicer if you wrapped the trigger tx initialization in a function

services/consensuscontext/test/create_block_test.go
```diff
@@ -51,3 +51,23 @@ func TestCreateBlock_HappyFlow(t *testing.T) {
...
+
+func TestReturnAllAvailableTransactionsFromTransactionPool_WithTriggers(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		h := newHarness(t, true)
+		txCount := uint32(2)
+		txCountWithTrigger := txCount + 1
+
+		h.expectTxPoolToReturnXTransactions(txCount)
+
+		txBlock, err := h.requestTransactionsBlock(ctx)
```
electricmonk: Use require.NoError

services/consensuscontext/test/create_block_test.go
```diff
@@ -51,3 +51,23 @@ func TestCreateBlock_HappyFlow(t *testing.T) {
...
+		if err != nil {
+			t.Fatal("request transactions block failed:", err)
+		}
+		if uint32(len(txBlock.SignedTransactions)) != txCountWithTrigger {
```
electricmonk: Use require.Len

services/consensuscontext/test/validate_block_test.go
```diff
@@ -10,154 +10,120 @@ import (
...
+
+		vm.When("ProcessTransactionSet", mock.Any, mock.Any).Return(output, nil)
+
+		stateStorage := &services.MockStateStorage{}
+		stateHashOutput := &services.GetStateHashOutput{
+			StateMerkleRootHash: input.ResultsBlock.Header.PreExecutionStateMerkleRootHash(),
+		}
+		stateStorage.When("GetStateHash", mock.Any, mock.Any).Return(stateHashOutput, nil)
+
+		s := consensuscontext.NewConsensusContext(
```
electricmonk: Why no harness here?

services/consensuscontext/validate_transactions_block.go
```diff
@@ -86,27 +88,33 @@ func validateTransactionsBlockMetadataHash(ctx context.Context, vctx *txValidato
...
 }
 
-func validateTxTransactionOrdering(ctx context.Context, vctx *txValidatorContext) error {
+func (s *service) validateTxTransactionOrdering(ctx context.Context, transactionBlock *protocol.TransactionsBlockContainer) error {
+	txs := transactionBlock.SignedTransactions
+	if s.config.ConsensusContextTriggersEnabled() {
+		if len(txs) == 0 {
+			return ErrTriggerEnabledAndTriggerMissing
+		}
+		txs = txs[:len(txs)-1]
```
electricmonk: I might be missing it, but what if the last tx isn’t a trigger here?

services/consensuscontext/validate_transactions_block.go
```diff
@@ -161,3 +176,33 @@ func isValidBlockTimestamp(currentBlockTimestamp primitives.TimestampNano, prevB
...
+			if validateTransactionsBlockIsTxTrigger(tx) {
+				return ErrTriggerDisabledAndTriggerExists
+			}
+		}
+	}
+	return nil
+}
+
+func validateTransactionsBlockIsTxTrigger(tx *protocol.SignedTransaction) bool {
+	if tx.Transaction().ContractName().Equal(primitives.ContractName(triggers_systemcontract.CONTRACT_NAME)) &&
```
electricmonk: This could be a one liner

services/consensuscontext/validate_transactions_block_test.go
```diff
@@ -85,23 +83,167 @@ func TestTransactionsBlockValidators(t *testing.T) {
 		require.Equal(t, ErrMismatchedPrevBlockHash, errors.Cause(err), "validation should fail on incorrect prev block hash", err)
 	})
 
-	t.Run("should return error for transaction block with failing tx ordering validation", func(t *testing.T) {
```
electricmonk: this is not here anymore since you moved the ordering validation to a separate step, right?

services/consensuscontext/validate_transactions_block_test.go
```diff
@@ -85,23 +83,167 @@ func TestTransactionsBlockValidators(t *testing.T) {
...
 
+func TestConsensusContextValidateTransactionsBlockTriggerTransactionNotForwardedToPreOrder(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		block := builders.BlockPairBuilder().Build()
+		cfg := config.ForConsensusContextTests(nil, true)
+		txPool := &services.MockTransactionPool{}
+		txPool.When("ValidateTransactionsForOrdering", mock.Any, mock.Any).Call(func(ctx context.Context, input *services.ValidateTransactionsForOrderingInput) {
+			require.True(t, len(input.SignedTransactions) == len(block.TransactionsBlock.SignedTransactions)-1)
+		}).Return(nil, nil)
+		s := &service{txPool, nil, nil, cfg, nil, nil}
```
electricmonk: this is dangerous, as you might be failing to initialize stuff that the service assumes is initialized in the `NewConcensusContext` factory func. 

why are you not using the harness? why are you initializing the service only partially? seems to me that `validateTxTransactionOrdering` needs to have a unit test without the service at all

services/consensuscontext/validate_transactions_block_test.go
```diff
@@ -85,23 +83,167 @@ func TestTransactionsBlockValidators(t *testing.T) {
...
 }
 
-func TestIsValidBlockTimestamp(t *testing.T) {
+func TestConsensusContextValidateTransactionsBlockTriggerDisabledTransactionNotRemovedForForwardedToPreOrder(t *testing.T) {
+	test.WithContext(func(ctx context.Context) {
+		cfg := config.ForConsensusContextTests(nil, false)
+		block := builders.BlockPairBuilder().WithCfg(cfg).Build()
+		txPool := &services.MockTransactionPool{}
+		txPool.When("ValidateTransactionsForOrdering", mock.Any, mock.Any).Call(func(ctx context.Context, input *services.ValidateTransactionsForOrderingInput) {
+			require.True(t, len(input.SignedTransactions) == len(block.TransactionsBlock.SignedTransactions))
```
electricmonk: use `require.Len`

services/consensuscontext/validate_transactions_block_test.go
```diff
@@ -85,23 +83,167 @@ func TestTransactionsBlockValidators(t *testing.T) {
...
+		err := s.validateTxTransactionOrdering(ctx, block.TransactionsBlock)
+		require.Error(t, err)
+
+		ok, err := txPool.Verify()
+		require.True(t, ok)
+		require.NoError(t, err)
+	})
+}
+
+func TestConsensusContextValidateTransactionsBlockTriggerCompliance(t *testing.T) {
```
electricmonk: very nice

test/builders/blocks.go
```diff
@@ -281,3 +290,77 @@ func (c *corruptBlockPair) WithEmptyResultsBlock() *corruptBlockPair {
...
+
+	block.ResultsBlock.Header.MutateTransactionsBlockHashPtr(txBlockHashPtr)
+	block.ResultsBlock.Header.MutateReceiptsMerkleRootHash(receiptMerkleRoot)
+	block.ResultsBlock.Header.MutateStateDiffHash(stateDiffHash)
+	block.ResultsBlock.Header.MutatePreExecutionStateMerkleRootHash(preExecutionRootHash.StateMerkleRootHash)
+
+	return block
+}
+
+func (b *blockPairBuilder) WithCfg(cfg config.ConsensusContextConfig) *blockPairBuilder {
```
electricmonk: please use a locally-defined interface here to avoid coupling to `ConsensusContextConfig`

test/builders/transactions.go
```diff
@@ -152,6 +153,18 @@ func TransactionInputArgumentsParse(t *protocol.Transaction) *protocol.ArgumentA
 	return argsArray.ArgumentsIterator()
 }
 
+func TriggerTransaction(pv primitives.ProtocolVersion, vid primitives.VirtualChainId) *protocol.SignedTransaction {
```
electricmonk: maybe this shouldn't be strictly a test builder? do you build the actual trigger tx differently?

services/consensuscontext/validate_transactions_block.go
noambergIL: order of validations ... before the ordering we validate trigger compliance. so if it passed then the last one is a trigger tx.

services/consensuscontext/validate_transactions_block_test.go
noambergIL: yes

test/builders/transactions.go
noambergIL: code changed and it is now regular txbuilder with extra.

services/consensuscontext/validate_transactions_block_test.go
noambergIL: changed to equals as i'm comparing two lens.

services/consensuscontext/test/validate_block_test.go
noambergIL: only copied code so didn't really look inside. however changed to the harness

PR #1266, opened 2019-08-07
Keep gossip dispatcher thread responsive by allowing topic buffers to overflow

ronnno:
Dispatcher thread is blocked when waiting for topic buffer to empty out,
this leads to deadlocks when topic A consumer thread is waiting for a gossip message to arrive on another topic B. the dispatcher is not dispatching because it's waiting for consumer of topic A to read the next message and clear the buffer, but this will never happen because topic B is now frozen since the dispatcher is deadlocked.


services/gossip/dispatcher.go
```diff
@@ -78,14 +85,15 @@ func (c *meteredTopicChannel) flushChannelAfterContextDone(expiredContext contex
 }
 
 func newMeteredTopicChannel(name string, registry metric.Registry) *meteredTopicChannel {
-	bufferSize := 100 // TODO - when this is reduced TestLeanHelix_RecoversFromDiskError fails - whats the correct buffer size?
+	bufferSize := 10 // TODO increase significantly once main peer socket size is reduced
```
electricmonk: peer socket size is only for `memoryTransport`... what's the equivalent for TCP?

electricmonk:
@ronnno note that this fails multiple tests

services/gossip/dispatcher.go
ronnno: The flow is a little different in direct transport: the client maintains an outgoing queue as a buffered channel. the buffer size is `const SEND_QUEUE_MAX_MESSAGES = 1000` when a client tries to send a message to a peer it first writes to this buffer _unless it's full_. if it's full the message is discarded and the error is more or less ignored.

after writing to the outgoing queue buffer, the client transport thread will synchronously send all messages through the TCP socket. it will not drop messages after they are accepted to the buffer. 

in-memory transport simplifies this overflow behavior by replacing N sockets with buffer of 1000 with just one incoming `peer.socket` channel (not really a socket) with a buffer of N*1000. other than that the overflow policy to discard messages without reporting is similar between the two implementation

electricmonk:
@ronnno note that there's a memory leak test failing, probably meaning that a channel does not get evicted properly during shutdown

ronnno:
opened #1305

PR #1268, opened 2019-08-09
Run go test outside of the Docker network on CI 'e2e' step

itamararjuan:
This PR changes the Docker test to run the various `go test ...` commands outside the Docker scope and within the machine orchestrating the step instead.

In addition, the test suite ran all adapters (some which don't even require the Docker network) these we're removed and only tests which require a "live" production network are run now.

The motivation here is to allow a more concise log flow for the tests being run. The previous installement was pretty hard to understand at a glance since the STDOUT from all Docker containers was intertwined with the go test output which was also lacking some scope.


docker/test/docker-compose.yml
```diff
@@ -118,24 +119,24 @@ services:
...
-      ETHEREUM_PRIVATE_KEY: f2ce3a9eddde6e5d996f6fe7c1882960b0e8ee8d799e0ef608276b8de4dc7f19 # this keypair is generated from the mnemonic passed to ganache on startup
-      ETHEREUM_PUBLIC_KEY: 037a809cc481303d337c1c83d1ba3a2222c7b1b820ac75e3c6f8dc63fa0ed79b18
-      EXTERNAL_TEST: 'true'
-    command: sh -c "sleep 15 && ./test.adapters.sh"
-    sysctls:
-      net.core.somaxconn: 25000
-    networks:
-      orbs-network:
-        ipv4_address: 192.168.199.7
+  # orbs-e2e:
```
electricmonk: why comment out? delete!

docker/test/test.sh
```diff
@@ -1,22 +1,65 @@
...
+docker ps -aq | xargs docker rm -fv
+sleep 3
 
 rm -rf _logs
 
+[[ -z $CONSENSUSALGO ]] && echo "Consensus algo is not set! quiting.." && exit 1
+
+# Only in Lean Helix disable the initial block height test for now
+if [[ $CONSENSUSALGO == "leanhelix" ]]; then
+ export REMOTE_ENV="true"
```
electricmonk: why are you using this env variable?

docker/test/test.sh
```diff
@@ -1,22 +1,65 @@
...
+      ETHEREUM_PRIVATE_KEY=f2ce3a9eddde6e5d996f6fe7c1882960b0e8ee8d799e0ef608276b8de4dc7f19 
+      ETHEREUM_PUBLIC_KEY=037a809cc481303d337c1c83d1ba3a2222c7b1b820ac75e3c6f8dc63fa0ed79b18 \
+      EXTERNAL_TEST='true'
+
+# the ethereum keypair is generated from the mnemonic passed to ganache on startup
+
+echo "The network has started with pre-existing (ancient) 500-some blocks"
+echo "Sleeping to allow the network to start closing new blocks.."
+echo "(So that the txpool won't throw our calls to the bin)"
+sleep 10
```
electricmonk: this should be replaced with a poll of the block height metric, but you can do it in a subsequent PR - open a bug or a task on your monday board

docker/test/test.sh
```diff
@@ -1,22 +1,65 @@
...
+
+echo "Running E2E tests (AND a humble stress-test) w/consensus algo: ${CONSENSUSALGO}"
+go test -count=1 -v ./test/e2e/...
+
+echo "Running Ethereum Connector tests w/consensus algo: ${CONSENSUSALGO}"
```
electricmonk: no need to run these tests with both algos

docker/test/test.sh
```diff
@@ -1,22 +1,65 @@
...
+go test -count=1 -v ./services/crosschainconnector/ethereum/adapter/...
+
+echo "Running Gamma tests w/consensus algo: ${CONSENSUSALGO}"
```
electricmonk: are you sure Gamma actually uses the consensus algo from this env variable in its e2e?

test.adapters.sh
```diff
@@ -1,13 +1,13 @@
 #!/bin/sh -xe
 
 for pkg in $(find . -type d -name e2e | grep -v vendor | grep -v ".git"); do
-    time go test $pkg -v
+    time go test $pkg -v -count=1
```
electricmonk: why did you even leave this file here if you run tests directly from `docker/test/test.sh`?

docker/test/docker-compose.yml
itamararjuan: You are right - will be deleted.  @electricmonk

docker/test/test.sh
itamararjuan: Because this env variable is already implemented to notify a remote env. In essence - we are running against a remote env which skips this test already. @electricmonk

docker/test/test.sh
itamararjuan: Agreed - I was thinking of implementing something already in this PR. I already have code for this inplace which only needs some work to make it accept arbitrary IPs like in here.

I gave up on it in the effort to not include too many moving pieces in the same PR and get the tests to run in this new constellation first.

docker/test/test.sh
itamararjuan: Do elaborate please @electricmonk  ?

docker/test/test.sh
itamararjuan: Gamma's consensus algo is not relevant here actually. This is only my lazyness of copying the same `echo` line to keep things descriptive. Gamma only uses the Ganache brought up here as far as I'm aware of. I don't think Gamma uses anything else than LH at this point, no?

test.adapters.sh
itamararjuan: I have a monday task to clean up the entire folder structure so it's nice and tidy. for the meanwhile I don't intend to delete this file until we sit with @netoneko and other contributors and understand who uses which of our fleet of `test` bash scripts. But for the meanwhile it seemed essential to at least make sure whoever uses this to test something isn't fooled by Go's test cache.

itamararjuan: Do you know with a 100% certainty that we should delete this bash script ?

electricmonk: I think it's there solely for the e2e to run

docker/test/test.sh
electricmonk: point being that maybe Ethereum and Gamma's tests should have a separate step, that's running only once regardless of consensus algo

electricmonk: they also need *only* ganache, so maybe my intention of running everything that needs any docker is wrong, and we should run one step for docker-based integration tests, and one for e2es

docker/test/test.sh
electricmonk: if I understand it correctly, `test.sh` will run once per consensus algo, so you'll run these tests twice for no reason

docker/test/test.sh
electricmonk: yes yes. good.

docker/test/test.sh
itamararjuan: We do intend to make sure Lean Helix is able to come up with old storage and close the first block. So in my eyes this is temporary in some way.

PR #1279, opened 2019-08-13
sanitization of contract names 

### LINKED ISSUE #571: Security of smart contract mechanism - sanitize contract name
To avoid similar names like "TheToken" and "TheToken " or "TheToken_" that will confuse users

### LINKED ISSUE #1067: Processor returns wrong error when trying to deploy a contract with a name that is system contract
When you try to deploy a contract with a name that matches a system contract 
i get this error:
{
  "RequestStatus": "COMPLETED",
  "TxId": "0x182CF8888a538d15BbcD88794B4b9d900D0fE668941955fBDB708e6468290302bE91E63089a25cE4",
  "ExecutionResult": "ERROR_SMART_CONTRACT",
  "OutputArguments": [
    {
      "Type": "string",
      "Value": "method '_init' not found on contract '_Elections'"
    }
  ],
  "OutputEvents": [],
  "TransactionStatus": "COMMITTED",
  "BlockHeight": "2",
  "BlockTimestamp": "2019-03-19T09:55:01.031Z"
}

it should be something like "contract already deployed"

noambergIL:
## Description

sanitization of contract names :
contract names checked before deploy that they are not system and are alphanumeric only.
also for deploy contract name is case insensitive. HOWEVER contract name is case sensitive for reading/accessing.
(example can deploy "randomName", then access it only as "randomName", but can't deploy "RANDOMNAME" as a new contract.

also added test to check all system contracts are checked as such for proper deploy

fixes #1067
fixes #571

### DIFF
--- a/bootstrap/gamma/deploy_native_test.go
+++ b/bootstrap/gamma/deploy_native_test.go
@@ -10,9 +10,13 @@ package gamma
 import (
 	"context"
 	"fmt"
+	"github.com/orbs-network/orbs-network-go/services/processor/native/repository/_Elections"
 	"github.com/orbs-network/orbs-network-go/test"
 	"github.com/orbs-network/orbs-network-go/test/acceptance/callcontract"
+	"github.com/orbs-network/orbs-network-go/test/builders"
 	"github.com/orbs-network/orbs-network-go/test/contracts"
+	"github.com/orbs-network/orbs-network-go/test/crypto/keys"
+	"github.com/orbs-network/orbs-spec/types/go/protocol"
 	"github.com/orbs-network/orbs-spec/types/go/protocol/consensus"
 	"github.com/orbs-network/scribe/log"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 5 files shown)


jlevison:
- dash (-) and underscore (_) should also be allowed. is there a specific reason to exclude them
- alphanumeric as in standard ascii or unicode? should be ascii (i.e.  'a' to 'z' and then '0' to '9' at char codes)
- are spaces trimmed (and then the name is compared) or denied due to leading/following spaces?
- contract name should be case insensitive, i.e. deployed as lowercase and used as lowercase when the code is searched in the registry - any other way is extremely confusing

regardless of the comments, i think this should be tested on the e2e/component level and not just the unit level, @electricmonk ?

electricmonk:
Agree with @jlevison about an e2e test - add it under `gamma/deploy_native_test` maybe?

noambergIL:
1) alpha-numeric is ascii - (look in code regex)
2) all others are rejected while talking with Oded.
3) if we deploy with "Random" and the allow gamma calls with "random" "Random" "RANDOM" is more confusing. We want the user to deploy and use one name but prevent scammers for adding different cases but also - and _ and " " in the middle (or beginning or end).
4) Also since contract name is part of the key saved in merkle and state this means Not backward compatible (elections !) and also lots of code to go through to make sure all contact with contract is lower cased.

PR #1296, opened 2019-08-28
Feature/reputation basic

noambergIL:
First part of the committee feature:
1) lean-helix now gives info about who proposed the block
2) spec now has field of block proposer in the block
3) membuff now supports eval with ending missing (forward compatibility)
4) fill in consensus context code for create/validate with new field in blocks.


services/consensusalgo/leanhelixconsensus/service.go
```diff
@@ -183,7 +183,8 @@ func (s *Service) validateBlockExecutionIfYoung(ctx context.Context, blockPair *
 			CurrentBlockHeight: blockPair.TransactionsBlock.Header.BlockHeight(),
 			PrevBlockHash:      blockPair.TransactionsBlock.Header.PrevBlockHashPtr(),
 			TransactionsBlock:  blockPair.TransactionsBlock,
-			PrevBlockTimestamp: prevBlockPair.TransactionsBlock.Header.Timestamp()})
+			PrevBlockTimestamp: prevBlockPair.TransactionsBlock.Header.Timestamp(),
+			BlockProposerAddress: blockPair.TransactionsBlock.Header.BlockProposerAddress()}) // TODO NOAM maybe need to ask lean helix lib for proposer
```
IdoZilberberg: @gadcl - (1) any reason block proposer would be different between TransactionBlock and ResultsBlock? If block proposer is always the same then this code is good and can remove the TODO comment. (2) If it's a duplicate of the same field in Transactions Block, can we just remove this field from Results Block?

services/consensuscontext/validate_results_block.go
```diff
@@ -61,6 +61,22 @@ func validateRxBlockHeight(ctx context.Context, vcrx *rxValidatorContext) error
 	return nil
 }
 
+func validateRxBlockProposer(ctx context.Context, vcrx *rxValidatorContext) error {
+	blockProposer := vcrx.input.ResultsBlock.Header.BlockProposerAddress()
+	// TODO NOAM BC new field BPA
```
IdoZilberberg: What does this comment mean?

services/consensuscontext/validate_results_block_test.go
```diff
@@ -253,3 +265,38 @@ func MockProcessTransactionSetThatReturns(err error) func(ctx context.Context, i
 		return someEmptyTxSetThatWeReturnOnlyToPreventErrors, err
 	}
 }
+
+func TestConsensusContextValidateResultsBlock_TestBlockProposerNotSame(t *testing.T) {
```
IdoZilberberg: Test names are misleading - one is for Tx Block and the other for Rx Block, but that's not what the same say.

services/consensusalgo/leanhelixconsensus/service.go
gadcl: consistent with the current `block pair` design, as to support future split of execution and ordering processes.

services/consensuscontext/validate_transactions_block_test.go
```diff
@@ -23,15 +23,27 @@ import (
 )
 
 func toTxValidatorContext(cfg config.ConsensusContextConfig) *txValidatorContext {
-	block := builders.BlockPairBuilder().Build()
-	prevBlockHashCopy := make([]byte, 32)
+	return toTxValidatorContextWithBc(cfg, false)
+}
+
+func toTxValidatorContextWithBc(cfg config.ConsensusContextConfig, isBc bool) *txValidatorContext {
```
IdoZilberberg: Please use full names - what's `isBc`?

IdoZilberberg:
I opened https://github.com/orbs-network/lean-helix-go/pull/68 for lean helix.
@gadcl @noambergIL take a look there too.

services/consensuscontext/validate_results_block.go
noambergIL: fixed comment

services/consensuscontext/validate_results_block_test.go
noambergIL: names have been improved but the first test checks block-results-header against function input. Second checks block-results-header against block-tx-header.

services/consensuscontext/validate_transactions_block_test.go
noambergIL: changed name to isBackwardsCompatible -> creats the block with an empty hash (len=0) block-proposer to simulate an older version

noambergIL:
@electricmonk validateBlockProposalInternal calls the consensuscontext validate so i feel there is no need for specific test.

PR #1299, opened 2019-08-29
Feature/reputation basic contract

noambergIL:
The contract and it's uses.


services/processor/native/repository/Committee/index.go
```diff
@@ -0,0 +1,78 @@
...
+func _split(oneArrayOfAddresses []byte) [][]byte {
+	numAddresses := len(oneArrayOfAddresses) / digest.NODE_ADDRESS_SIZE_BYTES
+	res := make([][]byte, numAddresses)
+	for i := 0; i < numAddresses; i++ {
+		res[i] = oneArrayOfAddresses[digest.NODE_ADDRESS_SIZE_BYTES*i : digest.NODE_ADDRESS_SIZE_BYTES*(i+1)]
+	}
+	return res
+}
+
+func _addressSliceToArray(a []byte) [20]byte {
```
golangcibot: `_addressSliceToArray` is unused (from `deadcode`)

noambergIL: fixed.

services/processor/native/repository/Committee/update.go
```diff
@@ -0,0 +1,52 @@
...
+
+func CommitteeMemberReputationSetEvent(address []byte, reputation uint32) {}
+func _updateOrderedCommittee(orderedCommittee [][]byte, blockProposer []byte)  {
+	for _, member := range orderedCommittee {
+		if bytes.Equal(member, blockProposer) {
+			_clearReputation(member)
+			events.EmitEvent(CommitteeMemberReputationSetEvent, member, uint32(0))
+			break
+		} else {
+			_degradeReputation(member)
```
electricmonk: wouldn't that degrade the reputation of addresses AFTER the block proposer? is that correct? aren't we punishing nodes that didn't even get the *chance* to propose a block?

services/consensuscontext/ordered_committee.go
```diff
@@ -108,21 +104,25 @@ func (s *service) callElectionsSystemContract(ctx context.Context, blockHeight p
...
-	return res, nil
+	return res
 }
 
-func toNodeAddresses(nodes map[string]config.ValidatorNode) []primitives.NodeAddress {
-	nodeAddresses := make([]primitives.NodeAddress, len(nodes))
-	i := 0
+func generateGenesisCommittee(nodes map[string]config.ValidatorNode) []primitives.NodeAddress {
+	res := make([]primitives.NodeAddress, 0, len(nodes))
 	for _, value := range nodes {
```
gadcl: range will diverge on different nodes?
We need to inflict order on genesis committee list (based on height) - same as before - support the blocks until election

services/processor/native/repository/Committee/committee.go
```diff
@@ -0,0 +1,72 @@
...
+	binary.LittleEndian.PutUint64(seedBytes, env.GetBlockHeight())
+
+	return _orderList(addresses, seedBytes)
+}
+func _orderList(addrs [][]byte, seed []byte) [][]byte {
+	addrsToSort := addrsAndGrades{ addrs, make([]float64, len(addrs))}
+	for i, addr := range addrs {
+		addrsToSort.grades[i] = _calculateGradeWithReputation(addr, seed)
+	}
+	sort.Sort(addrsToSort)
```
gadcl: does "sort" deterministically breaks ties?

services/processor/native/repository/Committee/index.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func _getElectedValidators() []byte {
+	outputArray := service.CallMethod(elections_systemcontract.CONTRACT_NAME, elections_systemcontract.METHOD_GET_ELECTED_VALIDATORS)
+	return outputArray[0].([]byte)
+}
+
+func _concat(addresses [][]byte) []byte {
+	oneArrayOfAddresses := make([]byte, 0, len(addresses)*digest.NODE_ADDRESS_SIZE_BYTES)
+	for _, addr := range addresses {
```
gadcl: doesn't range change order?

services/processor/native/repository/Committee/index.go
```diff
@@ -0,0 +1,72 @@
...
+	return []byte(fmt.Sprintf("Validator_%s_Rep", hex.EncodeToString(addr)))
+}
+
+func _getReputation(addr []byte) uint32 {
+	return state.ReadUint32(_formatReputation(addr))
+}
+
+func _degradeReputation(addr []byte) {
+	currRep := _getReputation(addr)
+	if currRep < ReputationBottomCap {
```
gadcl: for auditability reasons degrade reputation as much as needed - on query - it will retrieve the correct (not the capped on).  On compute of reputation use the bounds (cap above and below)

services/processor/native/repository/Committee/update.go
```diff
@@ -0,0 +1,52 @@
...
+		if bytes.Equal(member, blockProposer) {
+			return true
+		}
+	}
+	return false
+}
+
+func CommitteeMemberReputationSetEvent(address []byte, reputation uint32) {}
+func _updateOrderedCommittee(orderedCommittee [][]byte, blockProposer []byte)  {
+	for _, member := range orderedCommittee {
```
gadcl: range -changes order?
Important note: _updateOrderedCommittee - assumes orderedCommittee given is the same (members and order) as obtained on the current block height (on new consensus round)

services/processor/native/repository/Committee/update.go
```diff
@@ -0,0 +1,52 @@
...
+package committee_systemcontract
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/env"
+	"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/events"
+)
+
+func updateReputation() {
```
gadcl: As for your third comment from the email - we should check signer/trigger for updateReputation so to permit only "system access" => empty signer ..

services/processor/native/repository/Committee/committee.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func _calculateReputationMarkDownFactor(reputation uint32) float64 {
+	if reputation < ToleranceLevel {
+		return 1.0
+	}
+	return math.Pow(2, float64(reputation))
+}
+
+type addrsAndGrades struct {
```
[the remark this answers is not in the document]
noambergIL: done on purpose to avoid extra copy of the addresses list and doing an "in place" sort

services/processor/native/repository/Committee/committee.go
noambergIL: sort by grade/score then sort by address... so yes :)

services/processor/native/repository/Committee/index.go
noambergIL: only range of map is non-deterministic

services/processor/native/repository/Committee/update.go
noambergIL: oder of range of array is constant. since seed is the same (block height) then yes.

services/processor/native/repository/Committee/update.go
noambergIL: TestOrbsCommitteeContract_updateOrderedCommittee this shows that what you asked doesn't happen. the break stops the for loop after you find the proposer and zero it's reputation

services/consensuscontext/ordered_committee.go
noambergIL: after copy there is sort so no issue about that. but it doesn't do random here. maybe we do need a version of orderCommitteeWithInput to be exposed ?

services/processor/native/repository/Committee/index.go
noambergIL: doesn't that make some of the calculation opaque ? since if you audit you need to know the caps ?

noambergIL:
@gadcl added a call for order in genesis. 
Added signer and check that trigger is the caller.
separated the misses from the reputation.

PR #1323, opened 2019-09-16
Fix passing empty array of variable length to a contract

netoneko:
Fixes the bug when `f(1)` would fail if it was defined in the smart contract.

```golang
package main

import "fmt"

func f(a int, bs...uint) {
	fmt.Println(a, bs)
}

func main() {
	f(1)
}
```

### DIFF
--- a/services/processor/native/call.go
+++ b/services/processor/native/call.go
@@ -136,7 +136,14 @@ func (s *service) prepareMethodInputArgsForCall(methodInstance types.MethodInsta
 	}
 
 	if i < methodType.NumIn() {
+		if methodType.NumIn()-1 == i { // missing a single argument
+			if k := methodType.In(methodType.NumIn() - 1).Kind(); k == reflect.Slice {
+				return res, nil
+			}
+		}
+
 		return nil, errors.Errorf("method '%s' takes %d args but received less", functionNameForErrors, methodType.NumIn())
+
 	}
 
 	return res, nil
--- a/services/processor/native/call.go
+++ b/services/processor/native/call.go
@@ -107,16 +107,29 @@ func (s *service) prepareMethodInputArgsForCall(methodInstance types.MethodInsta
 		case reflect.Slice:
 			switch methodTypeIn.Elem().Kind() {
 			case reflect.Uint8:
+				if !arg.IsTypeBytesValue() {
+					return nil, errors.Errorf("method '%s' expects arg %d to be []byte but it has %s", functionNameForErrors, i, arg.StringType())
+				}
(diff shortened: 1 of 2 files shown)


services/processor/native/call.go
```diff
@@ -136,7 +136,14 @@ func (s *service) prepareMethodInputArgsForCall(methodInstance types.MethodInsta
 	}
 
 	if i < methodType.NumIn() {
+		if methodType.NumIn()-1 == i { // missing a single argument
```
ronnno: for non-variadic arrays (`[]byte`) in the last element you will allow the function call to proceed if the array was not included.

services/processor/native/call.go
```diff
@@ -136,7 +136,14 @@ func (s *service) prepareMethodInputArgsForCall(methodInstance types.MethodInsta
...
+			if k := methodType.In(methodType.NumIn() - 1).Kind(); k == reflect.Slice {
+				return res, nil
+			}
+		}
+
 		return nil, errors.Errorf("method '%s' takes %d args but received less", functionNameForErrors, methodType.NumIn())
+
```
ronnno: in rows 110, 112, 114, 116 there is no check that `arg` is in agreement with `Elem().Kind()`. consider adding tests for cases where an array or variadic are expected, but a different type is passed in

services/processor/native/call.go
ronnno: will `[][]byte` work in the last argument?

electricmonk:
@netoneko @ronnno let's involve @OdedWx in this

electricmonk:
also @netoneko why does the e2e fail?

netoneko:
@ronnno @noambergIL fixed all the comments (found `IsVariadic` call in `reflect` package which greatly simplified the code).

@electricmonk why involve @OdedWx? This is just a matter of implementing idiomatic Go and nothing more.

Regarding e2e - it fails for separate reasons not related to the pr.

services/processor/native/call.go
netoneko: Resolved.
[][]byte will not work because after introduction of `isVariadic` call you won't be able to pass `([]byte, []byte)` as `[][]byte` (not sure if you could before).

services/processor/native/call.go
netoneko: Fixed.

netoneko:
Now the e2e passes.

Previous e2e failures seem to be reproducible on master: https://circleci.com/gh/orbs-network/orbs-network-go/19223

`Error:      	Expected nil, but got: failed to deploy native contract: http status 202 Accepted`

PR #1336, opened 2019-09-22
Fixed Lean Helix memory leak

IdoZilberberg:
* Fixed lean-helix [memory leak](https://github.com/orbs-network/lean-helix-go/pull/77) - updated lean-helix dependency to [v0.2.2](https://github.com/orbs-network/lean-helix-go/releases/tag/v0.2.2)
* Enabled "profiling" config option - this enables the [debug/pprof endpoints](https://github.com/orbs-network/orbs-network-go/blob/master/bootstrap/httpserver/server.go#L221)
* Added new memory test under `memoryleak` build flag: `TestMemoryLeaks_DuringRuntime`. This, however, might not be a good measure of memory leak, because measuring after internal tests (such as `TestGazillionXXX`) have completed, means the system under test already went down (test harness canceled its context) thus all of its memory - leaked or otherwise - was already released.


itamararjuan:
I approved but please do notice that your tests have failed on CircleCI - be sure to update master and have them re-run before merging.

Issue #1339, opened 2019-09-23
TestPeriodicalTriggerKeepsGoingOnPanic Deadlock


ronnno:
```
goroutine 73 [chan receive]:
github.com/orbs-network/orbs-network-go/synchronization.(*PeriodicalTrigger).Stop(...)
	/home/circleci/project/synchronization/trigger.go:68
github.com/orbs-network/orbs-network-go/synchronization_test.TestPeriodicalTriggerKeepsGoingOnPanic.func1(0x8e9200, 0xc0001ba380, 0xc0001c8230)
	/home/circleci/project/synchronization/trigger_test.go:154 +0x1af
github.com/orbs-network/orbs-network-go/test.WithConcurrencyHarness(0x8ed1c0, 0xc0001c0400, 0xc00019cf88)
	/home/circleci/project/test/concurrency.go:37 +0x2c9
github.com/orbs-network/orbs-network-go/synchronization_test.TestPeriodicalTriggerKeepsGoingOnPanic(0xc0001c0400)
	/home/circleci/project/synchronization/trigger_test.go:140 +0x5a
testing.tRunner(0xc0001c0400, 0x8797a8)
	/usr/local/go/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:916 +0x35a

goroutine 74 [chan send]:
github.com/orbs-network/orbs-network-go/synchronization_test.(*collector).Error(0xc0001ca008, 0x864236, 0xf, 0xc000011180, 0x3, 0x3)
	/home/circleci/project/synchronization/trigger_test.go:29 +0x80
github.com/orbs-network/orbs-network-go/instrumentation/logfields.(*govnrErrorer).Error(0xc0001b4280, 0x8e2120, 0xc000011160)
	/home/circleci/project/instrumentation/logfields/govnr_error_handler.go:18 +0x204
github.com/orbs-network/govnr.recoverPanics(0x8e2000, 0xc0001b4280)
	/go/pkg/mod/github.com/orbs-network/govnr@v0.2.0/panic.go:24 +0xf3
panic(0x7d4b80, 0x8d9a50)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/orbs-network/orbs-network-go/synchronization_test.TestPeriodicalTriggerKeepsGoingOnPanic.func1.1()
	/home/circleci/project/synchronization/trigger_test.go:146 +0x40
[...]

PR #1340, opened 2019-09-23
Resolve flakiness in trigger_test.go

ronnno:
See #1339 for flakiness in `TestPeriodicalTriggerKeepsGoingOnPanic` 
additionally noticed similar flakiness in `TestPeriodicalTriggerStartsOk`.

This PR addresses both issues

### DIFF
--- a/synchronization/trigger_test.go
+++ b/synchronization/trigger_test.go
@@ -139,9 +139,11 @@ func TestPeriodicalTriggerKeepsGoingOnPanic(t *testing.T) {
 		}, nil)
 
 		// a second invocation of the handler means the trigger recovered from the first panic
+		var lastObservedInvocationCount uint32
 		require.True(t, test.Eventually(1*time.Second, func() bool {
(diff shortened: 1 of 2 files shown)


synchronization/trigger_test.go
```diff
@@ -140,21 +132,18 @@ func TestPeriodicalTriggerKeepsGoingOnPanic(t *testing.T) {
...
 			panic("we should not see this other than the logs")
 		}, nil)
 
-		// more than one error means more than one panic means it recovers correctly
-		for i := 0; i < 2; i++ {
-			<-logger.errors
-		}
+		// a second invocation of the handler means the trigger recovered from the first panic
+		require.True(t, test.Eventually(1*time.Second, func() bool {
+			return atomic.LoadUint32(&handlerInvocationCount) >= 2
```
IdoZilberberg: @ronnno maybe extract the value of atomic.Load...() to a variable, to eliminate the chance of confusion by getting different values on the 2 consecutive calls.

PR #1354, opened 2019-10-02
Update .version file automatically on release

itamararjuan:
This PR proposes a change to the script which builds the candidate Docker image for the Orbs node - if it identifies it's being run within a scope of a release - it will re-write the contents
of the `.version` file to indicate the correct version per the name of the Git Tag currently being pushed


netoneko:
Maybe it would be easier to just break the build if the contents of .version file and the tag don't match.

That way everyone would be forced to update the version file correctly.

Also `$CIRCLE_TAG` should start with `v`.

itamararjuan:
> please delete the version file altogether and always use the latest tag from git
> if we're releasing ($CIRCLE_TAG isn't empty), use it, otherwise use the latest tag from git and suffix something like the current date or build number.

I've made a few changes I think you migth like - have a look please.

itamararjuan:
> Maybe it would be easier to just break the build if the contents of .version file and the tag don't match.
> 
> That way everyone would be forced to update the version file correctly.
> 
> Also `$CIRCLE_TAG` should start with `v`.

I've made also a validation to make sure the version starts with 'v' - otherwise I fail the build.

docker/build/build.sh
```diff
@@ -1,4 +1,23 @@
 #!/bin/bash -e
+
+if [[ ! -z "$CIRCLE_TAG" ]]; then
```
IdoZilberberg: nitpicking: you can also use `-n` instead of `! -z`

PR #1356, opened 2019-10-03
integration test for new bytes20/32 and all code to run it

### LINKED ISSUE #1229: need better integration test for changing contract sdk
What i did:
i wanted to add a new sdk to contract ethereum.getTime - very similar to getBlock that exists.
i knew it is in processor, i knew that its in contract-sdk, i knew it has cross chain connector.
i didn't know it is also in vm so i only found out after it was in gamma and called it via a real contract

1) hard to follow as some of the moving from layer to layer is done via strings and switch on them.
and also its many layers
2) the code in each layer is very boiler plate-y with very very simple unit tests  

so i want
1) simple integration test where i write a mini contract with new sdk then run the contract and get a fake value.
2) less boiler plate code or auto generate it ?
3) i have issues with the vendors contract-sdk as it's not easy to know which one you have and update it, and in the same commit too.

noambergIL:
fixes #1229


test/sdk/fixed_bytes_contract_test.go
```diff
@@ -0,0 +1,78 @@
...
+				    Build(),
+				builders.Transaction().WithMethod(ContractName, "setAddress").WithArgs(bytes20).Build(),
+				builders.Transaction().WithMethod(ContractName, "setHash").WithArgs(bytes32).Build(),
+				builders.Transaction().WithMethod(ContractName, "getAddress").WithArgs().Build(),
+				builders.Transaction().WithMethod(ContractName, "getHash").WithArgs().Build(),
+				builders.Transaction().WithMethod(ContractName, "setAddress").WithArgs(almostEmptyBytes20).Build(),
+				builders.Transaction().WithMethod(ContractName, "getAddress").WithArgs().Build(),
+			}
+
+			out, err := harness.vm.ProcessTransactionSet(ctx, &services.ProcessTransactionSetInput{
```
electricmonk: as discussed, please hide all irrelevant stuff inside the harness, and create a function `h.process(ctx, tx1, tx2, tx3)`

docker/build/go.mod.template
```diff
@@ -7,6 +7,6 @@ go 1.12
 // so that built contracts won't break the system.
 
 require (
-	github.com/orbs-network/orbs-contract-sdk v1.2.0
+	github.com/orbs-network/orbs-contract-sdk v1.3.0
```
electricmonk: this file should no longer be needed

Issue #1273, opened 2019-08-11
Flakiness in TestReadFromEthereumLogsTakingFinalityIntoAccount


netoneko:
```
info 2019-08-11T10:20:23.303124Z transaction execution failed node=a32884 service=virtual-machine result=EXECUTION_RESULT_ERROR_SMART_CONTRACT error=Trying to read from an unbound contract function=virtualmachine.(*service).runMethod source=services/virtualmachine/execution.go:65 
```

```
--- FAIL: TestReadFromEthereumLogsTakingFinalityIntoAccount (7.42s)
    harness.go:51: TestReadFromEthereumLogsTakingFinalityIntoAccount running Gamma at http://127.0.0.1:27572
    require.go:201: 
        	Error Trace:	harness.go:80
        	            				ethereum_finality_test.go:119
        	            				context.go:17
        	            				ethereum_finality_test.go:92
        	Error:      	Not equal: 
        	            	expected: "SUCCESS"
        	            	actual  : "ERROR_SMART_CONTRACT"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-SUCCESS
        	            	+ERROR_SMART_CONTRACT
        	Test:       	TestReadFromEthereumLogsTakingFinalityIntoAccount
        	Messages:   	failed calling LogCalculator.sum
```

PR #1378, opened 2019-10-17
Fix flakiness of TestReadFromEthereumLogsTakingFinalityIntoAccount

vistra:
At some point there is a query immediately following a transaction. In one of my executions the query failed because it ran on a node that wasn't synced with the transaction.
This will poll on the query result to assure that the block height is at least as high as that of the transaction.
Will poll for no longer than 10 seconds.

### DIFF
--- a/bootstrap/gamma/harness.go
+++ b/bootstrap/gamma/harness.go
@@ -69,10 +69,17 @@ func DeployContract(t *testing.T, orbs *orbsClient.OrbsClient, account *orbsClie
 	SendTransaction(t, orbs, account, "_Deployments", "deployService", "LogCalculator", uint32(protocol.PROCESSOR_TYPE_NATIVE), []byte(code))
 }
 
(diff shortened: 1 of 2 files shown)


bootstrap/gamma/harness.go
```diff
@@ -69,10 +69,17 @@ func DeployContract(t *testing.T, orbs *orbsClient.OrbsClient, account *orbsClie
...
 	q, err := orbs.CreateQuery(sender.PublicKey, contractName, method, args...)
 	require.NoError(t, err, "failed creating query %s.%s", contractName, method)
-	res, err := orbs.SendQuery(q)
+
+	// Allow no more than 10 seconds for the network to sync
+	var res *codec.RunQueryResponse
+	require.True(t, test.Eventually(10*time.Second, func() bool {
+		res, err = orbs.SendQuery(q)
+		return err != nil || res.BlockHeight >= minBlockHeight
+	}), "network did not sync - unable to obtain response for a lower bound block height")
```
electricmonk: "state storage is out of sync"

electricmonk:
fixes #1273

vistra:
fixes #1273

PR #1386, opened 2019-10-23
Experimental libraries

netoneko:
Brings experimental libraries back after the regression that happened during go modules migration.

@itamararjuan please enable them only for non-tagged builds as it was before.

This is a serious regression that impedes work of Tech Engagement group. I suggest we merge it now and work on the fix (exclusion from tagged releases) afterwards, because Tech Engagement projects rely on experimental builds.


itamararjuan:
These changes are supposed to fail the e2e's because go.mod would require a change in this case
We have to do a manual process of adding this dep into the go.mod and see what go wants to do with the `go.mod` file.

electricmonk:
let's have a quick discussion tomorrow morning and understand what we can do. Adding external libraries to releases is a big no-no.

netoneko:
@itamararjuan it does not require changing a default `go.mod` file because it doesn't use it; e2e on LH fails because of completely separate issue. Notice that Benchmark e2e passes.

netoneko:
@electricmonk we're not currently planning for any releases, and the build is broken. We had an agreement about these libraries and they were removed without any discussion with us.

Breaking a build that is used for demos is also a big no-no. I am sure you guys can figure out how to exclude the libraries from the release build in a separate pr.

electricmonk:
No. Master should be releasable at all point. The external libraries are considered a non-critical component. They were removed by mistake due to a missing test that would’ve alerted us to the unintended removal.

Shai

PR #1396, opened 2019-10-27
Remove vendoring

itamararjuan:
This PR removes support for the vendoring  solution using git modules.


netoneko:
You forgot about good old `git-submodule-checkout.sh`

PR #1397, opened 2019-10-27
add committee members cound and IDs into metrics

### LINKED ISSUE #1379: Metrics - committee members
Add current committee members info to metrics  (count ..)

amir-arad:
TBD before this is completed:
Testing metrics?
Is this the right place to track?
Is this the right metric / format?

fix #1379


amir-arad:
> Re: the race - might be a race in the Text metric itself, please investigate
 see latest commit. However, when testing locally for races I've got #1399 , which seems unrelated.

PR #1401, opened 2019-10-29
TCP Transport will no longer hang when its port is already in use

### LINKED ISSUE #1388: Gossip server goes to infinite loop on address already in use
`serverMainLoop` should signal to `NewDirectTransport` that the port is taken, so the loop doesn't run forever.

https://circleci.com/gh/orbs-network/orbs-network-go/20341?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

electricmonk:
fixes #1388

### DIFF
--- a/services/gossip/adapter/tcp/direct_transport_test.go
+++ b/services/gossip/adapter/tcp/direct_transport_test.go
@@ -33,7 +33,7 @@ func TestDirectTransport_HandlesStartupWithEmptyPeerList(t *testing.T) {
 		defer transport.GracefulShutdown(ctx)
 
 		require.True(t, test.Eventually(test.EVENTUALLY_ADAPTER_TIMEOUT, func() bool {
-			return transport.server.IsListening()
+			return transport.IsServerListening()
 		}), "server did not start")
 	})
 }
@@ -47,7 +47,7 @@ func TestDirectTransport_SupportsTopologyChangeInRuntime(t *testing.T) {
 		superviseAll(harness, node1, node2, node3, node4)
 		defer shutdownAll(ctx, node1, node2, node3, node4)
 
-		waitForAllNodesToSatisfy(t, "server did not start", func(node *nodeHarness) bool { return node.transport.server.IsListening() }, node1, node2, node3, node4)
+		waitForAllNodesToSatisfy(t, "server did not start", func(node *nodeHarness) bool { return node.transport.IsServerListening() }, node1, node2, node3, node4)
 
 		firstTopology := aTopologyContaining(node1, node2, node3)
 		node1.transport.UpdateTopology(ctx, firstTopology)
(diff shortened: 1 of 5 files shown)


services/gossip/adapter/tcp/direct_transport_test.go
```diff
@@ -33,7 +33,7 @@ func TestDirectTransport_HandlesStartupWithEmptyPeerList(t *testing.T) {
 		defer transport.GracefulShutdown(ctx)
 
 		require.True(t, test.Eventually(test.EVENTUALLY_ADAPTER_TIMEOUT, func() bool {
-			return transport.IsServerListening()
+			return transport.server.IsListening()
```
amir-arad: why not use  `IsServerListening()` in tests?

services/gossip/adapter/tcp/server.go
```diff
@@ -12,56 +12,91 @@ import (
...
+type transportServer struct {
+	sync.RWMutex
+	listener                          adapter.TransportListener
+	listening                         bool
+	port                              int
+	logger                            log.Logger
+	incomingConnectionAcceptSuccesses *metric.Gauge
+	incomingConnectionAcceptErrors    *metric.Gauge
+	incomingConnectionTransportErrors *metric.Gauge
+	activeIncomingConnections         *metric.Gauge
```
amir-arad: separate metrics from core functionality

services/gossip/adapter/tcp/server.go
```diff
@@ -12,56 +12,91 @@ import (
...
 )
 
-func (t *DirectTransport) serverListenForIncomingConnections(ctx context.Context, listenPort uint16) (net.Listener, error) {
+type serverConfig interface {
+	GossipListenPort() uint16
+	GossipNetworkTimeout() time.Duration
+}
+
+type transportServer struct {
+	sync.RWMutex
```
amir-arad: ideally the mutex (along with `lock()`, `unlock()`) should be a private implementation detail of the server.

services/gossip/adapter/tcp/server.go
```diff
@@ -12,56 +12,91 @@ import (
...
+		incomingConnectionAcceptSuccesses: registry.NewGauge("Gossip.IncomingConnection.ListeningOnTCPPortSuccess.Count"),
+		incomingConnectionAcceptErrors:    registry.NewGauge("Gossip.IncomingConnection.ListeningOnTCPPortErrors.Count"),
+		incomingConnectionTransportErrors: registry.NewGauge("Gossip.IncomingConnection.TransportErrors.Count"),
+		activeIncomingConnections:         registry.NewGauge("Gossip.IncomingConnection.Active.Count"),
+	}
+}
+
+func (t *transportServer) getPort() int {
+	t.Lock()
+	defer t.Unlock()
```
amir-arad: shouldn't that be `rLock()` and `rUnlock()`?

services/gossip/adapter/tcp/server.go
electricmonk: It is. Embedding a mutex is the idiomatic practice.

PR #1411, opened 2019-10-30
docker compose DRY refactor

### LINKED ISSUE #1318: DRY in docker/test/docker-compose.yml
@electricmonk wrote: https://github.com/orbs-network/orbs-network-go/pull/1303#discussion_r322587156

> this file has become long and repetitive; please see if we can simplify it

amir-arad:
It can be further generalised at the cost of convolution
fixes #1318

### DIFF
--- a/docker/test/common-services.yml
+++ b/docker/test/common-services.yml
@@ -1,0 +1,24 @@
+version: '2.1'
+
+services:
+  ganache:
+    image: trufflesuite/ganache-cli
+    command: ganache-cli --gasLimit 90000000000 --gasPrice 1 -a 10 -m "vanish junk genuine web seminar cook absurd royal ability series taste method identify elevator liquid"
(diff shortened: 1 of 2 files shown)


docker/test/common-services.yml
```diff
@@ -0,0 +1,24 @@
+version: '2.1'
+
+services:
+  ganache:
+    image: trufflesuite/ganache-cli
+    command: ganache-cli --gasLimit 90000000000 --gasPrice 1 -a 10 -m "vanish junk genuine web seminar cook absurd royal ability series taste method identify elevator liquid"
+    ports:
+      - 8545:8545
+
+  signer-node:
```
netoneko: It would make a lot of sense to move the command here as well.

amir-arad: opened #1418

Issue #1181, opened 2019-06-05
Contract pre-compilation during startup causes denial of service


netoneko:
**Describe the bug**

When the gamma-server or orbs node goes up, it compiles a certain contract to make subsequent contract compilation faster by building up go cache.

During large scale deployments, the machine can simply run out of memory if enough containers start simultaneous compilation. I've seen memory go up to 4.5Gb for just 15 gamma containers and then go down to 700Mb when the compilation was finished.

**Steps To Reproduce**

Run a bunch of gamma containers at the same time.

**Expected behavior**

15 gamma containers go up and there is no CPU and memory penalty for running them together.

netoneko:
I think that we could build up the cache during docker build phase by exposing a certain method in `orbs-node` and `gamma-server` CLI that allows cache warm up.

It **is** weird, but I believe it is our best shot right now.

electricmonk:
I think it might be worth a shot - as long as it's an _optimization_ and Gamma is still usable outside of Docker as well.

netoneko:
Fixed in 7c1cca267b6ddaee9ca0b9a448f76e2960364f4e

PR #1414, opened 2019-10-31
Retry on warmup compilation failures and panic if eventually unsuccesful 

itamararjuan:
Title pretty much sums it up.  This is so we can avoid Lean Helix timing problems
when warm up compilation fails on some nodes on Docker E2Es


electricmonk:
Please investigate why the E2E fails in this PR

netoneko:
Found the original bug report: #1181

PR #1415, opened 2019-11-01
retry allocating a free port for e2e gossip until getting a free port

electricmonk:


electricmonk:
merging without review in an attempt to reduce flakiness of in-process e2e

PR #1420, opened 2019-11-03
adding liveness check for chains in e2e

itamararjuan:


electricmonk:
I don't like what you did here. It seems as if you're prematurely generalizing code that has too many differences. Either extract a common module and reuse it in both testnet polling and local e2e polling, or duplicate the code and extract later when there's more concrete evidence.

itamararjuan:
I'm closing this PR in favor of a different direction for this problem.

PR #1423, opened 2019-11-03
skip in-process e2es in CI

electricmonk:


netoneko:
This will increase the feedback loop because regular tests fail e2e much faster than docker e2e.

electricmonk:
@netoneko yes, please see #core-team on Slack for a list of pros and cons

PR #1424, opened 2019-11-03
remove services/processor/javascript - dead code

### LINKED ISSUE #1422: consider removing services/processor/javascript
This code is really dead.

noambergIL:
no one actually uses this code.
fixes #1422


electricmonk:
@talkol this is your old implementation. It will always remain in our hearts and in Git history but we'd like to eliminate it from tip of master. Do we have your blessing?

netoneko:
Can you please not do anything with it right now? It will just make my merge harder.

PR #1427, opened 2019-11-03
Pre-allocate gossip source ports on in-process E2E

vistra:
Pre allocates source ports to prevent collision with server ports, and falls-back to OS allocation in case of a collision


electricmonk:
are you sure this is a good idea?

vistra:
Decided to first try hard coded ports, if that still collides we can try this.

PR #1434, opened 2019-11-05
Flaky E2Es - run query repeatedly until getting a response at the expected block height

electricmonk:


test/e2e/harness.go
```diff
@@ -70,56 +70,51 @@ func newAppHarness() *harness {
...
+	out, err := h.client.SendTransaction(payload)
+	return out.TransactionResponse, txId, err
 }
 
-func (h *harness) eventuallyRunQueryWithoutError(timeout time.Duration, senderPublicKey []byte, contractName string, methodName string, args ...interface{}) (response *codec.RunQueryResponse, err error) {
+func (h *harness) runQueryAtBlockHeight(timeout time.Duration, expectedBlockHeight uint64, senderPublicKey []byte, contractName string, methodName string, args ...interface{}) (response *codec.RunQueryResponse, err error) {
 	test.Eventually(timeout, func() bool {
 		response, err = h.runQuery(senderPublicKey, contractName, methodName, args...)
-		return err == nil
+		return err == nil && response.BlockHeight >= expectedBlockHeight
```
vistra: is it ok for runQuery to return an error at some point and be successful later? If not then I think it should be		`return err != nil || response.BlockHeight >= expectedBlockHeight`

test/e2e/harness.go
```diff
@@ -70,56 +70,51 @@ func newAppHarness() *harness {
...
 	})
 	return response, err
```
vistra: If the height was never reached then this would return with a response for a lower height

test/e2e/harness.go
electricmonk: it's not enough that the error isn't nil - we're waiting for a block at least as high as `expectedBlockHeight`

test/e2e/harness.go
electricmonk: no, that's the whole point here

test/e2e/harness.go
electricmonk: yeah, you're actually right

test/e2e/harness.go
electricmonk: actually yes

PR #1435, opened 2019-11-06
refactor to use spec level ArgumentArray

noambergIL:
1) refactor to use spec level ArgumentArray building **this code is more technical skim only**
 * find a lot of places that need to be swtiched, removed almost all of test/builders/arguments code
* remove the builders as much as possible 
2) processor/native/call & call_test.go - **please look at this part more closely**
 * seperate transpilation of native types to prtocol.Argument -> all the swtich/case code can be removed. if the translpilation doesn't work there is no need to try and run the method or if it ran but the output type is not valid for transaction then fail on that.
* before running the contract method check that the native args match in numbers and types that of the method (code is MUCH more simple)
* to prove this code works use more or less the original call_test.go code with some minor adjustments : 
-one test to look at happy and un happy processMethodCall (replacese the call_arguments test that was in processor/native/test as a component test)
-many tests to prove we can send tx with all kinds of type of argeuments and that they work
-one test to prove that we force type on output arguments of a contract method


services/processor/native/call.go
```diff
@@ -43,211 +48,54 @@ func processMethodCall(executionContextId primitives.ExecutionContextId, contrac
 	return contractOutputArgs, contractOutputErr, err
 }
 
-var bigIntType = reflect.TypeOf(big.NewInt(0))
-
-func prepareMethodInputArgsForCall(methodInstance types.MethodInstance, args *protocol.ArgumentArray, functionNameForErrors string) ([]reflect.Value, error) {
-	res := []reflect.Value{}
+func verifyMethodInputArgs(methodInstance types.MethodInstance, functionNameForErrors string, args []interface{}) ([]reflect.Value, error) {
```
electricmonk: make sure that we have tests for:
- variadic with correct type
- variadic with incorrect type 
- variadic with both correct and incorrect types
- too many args
- not enough args

services/processor/native/test/harness.go
```diff
@@ -110,9 +112,9 @@ func (h *harness) expectSdkCallMadeWithAddressGetCaller(returnAddress []byte) {
 			input.OperationName == native.SDK_OPERATION_NAME_ADDRESS &&
 			input.MethodName == "getCallerAddress"
 	}
-
+	outputArgs, _ := protocol.ArgumentsFromNatives(builders.VarsToSlice(returnAddress)) // err ignored on purpose
```
electricmonk: why on purpose?

test/builders/arguments.go
```diff
@@ -7,60 +7,19 @@
...
-func Arguments(args ...interface{}) (res []*protocol.Argument) {
-	res = []*protocol.Argument{}
-	builders := ArgumentsBuilders(args...)
-	for _, builder := range builders {
-		res = append(res, builder.Build())
-	}
-	return
-}
-
+// Don't use this func in new tests, use the protocol.ArgumentArrayFromNatives and deal with err
```
electricmonk: use `// Deprecated`

services/processor/native/test/harness.go
noambergIL: because in many of the tests we "hard code" create this with a known type (string or []byte) it is unlikely that we will stop supporting this type so the error is unlikely to happen. 
i didn't want to panic in a case that is really unlikely.

electricmonk: so explain this in the comment

services/processor/native/call.go
noambergIL: all are present

PR #1437, opened 2019-11-06
Add tests for lean helix metric update

### LINKED ISSUE #1350: Add test that verifies lean helix view metric is being updated
**Is your feature request related to a problem? Please describe.**

See https://github.com/orbs-network/orbs-network-go/issues/1348
This can be generalized to: how to write a test that verifies a certain metric is being updated.

A clear and concise description of what the problem is. E.g, I'm always frustrated when [...]

**Describe the solution you'd like**

A test that deterministically triggers the behavior that *should* update the metric, and that fails if the metric is not updated.

vistra:
Fixes #1350 

A component test for the lean helix service that makes sure metrics are updated on elections and on commit.

Added two tests:
1. `TestMetricsAreUpdatedOnElectionTrigger` -  runs the service with a tiny election timeout and waits for an update on metrics`ConsensusAlgo.LeanHelix.CurrentElection.Number`, `ConsensusAlgo.LeanHelix.TimeSinceLastElection.Millis` and `ConsensusAlgo.LeanHelix.CurrentLeaderMemberId.Number`.

2. `TestMetricsAreUpdatedOnCommit` - triggers two commits (by feeding LH with consensus messages) and waits for an update on `ConsensusAlgo.LeanHelix.LastCommitted.TimeNano` and `ConsensusAlgo.LeanHelix.TimeSinceLastCommit.Millis`.
Two commits are required for `ConsensusAlgo.LeanHelix.TimeSinceLastCommit.Millis` to update.

A small infrastructure change to support it - added a Get() method to the in-memory metric registry, for the test to obtain metrics that are registered by the service internally.

Also made a small fix to `metric.Gauge` - added missing `AtomicLoad()`s


IdoZilberberg:
@vistra This is a big PR, can you elaborate what it does? Also it's better to have infrastructure changes (such as adding Get() with atomic ops, etc.) on a separate PR.

vistra:
@IdoZilberberg edited

electricmonk:
@ronnno @IdoZilberberg please review
@vistra you have test failures

vistra:
@ronnno @IdoZilberberg this is ready for review, who volunteers? :)

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+		const firstTermHeight = syncedBlockHeight + 1
+		const view = 0
+		blockPair := builders.BlockPair().WithHeight(firstTermHeight).WithTransactions(1).WithEmptyLeanHelixBlockProof().Build()
+
+		h.handlePreprepareMessage(ctx, blockPair, firstTermHeight, view, leaderNodeIndex)
+
+		commitMessages := []*interfaces.CommitMessage{}
+		for i := 0; i < h.networkSize(); i++ {
+			cmsg := h.handleCommitMessage(ctx, blockPair, firstTermHeight, view, randomSeed, i)
```
ronnno: protocol question - are we allowed to commit a block without receiving a single prepare message for it? I don't know if there is a safety issue here, since 2f commit messages will only arrive after sufficient prepare messages were sent in the real network with f +1 honest nodes. but maybe ask @gadcl if there are no view change scenarios where this may be dangerous?

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+			commitMessages = append(commitMessages, cmsg)
+		}
+
+		metrics := h.getMetrics()
+
+		// At this point the first block should be committed, lastCommitTime should update
+		require.True(t, test.Eventually(1*time.Second, func() bool {
+			return metrics.lastCommittedTime.Value() != 0
```
ronnno: can we require a reasonable diff from the current time - even 1 hour diff will make sure the value isn't 1 or FFFFF...

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+		}), "expected lastCommittedTime metric not to update on first commit")
+
+		// timeSinceLastCommitMillis will NOT update because this is the first commit
+		require.True(t, strings.Contains(metrics.timeSinceLastCommitMillis.String(), "samples=0"), "expected lastCommittedTime to not update on first commit")
+
+		// Starting the second term
```
ronnno: is there an easy refactor here to extract the common calls in both terms under a harness/driver method?

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+		h.handlePreprepareMessage(ctx, secondBlockPair, secondTermHeight, view, leaderNodeIndex)
+
+		for i := 0; i < h.networkSize(); i++ {
+			h.handleCommitMessage(ctx, secondBlockPair, secondTermHeight, view, randomSeed, i)
+		}
+
+		// A second commit should take place now and this time timeSinceLastCommitMillis should update
+
+		require.True(t, test.Eventually(1*time.Second, func() bool {
```
ronnno: can we check that `lastCommittedTime` increased in this step?

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+			matched, err := regexp.MatchString("samples=[1-9]", metrics.timeSinceLastCommitMillis.String())
+			if err != nil {
+				panic(err)
+			}
+			return matched
+		}), "expected timeSinceLastCommitMillis metric to update")
+	})
```
ronnno: consider adding a third term which does everything identically except it omits the final commit message. then require that no changes to `lastCommittedTime ` are made and that `timeSinceLastCommitMillis ` increases with the clock?

being the third step it would also justify the refactor i asked for in the comment to line 100 by the so called "rule of three"

instrumentation/metric/gauge.go
```diff
@@ -63,7 +63,7 @@ func (g *Gauge) UpdateUInt32(i int32) {
 }
 
 func (g *Gauge) Value() int64 {
-	return g.value
+	return atomic.LoadInt64(&g.value)
```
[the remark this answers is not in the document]
vistra: Yes, this is actually an unrelated fix

instrumentation/metric/registry.go
```diff
@@ -76,6 +77,17 @@ type inMemoryRegistry struct {
 	}
 }
 
+func (r *inMemoryRegistry) Get(metricName string) metric {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	for _, metric := range r.mu.metrics {
```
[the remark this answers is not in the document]
vistra: Agree, but since this pattern was also used in `.Remove()` and `.register()`, and since I don't see a reason for an intensive usage, I decided to keep it simple and use this pattern here as well.

services/consensusalgo/leanhelixconsensus/test/audit_mode_test.go
```diff
@@ -22,7 +22,7 @@ import (
 func TestHandleBlockConsensus_ExecutesBlocksYoungerThanThreshold_AndModeIsVerify(t *testing.T) {
 	with.Concurrency(t, func(ctx context.Context, parent *with.ConcurrencyHarness) {
 
-		h := newLeanHelixServiceHarness(5*time.Minute).start(parent, ctx)
+		h := newLeanHelixServiceHarness(5*time.Minute, time.Hour).start(parent, ctx)
```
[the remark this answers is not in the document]
vistra: Agreed

services/consensusalgo/leanhelixconsensus/test/harness.go
```diff
@@ -9,38 +9,59 @@ package test
...
+	consensus                     *leanhelixconsensus.Service
+	gossip                        *gossiptopics.MockLeanHelix
+	blockStorage                  *services.MockBlockStorage
+	consensusContext              *services.MockConsensusContext
+	instanceId                    lhprimitives.InstanceId
+	auditBlocksYoungerThan        time.Duration
+	consensusRoundTimeoutInterval time.Duration
+	metricRegistry                metric.Registry
+	logger                        log.Logger
+	T                             testing.TB
```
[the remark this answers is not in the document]
vistra: Yes

services/consensusalgo/leanhelixconsensus/test/harness.go
```diff
@@ -58,19 +79,29 @@ func (h *harness) resetAndApplyMockDefaults() {
...
+	sgnr, err := signer.New(cfg)
+	require.NoError(h.T, err)
 
-	h.consensus = leanhelixconsensus.NewLeanHelixConsensusAlgo(ctx, h.gossip, h.blockStorage, h.consensusContext, signer, parent.Logger, cfg, registry)
+	h.consensus = leanhelixconsensus.NewLeanHelixConsensusAlgo(ctx, h.gossip, h.blockStorage, h.consensusContext, sgnr, parent.Logger, cfg, h.metricRegistry)
 	parent.Supervise(h.consensus)
 	return h
 }
 
+func (h *harness) getMetrics() *metrics {
```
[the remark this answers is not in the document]
vistra: why isn't it functional? it returns the exact same metric objects used by the service, I think.

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
vistra: Good question. I also don't think there is a safety issue here, but will ask Gad. I think we can push this PR regardless, because if there's an issue here it would require an unrelated LH fix.

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
vistra: Good idea

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
vistra: I preferred not to over complicate the harness for a single test, but maybe we can do something local for this test.

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
vistra: I can also add this check in the first or second term right before emitting the final commit. So no three yet :)

btw - timeSinceLastCommit is currently only updated as a result of a commit, so it will not increase with the clock. Not sure if this was the intended behaviour..

services/consensusalgo/leanhelixconsensus/test/harness.go
```diff
@@ -114,3 +145,78 @@ func (h *harness) expectNeverToProposeABlock() {
...
+	})
+	require.NoError(h.T, err, "expected request ordering committee to succeed")
+	return out
+}
+
+func (h *harness) networkSize() int {
+	return NETWORK_SIZE
+}
+
+func (h *harness) myNodeIndex() int {
```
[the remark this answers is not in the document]
vistra: Agreed

services/consensusalgo/leanhelixconsensus/test/harness.go
```diff
@@ -83,12 +114,12 @@ func (h *harness) getCommitteeWithNodeIndexAsLeader(nodeIndex int) []primitives.
 	return res
 }
 
-func (h *harness) beLastInCommittee() {
-	h.expectConsensusContextRequestOrderingCommittee(1)
+func (h *harness) dontBeFirstInCommitee() {
+	h.expectConsensusContextRequestOrderingCommittee((h.myNodeIndex() + 1) % NETWORK_SIZE)
```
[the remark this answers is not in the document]
vistra: I did this to better reflect what the (previously called) `beLastInCommittee` was doing - which was to make you the second in the committee. But since it was dependent on the node index, I could only commit to it not being the first.

I agree that the naming can be better - will go with the simple `leaderInFirstView` and `validatorInFirstView`

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+		const syncedBlockHeight = 1
+		h.handleBlockSync(ctx, syncedBlockHeight)
+
+		orderedCommittee := h.requestOrderingCommittee(ctx)
+		orderedComitteeNodeIndicies := keys.NodeAddressesForTestsToIndexes(orderedCommittee.NodeAddresses)
+
+		leaderNodeIndex := orderedComitteeNodeIndicies[0] // leader for view 0 is always the first member in the ordered committee
+
+		prevBlockProof := protocol.BlockProofReader(nil) // the previous block (from sync) had no proof
```
electricmonk: this is noise, doesn't pertain to the test itself. Can you extract block generation to a named function or to the harness?

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
electricmonk: extract to a local func

services/consensusalgo/leanhelixconsensus/test/harness.go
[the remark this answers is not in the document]
ronnno: I thought it returned a clone that you cannot use to report changes...

services/consensusalgo/leanhelixconsensus/test/metrics_test.go
ronnno: `timeSinceLastCommit ` - the name is misleading. since it's a histogram it means that this is the actual intended way to measure it. but then maybe a different name such as `timeDeltaC2C` or `elapsedTimeToCommit` or `timeElapsedBetweenCommits`

PR #1439, opened 2019-11-07
improve error message on compilation timeout

### LINKED ISSUE #1387: Contract compilation is killed prematurely
I observe that sometimes a contract fails to compile with the following error:

```
could not build a shared object: error building go source: signal: killed, go build output: 
```

Instead of fixing the issue, the tests have a workaround that just keeps querying the contract until it succeeds.

Relevant builds: https://circleci.com/gh/orbs-network/orbs-network-go/20511

amir-arad:
also added tests to native compiler

fixes #1387


services/processor/native/adapter/native_compiler.go
```diff
@@ -202,6 +202,14 @@ func runGoCommand(ctx context.Context, workDir string, cmdArgs ...string) ([]byt
 		// "GOGC=off", (this improves compilation time by a small factor)
 	}
 	out, err := cmd.CombinedOutput()
+	if err != nil {
+		if deadline, ok := ctx.Deadline(); ok {
+			timeout := deadline.Sub(time.Now())
```
golangcibot: S1024: should use time.Until instead of t.Sub(time.Now()) (from `gosimple`)

services/processor/native/adapter/test/compiler_contract_test.go
```diff
@@ -87,7 +87,7 @@ type compilerContractHarness struct {
 
 func aNativeCompiler(t *testing.T, logger log.Logger) *compilerContractHarness {
 	tmpDir := test.CreateTempDirForTest(t)
```
ronnno: we're missing a `defer os.RemoveAll(tmpDir)`

services/processor/native/adapter/native_compiler_test.go
```diff
@@ -23,6 +26,69 @@ import (
...
+	})
+}
+
+func TestCompileTimeout(t *testing.T) {
+	with.Logging(t, func(parent *with.LoggingHarness) {
+		if testing.Short() {
+			t.Skip("Skipping compilation of contracts in short mode")
+		}
+		// give the test one minute timeout to compile
+		ctx, cancel := context.WithTimeout(context.Background(), 2000*time.Millisecond)
```
ronnno: why 2 seconds? it should be very low value, no?

services/processor/native/adapter/native_compiler_test.go
```diff
@@ -23,6 +26,69 @@ import (
...
+
+		tmpDir := test.CreateTempDirForTest(t)
+		defer os.RemoveAll(tmpDir)
+		cfg := &adapterTest.HardcodedConfig{ArtifactPath: tmpDir}
+		compiler := NewNativeCompiler(cfg, parent.Logger, metric.NewRegistry())
+
+		code := string(contracts.NativeSourceCodeForCounter(contracts.MOCK_COUNTER_CONTRACT_START_FROM))
+		_, err := compiler.Compile(ctx, code)
+		t.Log("Error : ", err)
+		require.Error(t, err, "compile should fail")
```
ronnno: don't you want to test that the error is descriptive?

services/processor/native/adapter/native_compiler_test.go
electricmonk: also, the comment above is copy-pasted from previous test, please remove
also, instead of `2000 * time.Millisecond` use `2 * time.Second`

services/processor/native/adapter/native_compiler.go
```diff
@@ -202,6 +202,14 @@ func runGoCommand(ctx context.Context, workDir string, cmdArgs ...string) ([]byt
 		// "GOGC=off", (this improves compilation time by a small factor)
 	}
 	out, err := cmd.CombinedOutput()
+	if err != nil {
+		if deadline, ok := ctx.Deadline(); ok {
+			timeout := time.Until(deadline)
+			if timeout < 0 {
+				err = errors.Wrap(err, "after context deadline exceeded")
+			}
+		}
```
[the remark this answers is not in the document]
amir-arad: > Why not just check `ctx.Err()`?

I am handling a specific case where context is timed out. `ctx.Err()` may be not nill in other cases.

The default message here was the text representation of `err`,  which was "signal: killed". I wanted to help the reader in the case of context timeout.

services/processor/native/adapter/native_compiler_test.go
amir-arad: > why 2 seconds? it should be very low value, no?
if it's too low, the compile command never starts. this is a best-effort, tested-on-my-machine timing to kill the command  while it executes

services/processor/native/adapter/test/config.go
```diff
@@ -0,0 +1,13 @@
+package test
+
+type HardcodedConfig struct {
```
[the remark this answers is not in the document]
amir-arad: this was extracted as-is from compiler_contract_test.go. I'll gladly go over it. excellent constructor suggestion.

as a side note, I'm mostly indifferent to fine-tuning of naming in such local scopes of simple code. It tends to trigger more debate that it's worth.

services/processor/native/adapter/native_compiler.go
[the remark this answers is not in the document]
ronnno: I see you found a way to combine both error messages, nice. just to reiterate - you can reliably detect a timeout by checking for `ctx.Err() == context.DeadlineExceeded` it is however a little stinky, because it a little too specific. however, it is documented and apparent from the code that this condition will be true iff a timeout happened...

services/processor/native/adapter/native_compiler_test.go
ronnno: if the test passes when the process takes longer than 2 seconds to launch then it's fine, otherwise we have a flaky test, no?

services/processor/native/adapter/native_compiler.go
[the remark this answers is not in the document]
amir-arad: thanks

PR #1440, opened 2019-11-07
fix trigger issue (1355) with empty signer 

### LINKED ISSUE #1355: Triggers transactions always fails because of missing signer
**Describe the bug**

Trigger has check inside the contract code for empty signer, However sdk fails if signer is empty (sdk doesn't know its trigger).

Short term : remove the check from _Trigger.trigger
Long term: add a schema ? add node signature to the tx ?

**Reproduce*
(look for in log for "result=EXECUTION_RESULT_ERROR_SMART_CONTRACT error=transaction is not signed by any Signer transaction-or-query" in any acceptance or e2e but make sure config is set to trigger on

noambergIL:
empty signer is now translated to a 20 byte of 0s in the sdk.
fix bug in validate_block_results that didn't pass the block proposer to the tx processing.
move trigger contract to dir without _
remove all comments from trigger.

fixes #1355


services/virtualmachine/signer_address.go
```diff
@@ -13,7 +13,12 @@ import (
 	"github.com/pkg/errors"
 )
 
+var EmptySignerAddress = []byte{0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0}
```
electricmonk: didn't we agree on some other number?

gadcl:
@noambergIL I took a look at the failure. It reflects two problems: (1)isolation - triggered methods should not fail the entire Trigger transaction.  (2) initial state of "elected validators" is empty, not handled in _Committee contract.  
Let's talk on Sunday

electricmonk:
Re: isolation - the current design is "all-or-nothing"; a contract-to-contract call that fails will fail the whole transaction. We'll need to come up with a way to signify if a service call is "fire and forget" or "transactional". Maybe even a different sdk call...

services/virtualmachine/signer_address.go
noambergIL: asked again and they don't really care

noambergIL:
Shai's case is a special case where no election ever happened so no committee exists in the state of _Elections.  It is now a valid case to have 0 validators ... then we just emit event.

noambergIL:
Added test to prove that network works in Committee Contract mode and also not in committee contract mode.

PR #1442, opened 2019-11-10
Fix benchmark consensus sync race

vistra:
There is a race when a node is updated with a new block concurrently through block sync and consensus. The consensus block sync handler will attempt to update the last committed block, but will find that the it doesn't match the current block and return with an error. Block sync in return will output an error to the log, and tests fail.

The solution suggested here is to not treat this as an error - it is a valid flow. Instead, add an info log and ignore.

### DIFF
--- a/services/consensusalgo/benchmarkconsensus/common.go
+++ b/services/consensusalgo/benchmarkconsensus/common.go
@@ -30,18 +30,20 @@ func (s *Service) getLastCommittedBlock() (primitives.BlockHeight, *protocol.Blo
 	return s.lastCommittedBlockUnderMutex.TransactionsBlock.Header.BlockHeight(), s.lastCommittedBlockUnderMutex
 }
 
-func (s *Service) setLastCommittedBlock(blockPair *protocol.BlockPairContainer, expectedLastCommittedBlockBefore *protocol.BlockPairContainer) error {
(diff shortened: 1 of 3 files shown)


services/consensusalgo/benchmarkconsensus/common.go
```diff
@@ -30,18 +30,20 @@ func (s *Service) getLastCommittedBlock() (primitives.BlockHeight, *protocol.Blo
...
 }
 
-func (s *Service) setLastCommittedBlock(blockPair *protocol.BlockPairContainer, expectedLastCommittedBlockBefore *protocol.BlockPairContainer) error {
+func (s *Service) setLastCommittedBlockIfPreviousBlockMatches(blockPair *protocol.BlockPairContainer, expectedLastCommittedBlockBefore *protocol.BlockPairContainer) bool {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 
 	if s.lastCommittedBlockUnderMutex != expectedLastCommittedBlockBefore {
-		return errors.New("aborting shared state update due to inconsistency")
+		s.logger.Info("setLastCommittedBlockIfPreviousBlockMatches ignoring block, expected previous block different than actual - probably updated concurrently from a different flow")
```
ronnno: we can return an error if the height of the expected last committed block is higher than the actual last committed block

PR #1446, opened 2019-11-10
Add/marvin/hook

IdoZilberberg:
This will run a 5-minute long test against the previous `master`.
Presently, the test will *not* run against the version currently being merged to `master`, but one before it.

At this point this is good enough, as we want to fill the DB with job results and give us a feel on how this works.

Next - to apply the test against the testnet with the current commit, edit `.circleci/config.yml` - move the job `call_marvin` after the job `deploy_testnet` and make it dependent on `deploy_testnet` with the following syntax:
```
- call_marvin:
          requires:
            - deploy_testnet
```

The reason we don't do this right now is because `deploy_testnet` is not working right, so `call_marvin` would not work right either.


.circleci/config.yml
```diff
@@ -118,6 +118,14 @@ jobs:
       - store_artifacts:
           path: _out
 
+  call_marvin:
+    docker:
+      - image: circleci/golang:1.12.9
+    resource_class: medium
```
electricmonk: why do we need any specific resource class here?

itamararjuan: Agree about the resource class its a simpe curl call no need for any specific compute demands

Issue #1328, opened 2019-09-16
TestDeploymentOfNativeContract assumes transaction returns immediately and flakes


electricmonk:
Either we need to increase the test timeout, or make the test busy-wait for tx to complete, or investigate what happens here

```
harness.go:206: started (+0.000s)
harness.go:206: first block committed (+0.021s)
harness.go:206: send deploy - start (+0.000s)
harness.go:167: 
Error Trace:	harness.go:167
            				deploy_native_test.go:40
Error:      	Expected nil, but got: failed to deploy native contract: http status 202 Accepted
Test:       	TestDeploymentOfNativeContract/TestDeploymentOfNativeContract_#1
Messages:   	expected deploy contract to succeed
```

https://circleci.com/gh/orbs-network/orbs-network-go/19212?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

PR #1449, opened 2019-11-11
Fix race with query immediately following contract deployment

vistra:
RunQuery returns an error when the contract is not deployed. This will suppress the error when polling for the correct block height. If polling is unsuccessful the last error will be returned.

Another related fix - when submitting a transaction synchronously, the SDK treats HTTP 202 (indicating the transaction is still pending) as an error. Added this as an exception in the test flow.

### DIFF
--- a/test/e2e/harness.go
+++ b/test/e2e/harness.go
@@ -70,12 +70,16 @@ func NewAppHarness() *Harness {
 	}
 }
 
+func isNotHttp202Error(err error) bool {
+	return !strings.Contains(err.Error(), "http status 202 Accepted")
+}
+
 func (h *Harness) DeployContract(from *keys.Ed25519KeyPair, contractName string, processorType orbsClient.ProcessorType, code ...[]byte) (*codec.TransactionResponse, error) {
 	timeoutDuration := 15 * time.Second
 	beginTime := time.Now()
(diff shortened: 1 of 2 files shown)


netoneko:
@OdedWx @electricmonk this is an issue that external developers will inevitably face, there needs to be a proper solution to this

electricmonk:
@netoneko I agree and this is a long running argument I have with @OdedWx and @talkol. Sadly I'm currently on the losing side of this argument, maybe your experience with customers can help tip the needle.

test/e2e/harness.go
```diff
@@ -75,7 +75,7 @@ func (h *Harness) DeployContract(from *keys.Ed25519KeyPair, contractName string,
 	beginTime := time.Now()
 
 	txOut, txId, err := h.SendDeployTransaction(from.PublicKey(), from.PrivateKey(), contractName, processorType, code...)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "http status 202 Accepted") { // TODO the SDK treats HTTP 202 as an error
```
electricmonk: wrap in a function `isNonError`?

vistra:
fixes #1328

Issue #1454, opened 2019-11-12
Marvin test should run only after successful e2e


IdoZilberberg:
**Describe the bug**
Presently Marvin test (stress test on testnet) runs before a new testnet is deployed.
This causes the test to run on the previous version of master on testnet, before the new version is deployed.

Move the job `call_marvin` as a dependency of `e2e_testnet_leanhelix`

Also fix test timings to:
`{"tpm":60, "duration_sec":720, "client_timeout_sec": 180}`

PR #1455, opened 2019-11-12
Issue 1454: Update marvin dependencies in CircleCI; change its timing

IdoZilberberg:
- [ ] CircleCI: Run marvin test only after successful e2e (which itself executes only after deployment of new version to testnet).

- [ ] Fix timings to 60tpm, 720s test duration, 180s max client duration (so total of 3 client cycles and slack notifications during test)


IdoZilberberg:
Closes #1454

PR #1456, opened 2019-11-13
Add support for external js build

netoneko:
Enables JS plugin release process: https://github.com/orbs-network/orbs-network-javascript-plugin/pull/10

The only significant difference is setting is ability to set `$BUILD_FLAG` from outside the script and preventing failure when trying to get last commit message related to `$CIRCLE_SHA1` - I am building from a different repo, so `$CIRCLE_SHA1` does not exist in this repo.


services/processor/javascript/test/e2e/deploy_javascript_test.go
```diff
@@ -25,7 +25,9 @@ func TestDeploymentOfJavascriptContract(t *testing.T) {
 		t.Skip("Skipping E2E tests in short mode")
 	}
 
-	t.Skip("not implemented")
+	if !jsEnabled() {
```
electricmonk: why not put the whole file under build flag?

netoneko: Because there are two different things here: under `javascript` build flag primarily runs the dummy plugin test. It ensures that the interface is doing ok and does not require an actual JS plugin.

To enable tests against actual plugin, I introduce `JS_ENABLED` variable in my tests.

netoneko:
@electricmonk `BUILD_FLAG` was defined as empty variable in the build script and it was impossible to set it from outside the script. I need to make the change it to pass `javascript` build tag to Docker.

netoneko:
@electricmonk @itamararjuan any updates on this?

PR #1461, opened 2019-11-14
[WIP] Allow a prepared node to commit in older view

ronnno:
upgraded lean-helix-go to v0.2.5

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/olekukonko/tablewriter v0.0.2 // indirect
 	github.com/orbs-network/go-mock v1.1.0
 	github.com/orbs-network/govnr v0.2.0
-	github.com/orbs-network/lean-helix-go v0.2.4
+	github.com/orbs-network/lean-helix-go v0.2.5
 	github.com/orbs-network/membuffers v0.3.6
 	github.com/orbs-network/orbs-client-sdk-go v0.12.0
 	github.com/orbs-network/orbs-contract-sdk v1.4.0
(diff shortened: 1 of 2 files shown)


electricmonk:
what's the problem this aims to fix?

vistra:
This allows a prepared node to participate in older views, to avoid
unneccesary view progression.

electricmonk:
do we have evidence of this happening?

vistra:
Not that I know of

ronnno:
Its theoretical and should affect tests much more than production config. However, we also don't have evidence of the context cancellation bug affecting tests performance and 202 issues.

electricmonk:
> However, we also don't have evidence of the context cancellation bug affecting tests performance and 202 issues.
I think that we do - @IdoZilberberg did some log analysis that originally pinpointed the context cancellation bug

I'm not against a solution - I'm wondering what provoked this fix at this point in time

electricmonk:
And specifically, I see no test in https://github.com/orbs-network/lean-helix-go/pull/87. I'd like to stress that fixing bugs without reproducing them in tests is an exception, not the rule, and often leads to regression.

Please submit a test in LH to reproduce the bug along with the PR.

ronnno:
I agree we should add a test to formulate logic behind the PR.

just to clarify what  orbs-network/lean-helix-go#87 is about:
1. this PR doesn't solve a bug. 
1. It removes a constraint (an `if` clause in the code) which was an over optimization, redundant, cluttered, and in addition, caused a minor inefficiency in runtime (a few network round trips that could be avoided). 
1. the PR is still under review and discussion. the reason it got to PR now was to gain CI results to support its correctness.
1. It's a subtle point in interpreting the spec, but the main added value is cleanliness and clarity by removing some lines of code.
1. this is a tweak to orbs-network/lean-helix-go#86 and ideally would be included in it had we noticed it before

ronnno:
> I think that we do - @IdoZilberberg did some log analysis that originally pinpointed the context cancellation bug

While we saw the cancellation issue in some cases - I did try to establish a strong correlation to the 202 flakiness issue by looking at examples in CI and could not find such correlation.
Having said that, the issue resolved in  orbs-network/lean-helix-go#86 was significant and clearly had potential to raise flakiness rates.

ronnno:
PR went stale, the included changes were already merged into master i a separate PR

PR #1463, opened 2019-11-14
added missing indirect dependency

ronnno:
dependency comes from Prometheus


IdoZilberberg:
@ronnno did you have to add this manually? doesn't go modules do it automatically?

ronnno:
its added automagically but needs to be checked in

PR #1466, opened 2019-11-17
Remove tests with benchmark consensus from e2e

vistra:
As they are redundant given benchmark consensus is tested on acceptance tests, and that we also run with LH.


electricmonk:
@itamararjuan FYI - we this was decided during sync today

itamararjuan:
Nice to know

PR #1467, opened 2019-11-17
Adding tac-like functionality for Mac OS based system so that test.sh would work

itamararjuan:
running test.sh locally would not bring meaningful output due to missing tac command on Darwin based systems. 

BSD-style system have most functionality of `tac` covered by `tail -r`

This PR deals with checking whether we are running on a Mac OS system and if so sets up a local function inside our test.common bash library for testing so that the system can have access to a similar `tac` functionality.

### DIFF
--- a/test.common.sh
+++ b/test.common.sh
@@ -3,6 +3,13 @@
 NIGHTLY=0
 mkdir -p _out
 
+OS=$(uname -s)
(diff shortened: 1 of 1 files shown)


electricmonk:
@itamararjuan did you verify that it works on your machine?

PR #1476, opened 2019-11-19
Elections logs change %v to %x

### LINKED ISSUE #1171: logs in election contract output non-human-readable messages
**Describe the bug**

when there is an execution error, for example:
`delegate from [246 126 179 99 53 28 97 125 82 183 212 28 119 252 74 255 249 169 126 167] to [247 174 98 44 119 208 88 15 2 188 178 249 35 128 214 30 63 110 70 108] with block-height 7778218 and tx-index 31 failed since current delegation is from block-height 7778218 and tx-index 31
`
the addresses are outputted as byte arrays (probably %v) and not hex strings (%x) which would make a lot more sense to us puny humans.

noambergIL:
so that addresses will be printed in hex.
fixes #1171


noambergIL:
closed due to weird go.sum issue.

PR #1478, opened 2019-11-20
 add bootstrap logger

### LINKED ISSUE #1473: log panics on node startup
When a node panics during its startup sequence (in the main goroutine), the panic will not be logged to the logger (disk, logz.io, console). This can happen if a config validation error occurs, or if the node fails to instantiate any of its subsystems (for instance, block storage could not open the blocks file).

Add panic recover and logging to `NewNode()` in order to resolve this.

amir-arad:
log all panics on main

fixes #1473

### DIFF
--- a/instrumentation/logger.go
+++ b/instrumentation/logger.go
@@ -13,6 +13,24 @@ import (
 	"os"
 )
 
+func GetBootstrapCrashLogger() log.Logger {
+	path := "./orbs-network-bootstrap.log"
+
+	logFile, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		panic(err)
+	}
+
+	fileWriter := log.NewTruncatingFileWriter(logFile)
+	outputs := []log.Output{
+		log.NewFormattingOutput(fileWriter, log.NewHumanReadableFormatter()),
+		log.NewFormattingOutput(os.Stdout, log.NewHumanReadableFormatter()),
+		log.NewFormattingOutput(os.Stderr, log.NewHumanReadableFormatter()),
+	}
+
+	return log.GetLogger().WithOutput(outputs...)
+}
+
 func GetLogger(path string, silent bool, cfg config.NodeConfig) log.Logger {
 	if path == "" {
 		path = "./orbs-network.log"
--- a/main.go
+++ b/main.go
@@ -14,39 +14,56 @@ import (
 	"github.com/orbs-network/orbs-network-go/config"
 	"github.com/orbs-network/orbs-network-go/instrumentation"
 	"github.com/orbs-network/orbs-network-go/synchronization/supervised"
+	"github.com/orbs-network/scribe/log"
+	"github.com/pkg/errors"
 	"os"
 )
 
 func main() {
-	httpAddress := flag.String("listen", ":8080", "ip address and port for http server")
-	silentLog := flag.Bool("silent", false, "disable output to stdout")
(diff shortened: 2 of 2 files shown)


instrumentation/logger.go
```diff
@@ -13,6 +13,24 @@ import (
...
+	path := "./orbs-network-bootstrap.log"
+
+	logFile, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		panic(err)
+	}
+
+	fileWriter := log.NewTruncatingFileWriter(logFile)
+	outputs := []log.Output{
+		log.NewFormattingOutput(fileWriter, log.NewHumanReadableFormatter()),
```
electricmonk: why write to a file at all? docker stdout seems to be the direction we're going to, no?

instrumentation/logger.go
```diff
@@ -13,6 +13,24 @@ import (
...
+		log.NewFormattingOutput(os.Stdout, log.NewHumanReadableFormatter()),
+		log.NewFormattingOutput(os.Stderr, log.NewHumanReadableFormatter()),
```
electricmonk: do we need both? aren't both being mapped to docker logs?

main.go
```diff
@@ -14,39 +14,56 @@ import (
...
+		node = bootstrap.NewNode(
+			cfg,
+			logger,
+		)
 
+		supervised.NewShutdownListener(logger, node).ListenToOSShutdownSignal()
+	}()
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Error("unexpected error during runtime", log.Error(errors.Errorf("unknown error: %v", r)))
```
electricmonk: as this goroutine is blocked on `WaitUntilShutdown`, I don't see how any panic can be thrown from it. It's not as if this `defer func()` is a "catch all exception handler".

main.go
```diff
@@ -14,39 +14,56 @@ import (
...
 )
 
 func main() {
-	httpAddress := flag.String("listen", ":8080", "ip address and port for http server")
-	silentLog := flag.Bool("silent", false, "disable output to stdout")
-	pathToLog := flag.String("log", "", "path/to/node.log")
-	version := flag.Bool("version", false, "returns information about version")
+	logger := instrumentation.GetBootstrapCrashLogger()
+	var node *bootstrap.Node
+	func() { // context of bootstrap crash logging
```
electricmonk: why do everything in an anonymous func? what's the point?

instrumentation/logger.go
amir-arad: Regardless of our plans, we are currently writing to a file, so bootstrap errors should currently go to that file as well.

instrumentation/logger.go
amir-arad: I'm not sure exactly what output goes where in prod. and I suspect it's a good thing not being able to assume that in the bootstrap phase (only prior to logger init): there should be 0 lines, so write them to all directions. better have 3 copies of a never-should-happen critical error, than none.

main.go
amir-arad: Agree.

errors happen: out of memory, seg faults, sneaky edge case bugs in 3rd party libraries etc. If they happen in the main routine, I'd like to log them to the right output (configured in the logger) before the server dies.

main.go
amir-arad: the anonymous function ends when the bootstrap is complete. it is the context of the deferred `recover()` for bootstrap errors

PR #1481, opened 2019-11-24
Use base docker for faster builds

vistra:
- Use a base docker that contains prefetched go modules, by cloning orbs-network-go into /tmp and running a build.
- The base docker is updated once a day as part of the nightly
- Enable the `docker_layer_caching` flag in CircleCI
- For e2e, use a dockerized version of go-junit-report, so it won't have to be installed on every build


netoneko:
Anyone who doesn't have access to our private ECR won't be able to build Docker images.

On top of that, I don't really see how is it going to reduce the build times in any significant way.

vistra:
The base image is optional, so we can also have dockerfiles that use the base golang image instead. Or we can make the base image public.
As for build times, It seems to speed up the e2e test flow by 2-3 minutes, but it's still WIP.

vistra:
@itamararjuan ready now, please review

.circleci/config.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
   flakiness:
     docker:
-      - image: itamararjuan/golang_with_node_aws
+      - image: circleci/golang:1.12.9
```
itamararjuan: How does that work for you? 
Looking at our flakiness script, it attempts to install some stuff with Node.js
So I don't think this can work unless you re-add steps that install Node.js on the Go container.

Can you share a bit more information as to why you are making this change?

netoneko:
If you're running the build inside of `circleci/golang:1.12.9` you could just invoke `build-binaries` and then go straight for the second part of `./docker/build.sh` where we use `Dockerfile.export` and so on, skipping a costly process of uploading Docker build context.

.circleci/config.yml
vistra: I think it only attempts to use NPM when running nightly.
Will revert anyway.

vistra:
@netoneko Is that a partial implementation of your idea?
https://github.com/orbs-network/orbs-network-go/compare/feature/test-build-binaries-step?expand=1
And you say it took 1:37 to build? Very interesting.

netoneko:
@vistra yes

electricmonk:
@vistra @itamararjuan is this still relevant?

vistra:
No, closing

PR #1482, opened 2019-11-24
Update go.mod.template

ronnno:
Removed external libraries from non-experimental version of go.mod.template

### DIFF
--- a/docker/build/build.sh
+++ b/docker/build/build.sh
@@ -55,13 +55,15 @@ docker build -f ./docker/build/Dockerfile.signer -t orbs:signer .
 # Builds experimental features (extra libraries)
 if [[ $CIRCLE_TAG != v* ]] ;
 then
+    export ORBS_EXPERIMENTAL="true"
+    
     # We use an experimental go.mod for these builds
     rm -rf _bin && mkdir -p _bin _dockerbuild
     rm -f ./_dockerbuild/go.mod.template
     SDK_VERSION=$(cat go.mod | grep orbs-contract-sdk | awk '{print $2}')
     cp ./docker/build/go.mod.template.experimental ./_dockerbuild/go.mod.t
     sed "s/SDK_VER/$SDK_VERSION/g" _dockerbuild/go.mod.t > _dockerbuild/go.mod.template
-
+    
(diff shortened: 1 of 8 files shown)


electricmonk:
@itamararjuan @netoneko apparently we were adding the external libraries to releases as well. It was added by @netoneko in 9e75d2f4d7e974b4114d6cc0bd1e0802c97e36dd and approved by @itamararjuan in #1386. This is a mistake, the external libraries should *only* be part of experimental builds (non-releases). 

Removing the import from `go.mod.template` causes @netoneko's new E2E to fail, as only the Docker-based E2E uses `go.mod.template.experimental`.

@itamararjuan please make the appropriate changes with @netoneko so that we can merge this PR.

itamararjuan:
The code has been changed to support building and testing experimental libraries only when needed with the effort to make the decision of when Orbs is about to be ran in "experimental" test spirit as concise as possible and boiled down to 1 variable which is set during the docker build stage (called `ORBS_EXPERIMENTAL`

@electricmonk  please review carefully to make sure we do not introduce any more hidden leakages

Thanks @netoneko  for the catch about the experimental test which would of fail future tag pushes. This has now also been addressed with the newly offered solution.

netoneko:
Always a pleasure.

docker/build/build.sh
```diff
@@ -55,27 +55,15 @@ docker build -f ./docker/build/Dockerfile.signer -t orbs:signer .
 # Builds experimental features (extra libraries)
 if [[ $CIRCLE_TAG != v* ]] ;
 then
+    export ORBS_EXPERIMENTAL="true"
+    
     # We use an experimental go.mod for these builds
     rm -rf _bin && mkdir -p _bin _dockerbuild
     rm -f ./_dockerbuild/go.mod.template
     SDK_VERSION=$(cat go.mod | grep orbs-contract-sdk | awk '{print $2}')
     cp ./docker/build/go.mod.template.experimental ./_dockerbuild/go.mod.t
```
electricmonk: can we instead of duplicating the logic of creating the go.mod file from template (actually overriding it), just set the `ORBS_EXPERIMENTAL` flag given the `CIRCLE_TAG` value, then choose either the experimental or production template according to the value of `ORBS_EXPERIMENTAL`?

PR #1484, opened 2019-11-27
Add metrics for connection latency per node (requested by @gadcl)

netoneko:
Gad wants to measure latency per connection to be able to tweak LH timings.

@IdoZilberberg can you please advise if this is good enough for Kibana?

### DIFF
--- a/services/gossip/adapter/tcp/outgoing_connection.go
+++ b/services/gossip/adapter/tcp/outgoing_connection.go
@@ -28,11 +28,16 @@ type timingsConfig interface {
 	GossipConnectionKeepAliveInterval() time.Duration
 }
 
+type singleConnectionMetric struct {
+	latency *metric.Histogram
+}
+
 type outgoingConnection struct {
 	logger         log.Logger
 	metricRegistry metric.Registry
 	config         timingsConfig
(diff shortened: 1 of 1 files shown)


electricmonk:
@IdoZilberberg it's a pity you merged this so quickly. 
@netoneko I'm afraid this might not be good enough, as keepalives are only being sent if no other message was sent.
a more complete solution would be to measure latency also on the client send function

vistra:
Also, I think this method of measuring latency is problematic as it doesn't take kernel buffers into account, see https://stackoverflow.com/questions/5407182/blocking-sockets-when-exactly-does-send-return

electricmonk:
following @vistra's comments on #1486 - 
@IdoZilberberg please revert this PR

IdoZilberberg:
@electricmonk The purpose of this metric is to facilitate config tweaking by measuring
latency between nodes on a "clean" vchain - without traffic. Within that specific use-case, it should be able to provide us the right metrics data.
I agree it needs to be changed so the measurement will not be affected by traffic, or be more accurate as mentioned by @vistra but for the time being, it looks better than nothing, I would not just revert it.
What say you?

electricmonk:
As long as consensus and block sync is running, there’s traffic and the metric will not be reported

IdoZilberberg:
Reverted by #1488 .
^ @electricmonk @vistra

PR #1489, opened 2019-11-28
E2E to use LeanHelix

ronnno:
Replaces the consensus algorithm used in e2e tests to LH. 
Preparation for full Elections E2E test


electricmonk:
@vistra please review

services/processor/native/adapter/native_compiler_test.go
```diff
@@ -32,18 +32,30 @@ func TestCompileValidContract(t *testing.T) {
...
+		for i := 0; i < concurrentCount; i++ {
+			go func() {
+				err := compilationOfMockCounterContract(t, parent)
+				doneChan <- err
+			}()
+		}
+
+		for i := 0; i < concurrentCount; i++ {
+			ctxErr := <-doneChan
+			require.Nil(t, ctxErr, "expected concurrent contract compilation to succeed")
```
vistra: require.NoError?

services/processor/native/adapter/native_compiler_test.go
```diff
@@ -32,18 +32,30 @@ func TestCompileValidContract(t *testing.T) {
...
 
-		code := string(contracts.NativeSourceCodeForCounter(contracts.MOCK_COUNTER_CONTRACT_START_FROM))
-		_, err := compiler.Compile(ctx, code)
-		t.Log("Error : ", err)
-		require.NoError(t, err, "compile should not fail")
+		const concurrentCount = 5
+		doneChan := make(chan interface{}, concurrentCount)
+		for i := 0; i < concurrentCount; i++ {
+			go func() {
+				err := compilationOfMockCounterContract(t, parent)
```
vistra: You can `require.NoError(err)` here and just pass nil in the channel, then read all the nils in line 55, a bit more to my taste :)

test/e2e/_data/main/rebuild_canned_blocks_main.go
```diff
@@ -24,9 +24,13 @@ func main() {
 
 	clearBlocksFile()
 
-	n := e2e.NewInProcessE2EAppNetwork(42, e2e.NewLoggerRandomer())
+	n := e2e.NewInProcessE2EAppNetwork(42, e2e.NewLoggerRandomer(), "/tmp/dummy")
```
vistra: Is this going to write on `/tmp/dummy`? If so, I think it should use a more specific name.
Also, can you elaborate a bit on why this was suddenly necessary?

test/e2e/block_storage_init_test.go
```diff
@@ -7,33 +7,29 @@
...
-	const expectedBlocks = 500
 	if testing.Short() {
 		t.Skip("Skipping E2E tests in short mode")
 	}
 
-	runMultipleTimes(t, func(t *testing.T) {
-		h := NewAppHarness()
+	h := NewAppHarness()
+
+	if h.envSupportsLoadingWithCannedBlocksFile() {
```
vistra: May be a bit cleaner to have a `h.expectedInitialBlockHeight` setting which is 500 when using the canned file and 0 otherwise, and have the test be identical for both flows.

bootstrap/node.go
```diff
@@ -43,7 +43,7 @@ func NewNode(nodeConfig config.NodeConfig, logger log.Logger) *Node {
 	ctx, ctxCancel := context.WithCancel(context.Background())
 	config.NewValidator(logger).ValidateMainNode(nodeConfig) // this will panic if config does not pass validation
 
-	nodeLogger := logger.WithTags(log.Node(nodeConfig.NodeAddress().String()))
+	nodeLogger := logger.WithTags(log.Node(nodeConfig.NodeAddress().String()), log.Uint32("virtual-chain-id", uint32(nodeConfig.VirtualChainId())))
```
netoneko: We already log virtual chain id (and there is special log field for it): https://github.com/orbs-network/orbs-network-go/blob/36e21b654a918ecc6a31b121022a9991897ef267/instrumentation/logger.go#L64

electricmonk: interesting
I remembered we used to, but when I looked at the log I couldn't find vcid anywhere
perhaps vcid is defined at the `main.go` level and the in-process E2Es don't use `main.go`

@ronnno Let's use the `logfields.VirtualChainId` tag but remove its usage from main, leaving it at the `node.go` level instead.

services/processor/native/adapter/native_compiler_test.go
ronnno: `require.NoError` calls `runtime.Goexit()` when an error is detected. this means we have to be extra careful in how we signal into `doneChan` and `defer` the signaling so it is executed in case an error is detected.

I accept the comment, but it is a little fragile. This is why I normally prefer to `require` only from the top goroutine

test/e2e/_data/main/rebuild_canned_blocks_main.go
ronnno: replacing with `""`, since they are similar in this context.

ronnno: This parameter was introduced by @netoneko while working on javascript processor. In this PR I needed to replace the `blocks` file so I did the minimal changes to make this `main` compile

test/e2e/block_storage_init_test.go
ronnno: I don't want to hide away the fact that the test does nothing (in some cases). Prefer this limitation to be explicit

bootstrap/node.go
ronnno: Please review...

electricmonk:
@ronnno merge this?

PR #1492, opened 2019-12-02
Direct transport roundtrip metric

vistra:
This adds a roundtrip metric to direct transport. 
This is done by replying with an ACK (4 bytes) to each received message, and timing the difference.
This is backward compatible as ACKs are entirely optional.


services/gossip/adapter/tcp/transmission_time_queue.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (c *TransmissionTimeQueue) push(sentAt time.Time) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	select {
+	case c.times <- sentAt:
+		return
+	default:
+		_ = <-c.times
```
golangcibot: S1005: '_ = <-ch' can be simplified to '<-ch' (from `gosimple`)

services/gossip/adapter/tcp/direct_transport_roundtip_metric_test.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/orbs-network/orbs-spec/types/go/primitives"
+	"github.com/orbs-network/orbs-spec/types/go/protocol/gossipmessages"
+	"github.com/stretchr/testify/require"
+	"net"
+	"regexp"
+	"testing"
+	"time"
+)
+
+func TestDirectTransport_SupportsBroadcastTransmissions(t *testing.T) {
```
electricmonk: isn't this a duplication from `direct_transport_test`?

services/gossip/adapter/tcp/direct_transport_roundtip_metric_test.go
```diff
@@ -0,0 +1,277 @@
...
+	var err error
+	test.Eventually(time.Second, func() bool {
+		server, err = startSilentServer(node.transport.GetServerPort())
+		return err == nil
+	})
+	require.NoError(t, err, "silent server failed to start")
+	return server
+}
+
+func TestDirectTransport_RoundtripMetricReflectsLatency(t *testing.T) {
```
electricmonk: for this test, wouldn't it be easier to create a single client that talks to a server implemented in this test? why even use the big harness?

services/gossip/adapter/tcp/direct_transport_roundtip_metric_test.go
vistra: It was moved here accidentally, will move back

services/gossip/adapter/tcp/direct_transport_roundtip_metric_test.go
vistra: Moved some tests to outgoing_connection_roundtrip_metric_test.go

services/gossip/adapter/tcp/direct_transport_roundtip_metric_test.go
```diff
@@ -0,0 +1,58 @@
...
+			"expected all nodes to have peers added",
+			func(node *nodeHarness) bool { return len(node.transport.outgoingConnections.activeConnections) > 0 },
+			node1, node2)
+
+		waitForAllNodesToSatisfy(t,
+			"expected all outgoing queues to become enabled after topology change",
+			func(node *nodeHarness) bool { return node.transport.allOutgoingQueuesEnabled() },
+			node1, node2)
+
+		roundtrip1to2 := node1.metricRegistry.Get(fmt.Sprintf("Gossip.OutgoingConnection.Roundtrip.%s", node2.address.String()[:6]))
```
electricmonk: you could create a global `func(NodeAddress) string` to create the metric name given the address, it would make it a bit less fragile

vistra:
@electricmonk The essence of this change is really simple, if the code doesn't reflect that then maybe some refactoring can help. 
Yes, let's discuss it.

netoneko:
@electricmonk we do need this metric. It was requested by @gadcl so he can calibrate LH timings.

The alternative would be to run special infrastructure that measures latency between AWS regions but it still won't give us production picture and won't account for self-hosted solutions that our validators use.

electricmonk:
as discussed f2f with @vistra, we're not moving forwards with this PR, as it's not clear whether the added value is justified. if @gadcl can come up with a reasonable approximation based on publicly available latency statistics, it might be preferable to making such severe changes to the codebase.

PR #1494, opened 2019-12-02
[WIP] rearrange circleci workflow

electricmonk:


netoneko:
Why do you have to split between node build and gamma build if they use essentially the same components and are compiled from the same source?

netoneko:
I suggest you take a look to things like that too: https://github.com/orbs-network/orbs-network-go/blob/destructure/e2e/and/ci/reorg/docker/test/test.sh#L15

To my knowledge we don't use node here, so there is no reason to run or install it.

electricmonk:
@netoneko re: install node.js - I agree, we need to clean up unused stuff in our CI folder, docker folder, and scripts in general

re: separating gamma and node builds - for the sake of parallelization and running further jobs that depend only on gamma or only on the node docker containers

PR #1510, opened 2019-12-23
Feature/poll/marvin continued

IdoZilberberg:
Hey, I've completed an endurance test on this code <br />for a duration of 0 hours 5 minutes 0 seconds <br />Key metrics from the run (compared to master):<br />*119* transactions processed successfully (119)<br />*0* transactions failed (0)<br />average service time: *98ms* (98ms)<br />Memory consumption: *32.13MB* *-%21.8*(39.13MB)<br />


netoneko:
There are 38 files in `.circleci` folder (excluding config.yml). Maybe Marvin-related code could be installed from the Marvin repo to keep things more tidy?

IdoZilberberg:
Most of the files contain wiring code between CircleCI jobs.
Most of the logic is actually in a separate repo (https://github.com/orbs-network/marvin).
We can consolidate the wiring into less scripts and move some of the remaining logic to marvin repo.
This can be done in subsequent PRs - the current PR is large enough as it is.
^ @netoneko

.circleci/config.yml
```diff
@@ -150,11 +150,13 @@ jobs:
   marvin_tests:
     docker:
       - image: itamararjuan/golang_with_node_aws
+    environment:
+      MARVIN_URL: "http://ec2-34-222-245-15.us-west-2.compute.amazonaws.com:4567"
```
electricmonk: why is it here and named thus if you define a different var `MARVIN_ORCHESTRATOR_URL` at the project level?

.circleci/marvin-analyze.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -e
```
electricmonk: why are you separating to `.sh` and `.js` file? 
I think it might be much clearer if everything was in javascript and `circle.yml` would just run node scripts.

.circleci/marvin-analyze.js
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env node
```
electricmonk: I think maybe create a subdir `marvin` inside `.circleci` and put all marvin scripts inside it. And the package.json file.

.circleci/marvin-reporter.js
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env node
+
+const execSync = require('child_process').execSync;
```
electricmonk: have you considered using the circleci Slack Orb instead?

.circleci/package.json
```diff
@@ -9,6 +9,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@orbs-network/judge-dredd": "^0.0.6",
```
electricmonk: what's this?

IdoZilberberg: This is the external package that marvin-analysis calls - it JUDGES whether the test passed or failed. Its logic is external to this repo because we want to be able to update it separately.

.circleci/marvin-reporter.js
IdoZilberberg: This would mean replacing a single `curl` command inside the script (all the rest is required logic that will remain anyway) with a call to an Orb I know nothing about.
I prefer using curl. This is also much easier to test & debug locally from command line.

.circleci/marvin-analyze.js
IdoZilberberg: Ok, but only in the next PR (which is almost ready), because doing it here will make a mess of merging between Itamar, me, master, and the next PR.

.circleci/marvin-analyze.sh
IdoZilberberg: The distinction here is having environment plumbing happen in sh files (which are most concise and readable for that purpose) and having logic in the js files (handling json, calling the external `judge-dredd` package for test analysis, etc.)

You're right that there are 2 files per step, but I feel that merging the 2 types would make it all far less readable and testable.

.circleci/config.yml
IdoZilberberg: Mistake - removed from config.yml.

IdoZilberberg:
@electricmonk See my comments

electricmonk:
I'm ok with postponing merging shell scripts and js files to a subsequent PR. But do it as soon as this one is merged. We have a complete mess with CI scripts as it is and there's absolutely NO point in making it messier.

PR #1518, opened 2020-01-13
Fix Github comment

IdoZilberberg:


electricmonk:
shouldn't we see Marvin commenting on this PR?

PR #1521, opened 2020-01-14
Fix state storage memory leak

### LINKED ISSUE #1520: memory allocated in disk read or tcp (network) read is never released
we see in the production network that the memory allocated at the tcp parsing and readChunk of the persistence layer is retained at these levels

this means that the lower levels of state for example, do not copy only what they need, but use the original message and retain the original slices

for example in the below, its wrong that memory is retained at this level and not the service, it means that we are leaking the entire message (state just needs key and value, not the entire message object, and it will store them in a map most likely)

```
github.com/orbs-network/orbs-network-go/services/blockstorage/adapter/filesystem.readChunk
services/blockstorage/adapter/filesystem/codec.go

  Total:    973.99MB   973.99MB (flat, cum) 75.41%
    517            .          .           	// check budget 
    518            .          .           	if budget.limit < budget.bytesRead+int(chunkLength) { 
    519            .          .           		return nil, fmt.Errorf("invalid block. size exceeds limit (%d)", budget.limit) 
    520            .          .           	} 
    521            .          .            
    522     973.99MB   973.99MB           	chunk := make([]byte, chunkLength) 
    523            .          .           	n, err := io.ReadFull(reader, chunk) 
    524            .          .           	if err != nil { 
    525            .          .           		return nil, err 
    526            .          .           	} 
    527            .          .            
```

netoneko:
The nature of the leak is that any block that contained the state diff was held up in memory because the state diffs were referencing the underlying byte array.

After applying the fix, testing on 3.9Gb blocks file for vchain 1960000 revealed that allocated memory after starting up went down from `878mb` to `478mb` which is `45.5%` improvement.

This PR partially resolves #1520

### DIFF
--- a/services/statestorage/adapter/memory/memory_persistence.go
+++ b/services/statestorage/adapter/memory/memory_persistence.go
@@ -50,7 +50,7 @@ func NewStatePersistence(metricFactory metric.Factory) *InMemoryStatePersistence
 		fullState:  adapter.ChainState{},
 		height:     0,
 		ts:         0,
-		proposer:    []byte{},
+		proposer:   []byte{},
 		merkleRoot: merkleRoot,
 	}
 }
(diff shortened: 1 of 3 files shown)


netoneko:
Before and after flame graphs of memory allocation:

Screen Shot 2020-01-14 at 18 06 11
Screen Shot 2020-01-14 at 18 06 28

PR #1526, opened 2020-01-27
Block sync - retry with smaller chunk sizes 

### LINKED ISSUE #1514: BlockSync - a node too far behind may stop syncing forever
A block sync request asks for a range of blocks, of a size defined in the node configuration.  If the chunk size (in bytes) is larger than the transport queue size (currently 20mb), the transmission of the chunk is guaranteed to fail. 

The chunk size configuration is defined by number of blocks per chunk, not by number of bytes. So the chunk size in bytes may vary significantly depending of the size of the blocks in the range. 

If the configuration allows, a node too far behind may ask for a chunk too-large to fit the send queue buffer, and the sync will never advance.

root cause of #1511

vistra:
In case of an error arising form the block chunk being too big, the block-sync server will retry sending the blocks with smaller and smaller chunk sizes (diving by 2 each time).

Fixes #1514

### DIFF
--- a/services/blockstorage/block_sync_server.go
+++ b/services/blockstorage/block_sync_server.go
@@ -110,7 +110,7 @@ func (s *Service) sourceHandleBlockSyncRequest(ctx context.Context, message *gos
 		return errors.Wrap(err, "block sync failed reading from block persistence")
 	}
 
-	var chunkSize = len(blocks)
+	chunkSize := len(blocks)
 	for {
 		lastAvailableBlockHeight := firstAvailableBlockHeight + primitives.BlockHeight(chunkSize) - 1
 		logger.Info("sending blocks to another node via block sync",
--- a/services/blockstorage/block_sync_server.go
+++ b/services/blockstorage/block_sync_server.go
@@ -9,6 +9,7 @@ package blockstorage
 import (
 	"context"
 	"github.com/orbs-network/orbs-network-go/instrumentation/trace"
+	"github.com/orbs-network/orbs-network-go/services/gossip"
 	"github.com/orbs-network/orbs-spec/types/go/primitives"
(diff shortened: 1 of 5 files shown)


services/blockstorage/block_sync_server.go
```diff
@@ -104,31 +105,55 @@ func (s *Service) sourceHandleBlockSyncRequest(ctx context.Context, message *gos
...
-			SignedChunkRange: (&gossipmessages.BlockSyncRangeBuilder{
-				BlockType:                blockType,
-				FirstBlockHeight:         firstAvailableBlockHeight,
-				LastBlockHeight:          lastAvailableBlockHeight,
-				LastCommittedBlockHeight: lastCommittedBlockHeight,
-			}).Build(),
-			BlockPairs: blocks,
-		},
+	// Must be an error indicating chunk-too-big
+	if !gossip.IsChunkTooBigError(lastError) {
```
electricmonk: why?

services/blockstorage/block_sync_server.go
```diff
@@ -104,31 +105,55 @@ func (s *Service) sourceHandleBlockSyncRequest(ctx context.Context, message *gos
...
+		panic(errors.Wrap(err, "unreachable code - expected chunk-too-big error"))
```
electricmonk: this is one of these ironic errors you end up finding in logs. You actually panic the goroutine here - is this what you intended to do?

services/blockstorage/block_sync_server.go
vistra: If it's a different error, it should have already returned at line 144.

services/blockstorage/block_sync_server.go
vistra: Yes - this is an assertion, if it actually executes then something in the state is fundamentally broken

PR #1527, opened 2020-01-28
new reputation calculation with a weighted non-repeated draw

noambergIL:
New algo to calculate reputation :
1) rep is now 0-6, (instead of 4-10) and rep 1 indicates a slash of half your chance to be leader.
2) algo is based on weighted draw
3) getAllX doesn't represent a committee order anymore.
4) input is assumed in consensus order so that the genesis has to first sort its config node addresses

### DIFF
--- a/services/processor/native/repository/Committee/committee.go
+++ b/services/processor/native/repository/Committee/committee.go
@@ -28,42 +28,41 @@ func _getOrderedCommitteeForAddresses(addresses []byte) [][]byte {
 	return _getOrderedCommitteeArray(addressArray)
 }
 
-func _getOrderedCommitteeArray(addresses [][]byte) [][]byte {
-	return _orderList(addresses, _generateSeed())
+func getNextOrderedCommittee() [][]byte {
+	addresses := _split(_getElectedValidators())
+	return _orderList(addresses, _generateSeed(env.GetBlockHeight()+1))
 }
 
-func _generateSeed() []byte {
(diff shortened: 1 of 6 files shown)


services/processor/native/repository/Committee/committee.go
```diff
@@ -7,14 +7,14 @@
...
-	"bytes"
 	"encoding/binary"
 	"github.com/orbs-network/orbs-contract-sdk/go/sdk/v1/env"
 	"github.com/orbs-network/orbs-network-go/crypto/hash"
-	"math"
-	"sort"
 )
 
+/**
+ * This function is meant ot be used via the callsystemcontract func ... it will not give same result when used with RunQuery
```
netoneko: Why won't it give the same result?

noambergIL: When you use callSystemContract its a very special case of runQuery where the state is done for the last committed block but the env is for current suggested block. Since we use the current block number as our random seed this changes the order of the committee ... this is all done on purpose.

electricmonk: @noambergIL maybe we can add some way to check whether we're running from `callSystemContract` and panic otherwise?

netoneko: @noambergIL cool, thanks!

noambergIL: @electricmonk i don't think we want to add an env that tells you if you are runquery/sendtx/callsystem.

PR #1551, opened 2020-04-19
Expose version during startup

### LINKED ISSUE #1547: Metrics reporting is incomplete during startup
**Describe the bug**

Most of the metrics we rely upon (even trivial like version and block storage height) are not being reported for a while until the node processes a lot of blocks.

**Steps To Reproduce**

Start a node with a huge amount of blocks.

**Expected behavior**

Report most of the metrics couple of seconds after http server responds to an endpoint.

**Logs**

Metrics from http://35.183.225.241/vchains/1960000/metrics during startup:

```json
{
  "Gossip.IncomingConnection.Active.Count": {
    "Name": "Gossip.IncomingConnection.Active.Count",
    "Value": 1
  },
  "Gossip.IncomingConnection.ListeningOnTCPPortErrors.Count": {
    "Name": "Gossip.IncomingConnection.ListeningOnTCPPortErrors.Count",
    "Value": 0
  },
  "Gossip.IncomingConnection.ListeningOnTCPPortSuccess.Count": {
    "Name": "Gossip.IncomingConnection.ListeningOnTCPPortSuccess.Count",
    "Value": 2
  },
  "Gossip.IncomingConnection.TransportErrors.Count": {
    "Name": "Gossip.IncomingConnection.TransportErrors.Count",
    "Value": 1
  },
  "Gossip.OutgoingConnection.Active.Count": {
    "Name": "Gossip.OutgoingConnection.Active.Count",
    "Value": 0
  },
  "Gossip.OutgoingConnection.KeepaliveErrors.Count": {
    "Name": "Gossip.OutgoingConnection.KeepaliveErrors.Count",
    "Value": 0
  },
  "Gossip.OutgoingConnection.MessageSize.Bytes": {
    "Name": "Gossip.OutgoingConnection.MessageSize.Bytes",
    "Min": 0,
    "P50": 0,
    "P95": 0,
    "P99": 0,
    "Max": 0,
    "Avg": 0,
    "Samples": 0
  },
  "Gossip.OutgoingConnection.SendErrors.Count": {
    "Name": "Gossip.OutgoingConnection.SendErrors.Count",
    "Value": 0
  },
  "Gossip.OutgoingConnection.SendQueueErrors.Count": {
    "Name": "Gossip.OutgoingConnection.SendQueueErrors.Count",
    "Value": 0
  }
}
```

netoneko:
fixes #1547


bootstrap/inmemory/network.go
```diff
@@ -72,6 +72,7 @@ func NewNetworkWithNumOfNodes(validators map[string]config.ValidatorNode, nodeOr
 		validatorNode := validators[address.KeyForMap()]
 		cfg := cfgTemplate.ForNode(address, privateKeys[address.KeyForMap()])
 		metricRegistry := metric.NewRegistry()
+		metric.RegisterConfigIndicators(metricRegistry, cfg)
```
ronnno, line 75: why is this needed? here?

netoneko: Because this is the place where the initialization happens.

I want to report the version right after we start the http server, because otherwise it will be blocked by initialization of other services and it will take some time.

netoneko: I guess adding the values directly to `dumpMetricsAsJSON` should be good enough, however, it would be also a good idea to split metrics and config values that we want to expose

ronnno: I agree about splitting runtime metrics and config values - i wanted to suggest this too but I don't know who else is relying on these "metrics" so if we remove them form the metrics url it may break something... for instance, do we show version number or public addresses in graphana?

ronnno: for instance, the status page shows the version.. that's one place that is relying on the metrics page for versions

ronnno: maybe the simplest way is to leave what existed as it was - and create another config handler in the HTTP server for your needs?

netoneko: The simplest way is indeed to leave what we have now but update `dumpMetricsAsJSON`.

We do not show text values in grafana because Prometheus rightfully does not believe that text is a metric and they simply won't allow it.

bootstrap/node.go
```diff
@@ -36,9 +36,15 @@ type Node struct {
...
 
-func getMetricRegistry(nodeConfig config.NodeConfig) metric.Registry {
+func GetMetricRegistry(nodeConfig config.NodeConfig) metric.Registry {
 	metricRegistry := metric.NewRegistry().WithVirtualChainId(nodeConfig.VirtualChainId()).WithNodeAddress(nodeConfig.NodeAddress())
 
+	version := config.GetVersion()
+
+	metricRegistry.NewText("Version.Semantic", version.Semantic)
+	metricRegistry.NewText("Version.Commit", version.Commit)
+	metricRegistry.NewText("Node.Address", nodeConfig.NodeAddress().String())
```
ronnno, line 46: this is redundant

ronnno: see line 40

netoneko: Fixed

PR #1554, opened 2020-04-27
feature/headers state machine

gadcl:


gadcl:
updated block sync design

PR #1556, opened 2020-05-03
update topology is from spec, also fix upper case Service in gossip

noambergIL:


services/gossip/adapter/peers.go
```diff
@@ -20,6 +25,14 @@ type gossipPeer struct {
 	hexOrbsAddress string
 }
 
+func NewGossipPeers(servicePeers []*services.GossipPeer) GossipPeers {
```
ronnno, line 28: its not very pretty that you are using your own `GossipPeer` and a type of the same name from the spec. now I also understand why you named the variable `servicePeers` which is a little cryptic (imho).

This will be more elegant if the local type was called `GossipEndpoint[Desc]` or something like that, then you can also change the param name to something simple like `peers` it makes a better separation of the packages imho..

services/management/adapter/memory_provider.go
```diff
@@ -66,6 +68,20 @@ func getCommitteeFromConfig(config MemoryConfig) []primitives.NodeAddress {
 	return committee
 }
 
+func getTopologyFromConfig(cfg MemoryConfig) []*services.GossipPeer {
+	peers := cfg.GossipPeers()
+	topology := make([]*services.GossipPeer, 0, len(peers))
+	for _, peer := range peers {
+		if nodeAddress, err := hex.DecodeString(peer.HexOrbsAddress()); err != nil {
```
ronnno, line 75: reverse the condition and throw away empty clause?

noambergIL: added log and comment

PR #1560, opened 2020-05-26
draft - work in progress

gadcl:
work in progress

### DIFF
--- a/services/consensuscontext/committee.go
+++ b/services/consensuscontext/committee.go
@@ -42,3 +42,30 @@ func (s *service) RequestValidationCommittee(ctx context.Context, input *service
 	}
 	return res, nil
 }
+
+
+func (s *service) RequestBlockProofOrderingCommittee(ctx context.Context, input *services.RequestBlockProofCommitteeInput) (*services.RequestBlockProofCommitteeOutput, error) {
+	return s.RequestBlockProofValidationCommittee(ctx, input)
+}
+
+func (s *service) RequestBlockProofValidationCommittee(ctx context.Context, input *services.RequestBlockProofCommitteeInput) (*services.RequestBlockProofCommitteeOutput, error) {
+	out, err := s.management.GetCommittee(ctx, &services.GetCommitteeInput{Reference:input.PrevBlockReferenceTime})
+	if err != nil {
+		return nil, err
+	}
+	committee := out.Members
+	s.logger.Info("committee size", log.Int("elected-validators-count", len(committee)), log.Uint32("max-committee-size", input.MaxCommitteeSize), trace.LogFieldFrom(ctx))
+
+	s.metrics.committeeSize.Update(int64(len(committee)))
+	committeeStringArray := make([]string, len(committee))
+	for j, nodeAddress := range committee {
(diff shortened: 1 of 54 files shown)


services/consensuscontext/committee.go
```diff
@@ -42,3 +42,30 @@ func (s *service) RequestValidationCommittee(ctx context.Context, input *service
 	}
 	return res, nil
 }
+
+
+func (s *service) RequestBlockProofOrderingCommittee(ctx context.Context, input *services.RequestBlockProofCommitteeInput) (*services.RequestBlockProofCommitteeOutput, error) {
```
noambergIL, line 47: 1) i would add here a comment telling that this is different because it doesn't care about order just who is. (committee)
2) are we 100% sure that we always ask a committee available by ref ? management is only "per-day" ?

test/acceptance/committee_election_test.go
```diff
@@ -25,6 +26,13 @@ func TestCommitteeElections_OneElectionAndCheckReputationChanges(t *testing.T) {
 		WithManagementPollingInterval(20*time.Millisecond).
 		WithLogFilters(log.DiscardAll()).
 		WithConsensusAlgos(consensus.CONSENSUS_ALGO_TYPE_LEAN_HELIX).
```
noambergIL, line 28: i don't like this way of over-riding config ... there is config for acceptance ... this way of override i want to kill.

test/acceptance/deploy_native_test.go
```diff
@@ -11,13 +11,21 @@ import (
...
+	testKeys "github.com/orbs-network/orbs-network-go/test/crypto/keys"
 	"github.com/orbs-network/orbs-spec/types/go/protocol"
 	"github.com/stretchr/testify/require"
 	"testing"
 )
 
 func TestDeployNativeContract(t *testing.T) {
-	NewHarness().Start(t, func(t testing.TB, ctx context.Context, network *Network) {
+	NewHarness().
+		WithSetup(func(ctx context.Context, network *Network) {
```
noambergIL, line 22: consider if this setup should be the default in the harness instead of adding it to each test ?

test/acceptance/node_sync_test.go
```diff
@@ -20,19 +21,22 @@ import (
 // (could have mocked the whole thing)
 // Either test with Benchmark Consensus which is makes it easier to generate fake proofs, or use real recorded Lean Helix blocks
 func TestInterNodeBlockSync_WithBenchmarkConsensusBlocks(t *testing.T) {
-
 	NewHarness().
 		WithConsensusAlgos(consensus.CONSENSUS_ALGO_TYPE_BENCHMARK_CONSENSUS).
 		//WithLogFilters(log.ExcludeEntryPoint("BenchmarkConsensus.Tick")).
 		//AllowingErrors().
 		WithSetup(func(ctx context.Context, network *Network) {
+			newRefTime := GenerateNewManagementReferenceTime(0)
```
noambergIL, line 29: again ... make this committee the default ... in fact i wanted to move all the acceptance and e2e to use a new ctor  of memory provider with a committee (it now does read it from the config ... but you override it each time)

PR #1564, opened 2020-06-02
Feature/sync out of order storage

gadcl:
support in memory based persistent storage


test/acceptance/api_responses_test.go
```diff
@@ -31,7 +36,12 @@ func TestResponseForTransactionOnValidContract(t *testing.T) {
 }
 
 func TestResponseForTransactionOnContractNotDeployed(t *testing.T) {
-	NewHarness().Start(t, func(t testing.TB, parent context.Context, network *Network) {
+	NewHarness().WithSetup(func(ctx context.Context, network *Network) {
```
noambergIL, line 39: should not stay ... will add new managememt memory provider and set it in the harness ctor.

services/blockstorage/internodesync/state_processing_blocks.go
```diff
@@ -84,9 +100,84 @@ func (s *processingBlocksState) processState(ctx context.Context) syncState {
...
+		if nextBlock, _ := s.storage.GetBlock(firstBlock.TransactionsBlock.Header.BlockHeight() + 1); nextBlock != nil { // will verify hash pointer to block
+			// prepend
+			blocks = append([]*protocol.BlockPairContainer{nextBlock}, blocks...)
+		} else {
+			ref, err := s.management.GetCurrentReference(ctx, &services.GetCurrentReferenceInput{})
+			if err != nil {
+				s.logger.Error("management.GetCurrentReference should not return error", log.Error(err))
+				return err
+			}
+			if s.managementReferenceGrace > 0 { // check is enabled // TODO: due to testing incompatibility with current reference time and time.Now()
```
noambergIL, line 127: since we only seem to chane the hashptr ... why do we care that we are not current ?
also management grace ... this check is weird because now is POST finality (20 minutes) so we check that current time + grace < now ... so this is wrong (which makes me feel it is all very smelly) ... 

i feel the grace check here may be over-kill

PR #1567, opened 2020-06-07
Add healthcheck

netoneko:
Discussion: maybe polling on metrics is not the best thing to do and there should be a dedicated healthcheck endpoint.

But we definitely need to output some key metrics for the status page.


bootstrap/httpserver/status.go
```diff
@@ -0,0 +1,110 @@
...
+	Payload   Payload
+}
+
+func (s *HttpServer) getStatus(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	status := StatusResponse{
+		Timestamp: time.Now(),
+		Status:    s.getStatusWarningMessage(),
+		Payload: Payload{
```
noambergIL, line 40: this is not what we agreed on ... for now payload will be the current metrics. it will be change in the future

bootstrap/httpserver/status.go
```diff
@@ -0,0 +1,80 @@
...
+			s.logger.Error("could not retrieve metric", log.String("metric", name))
+		}
+	}()
+
+	rows := s.metricRegistry.Get(name).Export().LogRow()
+	value = rows[len(rows)-1].Int
+	return value
+}
+
+func (s *HttpServer) getStringValueFromMetrics(name string) (value string) {
```
ronnno, line 70: no one is calling this method. do we need it?

bootstrap/httpserver/status.go
```diff
@@ -0,0 +1,80 @@
...
+		maxTimeSinceLastBlock = 600000000
+	}
+	if s.getGaugeValueFromMetrics("ConsensusAlgo.LeanHelix.LastCommitted.TimeNano")+maxTimeSinceLastBlock <
+		time.Now().UnixNano() {
+		return "Last Successful Committed Block was too long ago"
+	}
+
+	if len(s.config.ManagementFilePath()) != 0 && s.config.ManagementPollingInterval() > 0 {
+		maxIntervalSinceLastSuccessfulManagementUpdate := int64(s.config.ManagementPollingInterval().Seconds()) * 20
+		if s.metricRegistry.Get("Management.Data.LastSuccessfulUpdateTime").Export().LogRow()[0].Value().(int64)+maxIntervalSinceLastSuccessfulManagementUpdate <
```
ronnno, line 49: in `getGaugeValueFromMetrics ` you protect against metric names that are not yet defined. could this line panic too? consider moving panic safety to this method?...

bootstrap/httpserver/status.go
netoneko: removed

bootstrap/httpserver/status.go
netoneko: I am using `getGaugeValueFromMetrics` now instead.

PR #1569, opened 2020-06-17
[DO NOT MERGE] Enable block storage init test

netoneko:
Enables block storage init test per @ronnno suggestion.

The test still fails on CI, needs additional investigation why.

### DIFF
--- a/docker/test/test.sh
+++ b/docker/test/test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -xe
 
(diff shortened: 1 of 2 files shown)


ronnno:
@netoneko, Should we ask @itamararjuan  to look into this when he gets some time?

ronnno:
@netoneko I believe the reason this is failing is that we need a different blocks file for lean-helix  - the block proofs will look different

netoneko:
@ronnno seems like a reasonable assumption, but when we run regular e2e we also copy the blocks in `file_utils.go` (or am I am missing something?)

PR #1570, opened 2020-06-17
Management Service - Historical Data + Grace and system-time update

noambergIL:


services/management/management.go
```diff
@@ -102,81 +112,168 @@ func NewManagement(parentCtx context.Context, config Config, provider Provider,
...
 }
 
-func (s *service) GetProtocolVersion(ctx context.Context, input *services.GetProtocolVersionInput) (*services.GetProtocolVersionOutput, error) {
-	return &services.GetProtocolVersionOutput{
-		ProtocolVersion: s.getProtocolVersion(input.Reference).Version,
-	}, nil
+func (s *service) tryGetHistoricData(ctx context.Context, referenceTime primitives.TimestampSeconds) (*VirtualChainManagementData, error) {
+	s.Lock()
+	defer s.Unlock()
+	if referenceTime < s.cachedHistoricData.StartPageReference || referenceTime > s.cachedHistoricData.EndPageReference {
```
ronnno, line 178: so the cache is of 1 item at most?

services/management/management.go
```diff
@@ -102,81 +112,168 @@ func NewManagement(parentCtx context.Context, config Config, provider Provider,
...
+		historic, err := s.provider.Get(ctx, referenceTime)
+		if err != nil {
+			return nil, errors.Wrapf(err, "management provider failed to get historic reference data for reference-time %d", referenceTime)
+		}
+		s.cachedHistoricData = historic
+		s.metrics.pageCachedStartRefTime.Update(int64(s.cachedHistoricData.StartPageReference))
+		s.metrics.pageCachedEndRefTime.Update(int64(s.cachedHistoricData.EndPageReference))
```
ronnno, line 185: do we want to count cache hits vs misses?

services/management/management.go
```diff
@@ -102,81 +112,168 @@ func NewManagement(parentCtx context.Context, config Config, provider Provider,
...
+
 	return &services.GetSubscriptionStatusOutput{
-		SubscriptionStatusIsActive: s.getSubscriptionStatus(input.Reference).IsActive,
+		SubscriptionStatusIsActive: getSubscriptionStatus(input.Reference, data.Subscriptions).IsActive,
+	}, nil
+}
+
+func getProtocolVersion(refTime primitives.TimestampSeconds, protocolVersions []ProtocolVersionTerm) *ProtocolVersionTerm {
+	i := len(protocolVersions) - 1
+	for ; i > 0 && refTime < protocolVersions[i].AsOfReference; i-- {
```
ronnno, line 245: are you sure the array covers all possible values? if the first element is not 0 you may return the first element even if refTime is before AsOfReference

services/management/management.go
```diff
@@ -213,32 +318,37 @@ func (s *service) initMetrics(metricFactory metric.Factory) {
...
 
 	currentRef := s.data.CurrentReference
 
 	s.metrics.currentRefTime.Update(int64(currentRef))
 	s.metrics.genesisRefTime.Update(int64(s.data.GenesisReference))
-	s.metrics.lasUpdateTime.Update(time.Now().Unix())
+	s.metrics.pageStartRefTime.Update(int64(s.data.StartPageReference))
+	s.metrics.pageEndRefTime.Update(int64(s.data.EndPageReference))
+	s.metrics.lastUpdateTime.Update(time.Now().Unix())
+	if isSuccessful {
```
ronnno, line 332: why not count the errors (in an else clause)?

services/management/adapter/memory_provider.go
```diff
@@ -83,21 +83,23 @@ func getTopologyFromConfig(cfg MemoryConfig, logger log.Logger) []*services.Goss
...
+		EndPageReference:   mp.currentReference,
+		CurrentTopology:    mp.topology,
+		Committees:         mp.committees,
+		Subscriptions:      mp.isSubscriptionActives,
+		ProtocolVersions:   mp.protocolVersions,
 	}, nil
 }
 
-// for acceptance tests
+// for acceptance tests use one or the other no support for cross "types" adding with respect to paging
```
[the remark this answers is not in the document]
noambergIL: memory provider is for tests and gamma - and it doesn't support complicated add functionality

services/management/management.go
```diff
@@ -102,81 +112,168 @@ func NewManagement(parentCtx context.Context, config Config, provider Provider,
 	return s
 }
 
+/*
+ * Current and Genesis Reference functions
+ */
+func getCurrentReference(systemRef primitives.TimestampSeconds, data *VirtualChainManagementData) primitives.TimestampSeconds {
+	if data.CurrentReference != 0 { // public management
```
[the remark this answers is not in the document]
noambergIL: You misunderstand. public = management generated from ethereum the current ref changes in monotonous. Private there is a file(s) that represent the management they change from time to time. the current is 0 is only way other than a config to signal . this is by Oded's design.

services/management/management.go
noambergIL: one page (page can is eigher thousands of events of a full day TBD)

services/management/management.go
noambergIL: that is why there is an importance of order.

noambergIL: if you have [ {asof : 50 ...}, {asof: 250 ...}, {asof 200 ..} ] the 250 one is unreachable by design. it signifies there was a version we decided to supersede  but we keep a record of it.

services/management/management.go
noambergIL: i don't care how many errors were. we really just need to know when is the last successfull the actual number is irrelevant.

services/management/management.go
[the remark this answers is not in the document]
ronnno: so "public management" means "public network"? it would be clearer to make the comment "public network"

services/management/management.go
ronnno: i understand - but in your example if someone asks about ref 10 you will return the answer that matches 50

ronnno: ?

services/management/management.go
ronnno: it would be good for maintenance / debugging

services/management/management.go
noambergIL: efficiency is for next version. the metrics page is already very full of stuff

services/management/management.go
noambergIL: depends on the page. if the page is from 0 to 300 then its ok to return
in theory the page should start from earliest event.

PR #1571, opened 2020-06-22
Sync batch exceeds outgoing connection capacity 

### LINKED ISSUE #1514: BlockSync - a node too far behind may stop syncing forever

ronnno:
resolves #1514

### DIFF
--- a/services/gossip/adapter/tcp/outgoing_connections.go
+++ b/services/gossip/adapter/tcp/outgoing_connections.go
@@ -121,10 +121,16 @@ func (c *outgoingConnections) disconnectAll(ctx context.Context, peersToDisconne
 	}
 }
 
+var DataExceedsCapacityError = errors.Errorf("Data exceeds allowed size %d", SEND_QUEUE_MAX_BYTES)
+
 // TODO(https://github.com/orbs-network/orbs-network-go/issues/182): we are not currently respecting any intents given in ctx (added in context refactor)
 func (c *outgoingConnections) send(ctx context.Context, data *adapter.TransportData) error {
 	c.RLock()
 	defer c.RUnlock()
+
+	if data.TotalSize() > SEND_QUEUE_MAX_BYTES {
+		return DataExceedsCapacityError
+	}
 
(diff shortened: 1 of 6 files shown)


services/gossip/adapter/tcp/outgoing_connections.go
```diff
@@ -121,11 +121,17 @@ func (c *outgoingConnections) disconnectAll(ctx context.Context, peersToDisconne
...
 }
 
+var DataExceedsCapacityError = errors.Errorf("Data exceeds allowed size %d", SEND_QUEUE_MAX_BYTES)
+
 // TODO(https://github.com/orbs-network/orbs-network-go/issues/182): we are not currently respecting any intents given in ctx (added in context refactor)
 func (c *outgoingConnections) send(ctx context.Context, data *adapter.TransportData) error {
 	c.RLock()
 	defer c.RUnlock()
 
+	if data.TotalSize() > SEND_QUEUE_MAX_BYTES {
```
vistra, line 131: There is a small unhandled edge case here - the send queue can be partially full, causing this condition to pass but the subsequent transmission to fail silently as before

ronnno: Not really - if we wanted to address the case where the queue is temporarily full the correct behavior would be to wait for to to clear with a timeout.  
Consider that in the time that passes between the first failure and the second attempt with half the batch size the queue can be emptied. So halving the batch size is simply not the correct fix for this edge case. We will risk starvation in this case - unlikely as it may be

PR #1576, opened 2020-07-02
adding the new boyar contract of how to run in v2

itamararjuan:
This PR aims to eliminate Boyar's knowledge of how the service runs
and introduces a new entrypoint which Boyar can call


entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash +x
+
+/opt/orbs/orbs-node $@ | multilog t s16777215 n3 '!tai64nlocal' /opt/orbs/logs 2>&1
```
netoneko, line 3: Please document what these parameters mean

ronnno: nice to have - allow override 16777215 and 3 using env vars?

PR #1578, opened 2020-07-05
refactoring fixes and tests

gadcl:


test/e2e/network.go
```diff
@@ -105,7 +105,7 @@ func bootstrapE2ENetwork(logFilePrefix string, virtualChainId primitives.Virtual
 				experimentalExternalProcessorPluginPath,
 			)
 
-		if deployBlocksFile {
+		if deployBlocksFile && i != METRICS_NODE_INDEX{
```
ronnno, line 108: add a comment here to say that we require the metrics node to sync his blocks from his peers?

test/e2e/block_sync_test.go
```diff
@@ -0,0 +1,37 @@
...
+package e2e
+
+import (
+	"github.com/orbs-network/orbs-network-go/test"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestBlockSyncRecover(t *testing.T) {
```
ronnno, line 16: there are some hidden assumptions here - that the node we get the metrics from is the node that did not receive his blocks file.. can you please  open an issue to make it clearer or add some comments?

test/acceptance/node_sync_test.go
```diff
@@ -14,20 +14,25 @@ import (
 	"github.com/orbs-network/orbs-spec/types/go/protocol/consensus"
 	"github.com/stretchr/testify/require"
 	"testing"
+	"time"
 )
 
+const DEFAULT_MANAGEMENT_REFTIME_ONSET = 100
```
ronnno, line 20: FAKE_BLOCKS_AGE_SECONDS?

PR #1583, opened 2020-07-07
if self not in topology don't connect to anyone, disconnect from all.

noambergIL:


services/gossip/adapter/tcp/outgoing_connections.go
```diff
@@ -76,12 +76,29 @@ func (c *outgoingConnections) WaitUntilShutdown(shutdownContext context.Context)
...
 
-// note that bgCtx MUST be a long-running background context - if it's a short lived context, the new connection will die as soon as
-// the context is done
-func (c *outgoingConnections) connectForever(bgCtx context.Context, peerNodeAddress string, peer adapter.TransportPeer) {
+func (c *outgoingConnections) updateTopology(bgCtx context.Context, newTopology adapter.TransportPeers) {
 	c.Lock()
 	defer c.Unlock()
+	// If not in topology disconnect from outer world
+	if _, isInNewTopology := newTopology[c.nodeAddress.KeyForMap()]; !isInNewTopology {
+		if len(c.peerTopology) != 0 {
```
ronnno, line 84: the second if is not necessary

instrumentation/metric/registry.go
```diff
@@ -85,39 +87,28 @@ func (r *inMemoryRegistry) Get(metricName string) metric {
...
 
 func (r *inMemoryRegistry) register(m metric) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	for _, existing := range r.mu.metrics {
-		if existing.Name() == m.Name() {
-			err := errors.Errorf("a metric with name %s is already registered", m.Name())
-			panic(err)
-		}
+	key := fmt.Sprintf("%p", m)
```
netoneko, line 105: Why not `metric.Key` or something like that? I think it would be more clear

instrumentation/metric/registry.go
```diff
@@ -85,39 +87,28 @@ func (r *inMemoryRegistry) Get(metricName string) metric {
...
+	if _, isMetricInMap := r.mu.metrics[key]; isMetricInMap {
```
netoneko, line 106:
```
if val, ok := dict["foo"]; ok {
    //do something here
}
```

noambergIL: i want to keep this panic for now.

instrumentation/metric/registry.go
noambergIL: i wanted to limit the change. if i add key to the "named metric" i have to change in 3 different specific ctors (gauge, histogram, rate) i just want it limited here. when we do a big change in future we'll change it.

ronnno: maybe add a comment explaining why the pointer is used. I agree it's the right choice here but it's pretty confusing for someone who does not know the story

instrumentation/metric/registry.go
```diff
@@ -85,39 +87,28 @@ func (r *inMemoryRegistry) Get(metricName string) metric {
 			return metric
 		}
 	}
-	panic(errors.Errorf("a metric with the name of %s is not registered", metricName))
+	return nil
```
ronnno, line 90: `Get` locks the map for a full scan. this is beyond the scope of this PR but consider opening an issue or a TODO comment to add a lookup map that will speed up these queries (you can write to the map in "register", and every time you get some item from that list check if it's still in `metrics` so that you never return items that were removed, while always returning the object that was most recently registered...

if you agree with this approach better defer to a different PR (post - v2)...

noambergIL: the only get is in the status page. i can change to read lock in whatever next pr i do

PR #1590, opened 2020-07-09
Fix missing checks in file provider of weights and current ref time

### LINKED ISSUE #1582: A couple of missing checks for mgmt file provider
when management file is parsed it should be checked for simple rules :
Historic response (timestamp given):
start <= end <= current

Latest response (no timestamp given):
start <= end = current

In Committess check that weight is not 0

(will fail if first time of node, but will also keep working in regular node)

noambergIL:
fixes #1582 

Historic response (timestamp given):
start <= end <= current

Latest response (no timestamp given):
start <= end = current

check weights are not 0

### DIFF
--- a/services/management/adapter/file_provider.go
+++ b/services/management/adapter/file_provider.go
@@ -57,7 +57,8 @@ func (mp *FileProvider) Get(ctx context.Context, referenceTime primitives.Timest
 		}
 	}
 
-	managementData, parseErr := mp.parseData(contents)
+	isHistoric := referenceTime != 0
+	managementData, parseErr := mp.parseData(contents, isHistoric)
 	if parseErr != nil {
 		mp.logger.Error("Provider file parsing error", log.Error(parseErr))
 		return nil, parseErr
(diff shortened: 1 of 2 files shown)


services/management/adapter/file_provider.go
```diff
@@ -169,7 +169,14 @@ func (mp *FileProvider) parseData(contents []byte) (*management.VirtualChainMana
 	vcString := fmt.Sprintf("%d", mp.config.VirtualChainId())
 	vcData, ok := data.VirtualChains[vcString]
 	if !ok {
-		return nil, errors.Errorf("could not find current vc in data")
+		return nil, errors.Errorf("could not find current vc in data (%d)", mp.config.VirtualChainId())
+	}
+
+	if data.CurrentRefTime != 0 {
+		if data.CurrentRefTime < data.PageStartRefTime || data.CurrentRefTime > data.PageEndRefTime {
```
ronnno, line 176: `data.CurrentRefTime > data.PageEndRefTime` if we have historic polling dont we want current to be greater than page end?...

PR #1594, opened 2020-07-13
Filesystem metrics and Histogram bug fix

ronnno:
- fixed formatting of results for non time values histograms
- added metrics to filesystem reading and writing (block sizes, read/write rates)
- added `UpdateMax` to gauge metric to track maximal values


noambergIL (approved):

noambergIL (changes requested):
need to find out what this new package is and if we can not have it.

PR #1595, opened 2020-07-13
Blocks file diagnostic tool

ronnno:
- depends on #1594 
- add a thin `diagnostics_parser.go` to wrap the codec. The alternative to using this wrapper is to make `Codec` public. Which is better? @noambergIL 
- `analyze_blocks.go` is a command line tool which receives a blocks file, scans all blocks and prints analysis. Currently only a very minimal analysis is done (histogram of block sizes - and one which does not contain all blocks - very basic)

Notice - we can delete the parser class if we make the codec public, the UI tool is easily extendable to answer any question about all blocks in any blocks file.

Tests are missing - need feedback before finalizing

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,12 +11,15 @@ require (
 	github.com/c9s/goprocinfo v0.0.0-20190309065803-0b2ad9ac246b
 	github.com/cespare/cp v1.1.1 // indirect
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd
+	github.com/coreos/etcd v3.3.13+incompatible
+	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f
 	github.com/deckarep/golang-set v1.7.1 // indirect
 	github.com/edsrzf/mmap-go v1.0.0 // indirect
 	github.com/elastic/gosigar v0.10.4 // indirect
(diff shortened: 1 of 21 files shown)


noambergIL:
This is a main that creates a histogram of block sizes ? i am not sure why should it be in ONG ? and why expose internal stuff ? 
this could be a tool outside that reads blocks from file and uses membuf to diagnose

i don't think it belongs to ONG base code

ronnno:
I agree - it should not be in ONG and that is also why I used `_main` in the folder. it's not part of the compilation. its an illustration of how you would use the codec or the `DiagnosticsParse` - it can be removed to a different repo.

that is not the main point. the point is how do we externalize the codec to other apps. there are three options:
1. make Codec capitalized
2. create a wrapper such as `DiagnosticParse` which is capitalized
3. Copy the Codec to a different repo - and either live with the duplicity of code or convert ONG to use that external dependency.

After implementing `DiagnosticParser` here - when my aim was to avoid production code changes as much as possible. I now start to think that duplicating the parser code is really the best option with the least friction and dependency to ONG.

What do you think?

ronnno:
reached the conclusion that it's best to capitalize (make public) the codec and reference it from an outside repo for the diagnostic parser

PR #1603, opened 2020-07-19
refactor to reduce code in metrics, and make it easier to understand

noambergIL:


instrumentation/metric/registry_test.go
```diff
@@ -12,12 +12,12 @@ import (
 )
 
 func TestInMemoryRegistry_ExportAll(t *testing.T) {
-	registry := NewRegistry()
-	gauge := registry.NewGauge("hello")
-	gauge.Add(1)
-
-	gaugeValue := registry.ExportAll()["hello"].(gaugeExport)
-	require.EqualValues(t, gaugeValue.Value, 1)
+	//registry := NewRegistry()
```
netoneko, line 15: Probably a good idea to uncomment this

noambergIL: i mean to rewrite export all ... this test is very vague (create one guage export it and check that there is one)

noambergIL: i am going to refactor export so the test will be different.

PR #1612, opened 2020-07-28
Metrics update names + structure (merge pending)

noambergIL:
Some name changes (with Backwards compatible for Prometheus)
each metric now has no "name"/"Value" pair ... only value.
a few "string" value metrics are also unmarshaled (topology, committee)
Structure of /metrics is now more flat 
Structure of /status is nested json
fix /status to show error string when needed


bootstrap/httpserver/status.go
```diff
@@ -10,20 +10,27 @@ import (
 type StatusResponse struct {
 	Timestamp time.Time
 	Status    string
-	Error     string
+	Error     string `json:",omitempty"`
```
ronnno, line 13: nice :)

instrumentation/metric/histogram_race.go
```diff
@@ -18,16 +18,16 @@ type Histogram struct {
...
 	return &Histogram{name: name}
 }
 
 func (h *Histogram) Name() string {
 	return h.name
 }
 
-func (h Histogram) Export() exportedMetric {
-	return &histogramExport{}
+func (h Histogram) Export() interface{} {
```
ronnno, line 29: this does not bother us, just curious why this did not work

instrumentation/metric/rate_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func testRateMeasure(t *testing.T, measure func(rate *Rate)) {
 	start := time.Now()
-	rate := newRateWihStart("tps", start)
+	rate := newRateWihStart("tps", "tps", start)
```
ronnno, line 17: with?

instrumentation/metric/reader.go
```diff
@@ -0,0 +1,90 @@
...
+	nameParts := deconstructNesting(name)
+	return travelNesting(mr, nameParts)
+}
+
+func deconstructNesting(name string) []string {
+	nameParts := strings.Split(name, ".")
+	nameLen := len(nameParts)
+
+	knownType := isNameAKnownType(nameParts[nameLen-1])
+	if nameLen > 0 && knownType {
```
ronnno, line 63: `nameLen` is always positive otherwise you would panic on the previous line no? maybe you want to check that it's greater than 1 here to ensure the resulting name will not be empty?

instrumentation/metric/text.go
```diff
@@ -39,10 +37,13 @@ func (t *Text) Name() string {
...
 }
 
-func (t *Text) Export() exportedMetric {
-	return textExport{
-		t.name,
-		t.value.Load().(string),
+func (t *Text) Export() interface{} {
+	value := t.value.Load().(string)
+	var x []interface{}
+	if err := json.Unmarshal([]byte(value), &x); err == nil {
```
ronnno, line 43: add a comment to explain why unmarshaling is required if this is a text type? is it to parse escape characters?

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,52 @@
...
+	metricReader, err := metric.NewReader(h.metricsUrl)
+	require.NoError(t, err, "should succeed reading metrics back into map")
+	require.NotEqual(t, 0, len(metricReader), "should not be empty")
+
+	blockHeight, found := metricReader.GetAsInt(blockstorage.MetricBlockHeight)
+	require.True(t, found, "block height should be found")
+	require.True(t, blockHeight > int64(CannedBlocksFileMinHeight))
+}
+
+func localWrite(t *testing.T, data []byte, filename string) {
```
ronnno, line 48: why do you check that you can write the output to a file in this test?

if this remains, please notice the error message below isn't w

instrumentation/metric/histogram_race.go
noambergIL: this is the races build version and i forgot to write the function so it didn't compile. very annoying this histogram races :)

PR #1616, opened 2020-08-11
Config cleanup

gadcl:


gadcl:
default configuration for production
- contains Ethereum connector config - should be removed with code cleanup
- contain logger config - should be removed with upgrade

updates to Boyar config:
- block-sync-reference-max-allowed-distance => remove
- active-consensus-algo => remove (default exists)
- consensus-context-triggers-enabled  => remove (default exists)
- transaction-pool-time-between-empty-blocks  => remove (default exists)
- lean-helix-consensus-round-timeout-interval  => remove (default exists)
- block-sync-no-commit-interval   => remove (default exists)
- consensus-context-system-timestamp-allowed-jitter   => remove (default exists)
- profiling    => remove (default exists)
- management-network-liveness-timeout => changed to => committee-grace-period  
      (on private network set value "committee-grace-period": "876000h”)

PR #1623, opened 2020-08-19
Disable eth conn status check

ronnno:
Since Ethereum connector is suspended no need to poll the status of the node


noambergIL:
duplicate, so deleting

PR #1630, opened 2020-10-12
Experimental/profile test

netoneko:
It seems that the merkle tree takes up a lot of memory space compared to actual data (667mb vs 148mb). The contract that is used is standard benchmark token with 619581 keys.

Another curious thing is that the memory is inflated over 3gb for 1gb block file, which is unusual and beyond the standard size of the chain that we expect.

Here is the link to the file that triggers the problem: https://s3.eu-central-1.amazonaws.com/orbs-tech-engagement-temp/blocks

image


duycon6868:
6AOKSIJV@Duycon6868

PR #1636, opened 2020-10-29
fix restart notifications

gadcl:


noambergIL (approved):

bootstrap/httpserver/status.go
```diff
@@ -41,40 +42,39 @@ func (s *HttpServer) getStatus(w http.ResponseWriter, r *http.Request) {
...
 
 	if len(s.config.ManagementFilePath()) != 0 && s.config.ManagementPollingInterval() > 0 {
 		maxIntervalSinceLastSuccessfulManagementUpdate := int64(s.config.ManagementPollingInterval().Seconds()) * 20
-		managementLastSuccessfullUpdate := s.getGaugeValueFromMetrics("Management.Data.LastSuccessfulUpdateTime")
-		if managementLastSuccessfullUpdate == 0 {
-			return "Management Service has never successfully updated"
-		}
-		if managementLastSuccessfullUpdate + maxIntervalSinceLastSuccessfulManagementUpdate < time.Now().Unix() {
-			return "Last successful Management Service update was too long ago"
+		if managementLastSuccessfullUpdate, err := s.getGaugeValueFromMetrics("Management.Data.LastSuccessfulUpdateTime"); err == nil {
```
noambergIL, line 57: i don't like this. instead move the metricObj := s.metricRegistry.Get(name) to be the big if instead of errors.

services/blockstorage/adapter/filesystem/file_system_persistence.go
```diff
@@ -25,17 +25,20 @@ import (
...
 
 type metrics struct {
-	sizeOnDisk *metric.Gauge
+	sizeOnDisk          *metric.Gauge
+	indexLastUpdateTime *metric.Gauge
 }
 
 const blocksFilename = "blocks"
 
 func newMetrics(m metric.Factory) *metrics {
```
noambergIL, line 38: if you are already touching this how about rename the functiont to generateBlockStorageMetrics ?

services/blockstorage/adapter/filesystem/file_system_persistence.go
```diff
@@ -242,6 +245,9 @@ func buildIndex(r io.Reader, firstBlockOffset int64, logger log.Logger, c blockC
 		if err != nil {
 			return nil, errors.Wrap(err, "failed building block height index")
 		}
+		if metrics != nil {
```
noambergIL, line 248: no. remove the if , in all the tests use NewRegistry()
not a good idea that in production code you allow a value to be null and ignore just to make your tests easier.

services/blockstorage/adapter/filesystem/file_system_persistence.go
```diff
@@ -268,6 +274,7 @@ func (f *BlockPersistence) WriteNextBlock(blockPair *protocol.BlockPairContainer
 		return false, f.bhIndex.getLastBlockHeight(), errors.Wrap(err, "failed to update index after writing block")
 	}
 
+	f.metrics.indexLastUpdateTime.Update(time.Now().Unix())
```
noambergIL, line 277: this may fail if metrics is nil

noambergIL (changes requested):
please view my comments

services/blockstorage/adapter/filesystem/file_system_persistence.go
gadcl: this is the case with all ONG metrics updates ..

noambergIL (approved):

services/blockstorage/adapter/filesystem/file_system_persistence.go
```diff
@@ -25,17 +25,20 @@ import (
...
 }
 
 const blocksFilename = "blocks"
 
-func newMetrics(m metric.Factory) *metrics {
+func generateBlockStorageMetrics(m metric.Factory) *metrics {
 	return &metrics{
-		sizeOnDisk: m.NewGauge("BlockStorage.FileSystemSize.Bytes"),
+		sizeOnDisk:          m.NewGauge("BlockStorage.FileSystemSize.Bytes"),
+		indexLastUpdateTime: m.NewGauge("BlockStorage.FileSystemIndex.LastUpdateTime"),
```
ronnno, line 41: should be instantiated with the current timestamp value not zero

gadcl: why? building the index occurs immediately after init (init -> open blocks file -> build index), and on build index or any append block current timestamp is used

ronnno: a possible race condition. we should at least make sure that we don't ever Error when the value is 0

PR #1641, opened 2020-11-05
Registry metric double name fix, clean up outgoing cons metrics

noambergIL:
registry work with names not pointer (except remove that checks same object), try and remove metric from outgoing_connection.go and queue.go before creating new one to avoid "double register"

### DIFF
--- a/instrumentation/metric/registry.go
+++ b/instrumentation/metric/registry.go
@@ -8,7 +8,6 @@ package metric
 
 import (
 	"context"
-	"fmt"
 	"github.com/orbs-network/govnr"
 	"github.com/orbs-network/orbs-network-go/synchronization"
 	"github.com/orbs-network/orbs-spec/types/go/primitives"
@@ -80,33 +79,29 @@ func (r *inMemoryRegistry) WithNodeAddress(nodeAddress primitives.NodeAddress) R
 func (r *inMemoryRegistry) Get(metricName string) metric {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	for _, metric := range r.mu.metrics {
-		if metric.Name() == metricName {
-			return metric
-		}
-	}
-	return nil
+	return r.mu.metrics[metricName]
 }
 
+// Only if the actual metric (the object/pointer) is the same remove metric
 func (r *inMemoryRegistry) Remove(m metric) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	key := fmt.Sprintf("%p", m)
-	if _, isMetricInMap := r.mu.metrics[key]; isMetricInMap {
-		delete(r.mu.metrics, key)
+	if mapMetric, isMetricInMap := r.mu.metrics[m.Name()]; isMetricInMap {
+		if mapMetric == m {
+			delete(r.mu.metrics, m.Name())
+		}
 	}
 }
 
 func (r *inMemoryRegistry) register(m metric) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	key := fmt.Sprintf("%p", m)
-	if _, isMetricInMap := r.mu.metrics[key]; isMetricInMap {
+	if _, isMetricInMap := r.mu.metrics[m.Name()]; isMetricInMap {
 		err := errors.Errorf("a metric with name %s is already registered", m.Name())
 		panic(err)
 	}
 
-	r.mu.metrics[key] = m
+	r.mu.metrics[m.Name()] = m
 }
 
 func (r *inMemoryRegistry) NewRate(name string) *Rate {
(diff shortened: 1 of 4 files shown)


instrumentation/metric/registry.go
```diff
@@ -80,33 +79,29 @@ func (r *inMemoryRegistry) WithNodeAddress(nodeAddress primitives.NodeAddress) R
...
-	for _, metric := range r.mu.metrics {
-		if metric.Name() == metricName {
-			return metric
-		}
-	}
-	return nil
+	return r.mu.metrics[metricName]
 }
 
+// Only if the actual metric (the object/pointer) is the same remove metric
```
ronnno, line 85: would be nice if we would return two booleans - or a tristate:
1. metric was found
2. metric was removed

but if it's not required no need,... just a thought

services/gossip/adapter/tcp/outgoing_connection.go
```diff
@@ -52,20 +52,37 @@ func newOutgoingConnection(peer adapter.TransportPeer, parentLogger log.Logger,
...
 
 	return client
 }
 
+// This is a round-about way to clean up these metrics that can be left over from previous connection
+func generateMetrics(peerHexAddress string, metricFactory metric.Registry) (*metric.Gauge, *metric.Gauge){
+	sendErrorsName := fmt.Sprintf("Gossip.OutgoingConnection.SendError.%s.Count", peerHexAddress)
+	sendErrorMetric := metricFactory.Get(sendErrorsName)
+	if sendErrorMetric != nil {
+		metricFactory.Remove(sendErrorMetric)
```
ronnno, line 76: remove works seamlessly even if it does not exist. so the check is redundant

services/gossip/adapter/tcp/outgoing_connection.go
```diff
@@ -52,20 +52,37 @@ func newOutgoingConnection(peer adapter.TransportPeer, parentLogger log.Logger,
...
+	}
+	sendQueueErrorsName := fmt.Sprintf("Gossip.OutgoingConnection.EnqueueErrors.%s.Count", peerHexAddress)
+	sendQueueErrorMetric := metricFactory.Get(sendQueueErrorsName)
+	if sendQueueErrorMetric != nil {
+		metricFactory.Remove(sendQueueErrorMetric)
+	}
+	return metricFactory.NewGauge(sendErrorsName),  metricFactory.NewGauge(sendQueueErrorsName)
```
ronnno, line 83: this will work when there is only one goroutine trying to add the new metric. when two race it could lead to panic.

I see two alternatives:
1. add a reference count to each metric, then whenever you `NewGauge` for an existing metric you increase the reference count and return the preexisting metric. then, `remove` kills it only when the reference count 
2. `NewGauge` could always replace under lock in case the old name is already registered. 

each method (the two alternatives and your solution) has advantages and disadvantages. what do you think?

services/gossip/adapter/tcp/queue.go
```diff
@@ -31,16 +31,21 @@ type transportQueue struct {
...
 	}
 	q.protected.bytesLeft = maxSizeBytes
 
-	q.usagePercentageMetric =
-		metricFactory.NewGaugeWithPrometheusName(fmt.Sprintf("Gossip.OutgoingConnection.QueueUsage.%s.Percent", peerNodeAddress), fmt.Sprintf("Gossip.OutgoingConnection.Queue.Usage.%s.Percent", peerNodeAddress))
+	// round-about way to remove old queue metric if exists
+	queueUsageName := fmt.Sprintf("Gossip.OutgoingConnection.QueueUsage.%s.Percent", peerNodeAddress)
+	queueUsageMetric := metricFactory.Get(queueUsageName)
+	if queueUsageMetric != nil {
+		metricFactory.Remove(queueUsageMetric)
```
ronnno, line 46: once again, remove is enough

ronnno (approved):
please see comments. there is a possible panic in a race condition since remove and register happen under two different locks. I'm not sure it's a scenario we should be worried about in this case but  I'd like your opinion on the alternatives i suggest...

services/gossip/adapter/tcp/outgoing_connection.go
noambergIL: All the connections are open and closed under lock for the outgoing_connections so it is not needed.
the newgauge doesn't need to know there is a "replace" thing its not its job.

services/gossip/adapter/tcp/outgoing_connection.go
gadcl: perhaps consider logging if the metric already existed - even if just for now, so we don't mute out this error

gadcl (approved):

PR #1651, opened 2020-12-27
Mgmt url related small fixes

noambergIL:
change timeouts in config (30sec poll, 500mb filesize),
add timeout for the http get in the mgmt-file-provider

### DIFF
--- a/services/management/adapter/file_provider.go
+++ b/services/management/adapter/file_provider.go
@@ -79,6 +79,9 @@ func (mp *FileProvider) generatePath(referenceTime primitives.TimestampSeconds) 
 
 func (mp *FileProvider) readUrl(path string) ([]byte, error) {
 	res, err := mp.client.Get(path)
+	if res != nil && res.Body != nil {
+		defer res.Body.Close()
+	}
 
 	if err != nil || res == nil {
 		return nil, errors.Wrapf(err, "Failed http get of url %s", path)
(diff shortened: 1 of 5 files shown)


services/management/adapter/file_provider.go
```diff
@@ -78,7 +83,8 @@ func (mp *FileProvider) generatePath(referenceTime primitives.TimestampSeconds)
 }
 
 func (mp *FileProvider) readUrl(path string) ([]byte, error) {
-	res, err := http.Get(path)
+	res, err := mp.client.Get(path)
```
talkol, line 86: check err before doing defer or else res might not be defined

this is from the official docs:

```
resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
```

talkol (changes requested):

services/management/adapter/file_provider.go
```diff
@@ -78,15 +78,21 @@ func (mp *FileProvider) generatePath(referenceTime primitives.TimestampSeconds)
...
 	} else if res.ContentLength > 0 && uint32(res.ContentLength) > mp.config.ManagementMaxFileSize() { // TODO when no length given find other way ?
 		return nil, errors.Wrapf(err, "Failed http get response too big %d", res.ContentLength)
 	}
 
-	return ioutil.ReadAll(res.Body)
+	contents, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not read http body")
+	}
+	defer res.Body.Close()
```
talkol, line 93: We want to do close the Body if `res.Body != null`, right now if you had an error reading the body you will not close it

talkol (approved):

PR #1652, opened 2021-03-11
Feature/vc max sizes

noambergIL:
merge only after we make sure the spec is correct sha and we finalize the default sizes.


gadcl (approved):
looks good

